PR #4, opened 2019-11-13
initial startup speed-up

maedoc:
When just trying to write unit tests and run them, the startup time of
the library is slow enough to get distracted (8.2 s on i5 SSD Mac).
Looking at the extensive log output, I saw enough delay between warnings
of different models to realize that importing all models is causing this
delay.

I then commented the imports of tvb.simulator.models.__init__ and saw
startup time drop to 0.5 s.  I would guess that this speed drop is due
to the gufunc or neotraits class registration, not yet sure which.

The current fix is breaking code such as

     from tvb.simulator.models import Linear

so it would be useful to resolve the underlying problem *or* implement
a lazy loading approach, since, at runtime, only one model is usually
in use; loading all is excessive.

@mihandrei @liadomide ideas?


maedoc:
The workaround implemented works but requires some familiarity with Python's modules to understand.  The speed up is as reported before since the model modules are lazily loaded.

liadomide:
I have a small suggestion: to leave just yourself as author on this file, or to remove authors entirely from here. Now, I think it is an incomplete list of authors for tvb models

maedoc:
@liadomide why is this assigned after merge?

liadomide:
I was playing with GitHub GUI.
I wanted to have a filled top view of the PRs and someone assigned to each (for the closed ones the people involved).
.... but mainly playing with the GUI. Sorry for the spam :-)

maedoc:
no problem play away, just wanted to make sure I'm not missing anything!

PR #5, opened 2019-11-13
Fix phase plane plot for current matplotlib

i-Zaak:
The `set_color_cycle` has been deprecated in matplotlib since version 1.5. Time to move on... Currently the `PhasePlaneInteractive` crashes (canvas turns white) on a change of parameter.


maedoc:
Can you just confirm that the matplotlib version TVB uses has the new thing?  I remember we had a really old mpl for a while.

i-Zaak:
Short answer: yes. Long answer: the only mention of a matplotlib version in this repo I have found is the `matplotlib==2.1.0` in the `build/Dockerfile-py2`. That version has both `axes.Axes.set_color_cycle` and `axes.Axes.set_prop_cycle`, where the former produces a deprecation warning.

PR #6, opened 2019-11-14
fixed constraints handling in the simulator

i-Zaak:
The constraints are transferred from the `Model` to the `Integrator` in the `Simulator.preconfigure` function. However, while it correctly takes the `Model.state_variable_boundaries`, it creates an incorrect attributes `Integrator.constraint_state_variable_indices` and `Integratorconstraint_state_variable_boundaries`. These are then not used in the integration. 

See the before / after time series for a variable with [0, np.inf] constraint:

download (8)

download (7)

### DIFF
--- a/scientific_library/tvb/simulator/simulator.py
+++ b/scientific_library/tvb/simulator/simulator.py
@@ -207,11 +207,11 @@ class Simulator(HasTraits):
                 indices.append(self.model.state_variables.index(sv))
                 boundaries.append(sv_bounds)
             sort_inds = numpy.argsort(indices)
-            self.integrator.constraint_state_variable_indices = numpy.array(indices)[sort_inds]
-            self.integrator.constraint_state_variable_boundaries = numpy.array(boundaries)[sort_inds]
+            self.integrator.bounded_state_variable_indices = numpy.array(indices)[sort_inds]
+            self.integrator.state_variable_boundaries = numpy.array(boundaries)[sort_inds]
         else:
-            self.integrator.constraint_state_variable_indices = None
-            self.integrator.constraint_state_variable_boundaries = None
+            self.integrator.bounded_state_variable_indices = None
+            self.integrator.state_variable_boundaries = None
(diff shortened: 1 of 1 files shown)


maedoc:
I wanted to make a similar point as Lia.  For this sort of thing, it is important to first establish a failing unit test or fix one that should have failed, and then fix the bug itself. Otherwise we risk a future regression.

i-Zaak:
Agree -- I have an example for the Montbrio model (in the figures), but the same should happen for the reduced Wong Wang model. Or one could write unit test(s) with a simple model which is guaranteed to hit the constraints in few steps...

maedoc:
That is why we have the Identity and Linear models, to make writing such tests easy.

Can @dionperd comment on this?

maedoc:
I looked over the code here, the [unit test](https://github.com/the-virtual-brain/tvb-root/blob/master/scientific_library/tvb/tests/library/simulator/integrators_test.py#L129) just checks that an integrator applies bounding correctly, but there's no test that the simulator does this transfer correctly.  I would let it slide for now and merge this, since the changes match what the unit test *does* check.

dionperd:
Hello. It seems correct now:
https://github.com/the-virtual-brain/tvb-root/blob/dbb0449b301013763a1b4d559728b69ab2aa0045/scientific_library/tvb/simulator/simulator.py#L193

Is it really necessary to make a test about the preconfiguration of the integrator now?
If you show me other similar tests, I could design one.
Just writing a model that is meant to hit the boundaries is not enough and not specific enough in my opinion. A proper test should show that those 10 lines in simulator.preconfigure do what they are expected to do, in all different cases of boundaries.

Maybe I could make a similar test to the one we have now but setting the boundaries inside a dummy model specifically made for the test.

maedoc:
I would suggest we don't write a test after the fact, but next time, try to come up with a small test which fails before fixing a bug.

i-Zaak:
I will write the test eventually to get myself familiar with the tvb tests
in general. Eventually might be in January though...

PR #7, opened 2019-11-14
WIP Gradients

maedoc:
This is a *work in progress* to enable gradient calculation in TVB, which would enable many of the modern machine learning techniques around data fitting.


maedoc:
I'm going to close this PR as I am not sure this will be sustainable in the long run.  We can keep the branch and reopen in the future.

PR #8, opened 2019-11-14
WIP Test GitHub Actions for CI Docker image build

maedoc:
GitHub added actions which looked a lot like GitLab's CI or Jenkins in functionality.  It would be nice if only for our PR workflow to have the checks that tests pass and that the images are built.  This doesn't replace Travis/Jenkins for TVB but instead complements the Git workflow.

### DIFF
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -12,3 +12,9 @@ jobs:
     - uses: actions/checkout@v1
     - name: Build the Docker image
       run: cd tvb_build/docker && docker build . --file Dockerfile --tag tvb
+    - name: Login to registry
+      run: docker login docker.pkg.github.com -u USERNAME -p ${{ secrets.GITHUB_TOKEN }}
+    - name: Tag new image for registry
+      run: docker tag tvb docker.pkg.github.com/the-virtual-brain/tvb-root/tvb:latest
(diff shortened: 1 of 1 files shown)


maedoc:
So this works, but it takes almost half an hour to build & push the image.

liadomide:
How about doing the build and push of the docker image from Jenkins (https://jenkins.codemart.ro/view/TVB/job/TVB%20-%20Root/ WIP), and on Github just pull the image from Docker Hub and run unit-tests ?
I foresee seldom changes in the Dockerfile, and most of the changes happening in the Python code

maedoc:
Yes, I agree: the main thing to avoid is rebuilding the Docker image, especially for testing, because the pre-built image can be run even without adding source code, e.g.
```bash
image=tvb/root:master
docker run -v $PWD:/opt/tvb-root $image pytest /opt/tvb-root
```
On GitLab we do something like
```yaml
build_image:
  script:
    - docker build ...
  only:
    - docker-*
```
to only build when a branch name matches `docker-*`.  Maybe something similar is possible here or Jenkins, if you think the latter is easier, that would be fine with me (maybe you already have it setup?)

maedoc:
I will leave this closed as a proof of concept for now, and we can revisit it later.

PR #11, opened 2019-11-28
Fix ReducedWongWangExcInh parameter tau_i

neich:
Parameter tau_i has a minimum of 50.0 which prevents its initialization to the default value of 10.0 (value used in Deco 2014)

### DIFF
--- a/scientific_library/tvb/simulator/models/wong_wang_exc_inh.py
+++ b/scientific_library/tvb/simulator/models/wong_wang_exc_inh.py
@@ -161,7 +161,7 @@ class ReducedWongWangExcInh(ModelNumbaDfun):
     tau_i = NArray(
         label=r":math:`\tau_i`",
         default=numpy.array([10., ]),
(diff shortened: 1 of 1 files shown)


dionperd:
I agree with the change. This was overlooked. Thanks!
Does this really prevent the initialization with the default value of 10.0 outside the GUI?

I would also suggest the following changes:
tau_i: lo=5.0,, hi=100
tau_e: lo=10.0,

just in case a user wants to experiment with equaling them to the faster or slower time scale.

neich:
Ok, it seems that the problem is only with the GUI, the parameter is internally initialized to 10. Anyway, I still think it should be corrected to make it consistent (the initial value should not be outside the defined range)

About tau_e I'm not sure

maedoc:
I'm ok for the change.  The parameter bounds are currently for the GUI only, but in the future would serve to set priors for data fitting use cases, so it's worth thinking about their values a bit.

PR #13, opened 2019-12-06
Correct boundaries

dionperd:
I have made corrections and refactored the state boundaries handling of the simulator.

1. I have moved the user model input processing to the base Model class, and I am allowing also setting infinite boundaries.
None and infinite boundaries are set equal to the maximum "single" float value.
We can go for double as well if you wish.

2. Therefore, there is less configuration now in the Simulator. I am making sure that the resulting boundaries are of float64 format though.

3.  I have made new tests, separately for the the Model configuration of the boundaries (in models_tests), the respective simulator configuration of the integrator (in simulator_tests), and the actual bounding of the state via an integrators' test, extended to all of the available integration schemes.

### DIFF
--- a/scientific_library/tvb/simulator/integrators.py
+++ b/scientific_library/tvb/simulator/integrators.py
@@ -464,17 +464,21 @@ class SciPySDE(SciPyODEBase):
             self.clamp_state(X_next)
         return X_next
 
+
 class VODE(SciPyODE, Integrator):
     _scipy_ode_integrator_name = "vode"
     _ui_name = "Variable-order Adams / BDF"
+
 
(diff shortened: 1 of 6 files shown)


dionperd:
I made the changes required by Marmaduke.

I refactored using a _setup_sv_boundaries method.
I am now using double precision.

Adjusted (and also renamed) the tests accordingly.

maedoc:
>  have also added a Warning for the case that the user sets a non dict boundary model input, in which case it is ignored. Shall we make this an error?

warning is enough to assist debugging; if someone decides to pass in something that behaves like a dict but fails a `isinstance(obj, dict)` check then it is their fish to fry

PR #14, opened 2019-12-13
TVB-2556 Fix after lazy models import changes

bvalean:
In tvb-framework and library tests we have some use cases where all the models are needed (e.g ui forms and tests). For this we propose adding an enum with all the models in tvb.simulator.models.init.py.

### DIFF
--- a/framework_tvb/tvb/adapters/simulator/model_forms.py
+++ b/framework_tvb/tvb/adapters/simulator/model_forms.py
@@ -28,33 +28,33 @@
 #
 #
 
-from tvb.simulator.models import *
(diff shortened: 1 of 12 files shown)


maedoc:
The enum is a good idea, wish I had thought of it.  I don't like the the `get_class()` style but it is certainly easier to code-navigate that rather than module properties (which are just bending-over-backwards so the init optimization retains compatibility) so that's fine too.

+1 for merge

liadomide:
Then, @bvalean pls update this to be compatible with the latest state in master, let's check quickly that outcome, and then we can merge.

PR #16, opened 2019-12-16
TVB-2565 fix numpy integers serialization

bvalean:
In Phase Plane page there was a problem with numpy integer serializator for Linear and Epileptor Resting State models. Review this fix in the future.


maedoc:
could you use NumPy's base types as explained here:

  https://stackoverflow.com/a/37727662

PR #17, opened 2019-12-28
TVB-2563 Adjust Zerlaut models

liadomide:
Migrate changes from tvb-library into tvb-root repo.
Basically changes from here https://github.com/the-virtual-brain/tvb-library/pull/91 are copied into tvb-root 
We should see how this new model code behaves in Phase Plane and Simulator and aftrwards merge.
We now even have a working example in a notebook


liadomide:
@lionelkusch we would appreciate your feedback on these changes.
Pls also look at the comments in Jira complementing these changes:
https://req.thevirtualbrain.org/browse/TVB-2563

PR #18, opened 2020-01-07
TVB-2579 Fix state variables reset on changing models

bvalean:
If we modify the axis range in the tvb-web gui Phase Plane page for a model, these new values will be used as default ranges when this model is selected again.


liadomide:
@mihandrei we would greatly appreciate your feedback on these changes

liadomide (approved):

PR #20, opened 2020-01-13
TVB-2572 Integrate Swagger with REST api

gabiflorea10:


framework_tvb/tvb/interfaces/rest/server/rest_api.py
```diff
@@ -28,7 +28,7 @@
 #
 #
 
-from flask_restful import Api
+from flask_restplus import Api
```
liadomide: Are you certain there are only 2 imports ?
We should make sure we do not depend on both libraries

framework_tvb/tvb/interfaces/rest/server/run.py
```diff
@@ -76,19 +76,21 @@ def initialize_flask():
...
+
+    name_space.add_resource(GetUsersResource, '/users')
+    name_space.add_resource(GetProjectsListResource, '/users/<string:username>/projects')
+    name_space.add_resource(GetDataInProjectResource, '/projects/<string:project_gid>/data')
+    name_space.add_resource(GetOperationsInProjectResource, '/projects/<string:project_gid>/operations')
+    name_space.add_resource(RetrieveDatatypeResource, '/datatypes/<string:datatype_gid>')
+    name_space.add_resource(GetOperationsForDatatypeResource, '/datatypes/<string:datatype_gid>/operations')
+    name_space.add_resource(FireSimulationResource, '/simulation/<string:project_gid>')
+    name_space.add_resource(LaunchOperationResource, '/operations/<string:project_gid>/algorithm'
+                                                         '/<string:algorithm_module>/<string:algorithm_classname>')
```
liadomide: Is this code properly aligned ?

liadomide:
flask_restplus should be added in setup.py also, right?

framework_tvb/tvb/interfaces/rest/server/rest_api.py
gabiflorea10: Yes, there are only 2 imports.

framework_tvb/tvb/interfaces/rest/server/run.py
gabiflorea10: The problem will be fixed in a later commit.

gabiflorea10:
> flask-restplus should be added in setup.py also, right?

yes

gabiflorea10:
> Changes look ok. Pls just answer my concerns in the comments. If everything confirmed, then we can merge

I suggest to not merge yet, because some methods are not working right know.
After fixing these methods, the code could be merged.

PR #21, opened 2020-01-20
Package tvb-rest-client to pypi 

gabiflorea10:


gabiflorea10:
> First commit refers to a package on pypi test.

tvb_build/package_for_pip_client.sh
```diff
@@ -0,0 +1,42 @@
...
+
+for pipPackage in "${folders2pack[@]}"; do
+
+    echo "============================="
+    echo " Packing: " $pipPackage
+    echo "============================="
+
+    cd $pipPackage
+    python setup_client.py sdist
+    python setup_client.py bdist_wheel
```
liadomide: These 2 lines I expect, but the loop over folders2pack does not make any sense.
Have you considered to simply add the setup_client.py use in the already existent SH (package_for_pip) ?

framework_tvb/setup_client.py
```diff
@@ -0,0 +1,72 @@
...
+import os
+import shutil
+import setuptools
+
+
+VERSION = "2.0a1"
+
+TVB_TEAM = "Mihai Andrei, Lia Domide, Stuart Knock, Bogdan Neacsa, Paula Sansz Leon, Marmaduke Woodman"
+
+TVB_INSTALL_REQUIREMENTS = ["allensdk", "BeautifulSoup4", "cherrypy", "flask", "flask-restplus", "formencode", "Jinja2",
```
liadomide: This list would be too long as a dependency path

framework_tvb/setup_client.py
```diff
@@ -0,0 +1,72 @@
...
+                            "h5py", "networkx", "nibabel", "numpy", "Pillow", "psutil", "scipy",
+                            "simplejson", "sqlalchemy", "sqlalchemy-migrate", "tvb-data", "tvb-gdist", "tvb-library"]
+
+with open(os.path.join(os.path.dirname(__file__), 'README.rst')) as fd:
+    DESCRIPTION = fd.read()
+
+setuptools.setup(name="tvb-rest-client-test27",
```
liadomide: Have you considered adding this to the existing setup.py ?

tvb_build/package_for_pip_client.sh
gabiflorea10: That's true. The loop over folders2pack should be removed. 
For second question... Yes, I took into consideration this solution too, but I think that probably we want to package the rest-client independently of the entire framework and having two different SH for packaging allow us to do this.

framework_tvb/setup_client.py
gabiflorea10: As I aforementioned, I followed the idea of having a clear separation between framework packaging and rest-client packaging. 
But, If we prefer to have all packaging setups together, we can try to do this.
Also, for the previous message related to SH, we can try to put all together, if this is the preferred way.

framework_tvb/setup_client.py
gabiflorea10: Yes, it should be refactored.

tvb_build/package_for_pip_client.sh
liadomide: We rather want to have one single Jenkins build doing all Pypi packages. With any new release in the future I hope we can release everything at the same time

framework_tvb/setup_client.py
liadomide: Yes, I would prefer to put things together

liadomide:
I close this because Paula took over and did the changes in tvb-rest directly

PR #23, opened 2020-01-28
fix integrator test failure

maedoc:
this addresses minor failure to configure noise causing some stochastic integrator test failures.


maedoc:
I am surprised that Hudson/Jenkins did not see this?

liadomide:
On my env the test passes both with and without this change.
Same on Jenkins....

PR #25, opened 2020-01-29
Fixes for MATLAB API

maedoc:
This should address [TVB-2586](https://req.thevirtualbrain.org/browse/TVB-2586) with the demos updates for Py3 & neotraits compat.


liadomide:
I can not test these, but look decent to me. Should we merge?

PR #28, opened 2020-02-05
TVB-2578 Default simulation name

adrian-dordea:
set default simulation name pre-filled on the Burst page

### DIFF
--- a/framework_tvb/tvb/adapters/simulator/simulator_adapter.py
+++ b/framework_tvb/tvb/adapters/simulator/simulator_adapter.py
@@ -363,9 +363,7 @@ class SimulatorAdapter(ABCAsynchronous):
             ts_index.state = 'INTERMEDIATE'
 
             state_variable_dimension_name = ts.labels_ordering[1]
-            if ts_index.user_tag_1:
-                ts_index.labels_dimensions[state_variable_dimension_name] = ts.user_tag_1.split(';')
-            elif m_name in self.HAVE_STATE_VARIABLES:
+            if m_name in self.HAVE_STATE_VARIABLES:
                 selected_vois = [self.algorithm.model.variables_of_interest[idx] for idx in monitor.voi]
                 ts.labels_dimensions[state_variable_dimension_name] = selected_vois
                 ts_index.labels_dimensions = json.dumps(ts.labels_dimensions)
--- a/framework_tvb/tvb/adapters/simulator/simulator_adapter.py
+++ b/framework_tvb/tvb/adapters/simulator/simulator_adapter.py
@@ -40,6 +40,8 @@ Few supplementary steps are done here:
 .. moduleauthor:: Stuart A. Knock <Stuart@tvb.invalid>
 
 """
+import json
+
 from tvb.core.neotraits.view_model import ViewModel, DataTypeGidAttr
 from tvb.datatypes.connectivity import Connectivity
 from tvb.datatypes.cortex import Cortex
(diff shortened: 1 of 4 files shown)


framework_tvb/tvb/adapters/simulator/simulator_fragments.py
```diff
@@ -170,7 +170,7 @@ class SimulatorFinalFragment(ABCAdapterForm):
 
     def __init__(self, prefix='', project_id=None):
         super(SimulatorFinalFragment, self).__init__(prefix, project_id)
-        self.simulation_name = ScalarField(Attr(str, doc='Name for the current simulation configuration',
+        self.simulation_name = ScalarField(Attr(str, doc='Name for the current simulation configuration', default="default_simulation",
```
liadomide: I would expect the name to be different at every run.
In versions 1.* of TVB-Distribution you can see how it was behaving. and giving names "simulation-x" where x is incremental, a number

framework_tvb/tvb/adapters/simulator/simulator_fragments.py
```diff
@@ -168,9 +168,10 @@ def __init__(self, prefix='', project_id=None):
 
 class SimulatorFinalFragment(ABCAdapterForm):
 
-    def __init__(self, prefix='', project_id=None):
+    def __init__(self, prefix='', project_id=None, simulation_number=1):
         super(SimulatorFinalFragment, self).__init__(prefix, project_id)
-        self.simulation_name = ScalarField(Attr(str, doc='Name for the current simulation configuration',
+        default_simulation_name = "default_simulation_" + str(simulation_number)
```
liadomide: I would put here "simulation_"+count, to match the previous names:
https://www.thevirtualbrain.org/tvb/zwei/image-zoom/122365/27?stage=brainsimulator

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -794,7 +794,8 @@ def set_simulation_length(self, **data):
         is_simulator_load = common.get_from_session(common.KEY_IS_SIMULATOR_LOAD) or False
         session_burst_config = common.get_from_session(common.KEY_BURST_CONFIG)
 
-        next_form = SimulatorFinalFragment()
+        simulation_number = self.burst_service2.get_available_bursts(common.get_current_project().id).__len__() + 1
```
liadomide: I know it was not generated at this change, but pls rename `self.burst_service2` into `self.burst_service`

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -780,7 +780,8 @@ def set_simulation_length(self, **data):
         is_simulator_load = common.get_from_session(common.KEY_IS_SIMULATOR_LOAD) or False
         session_burst_config = common.get_from_session(common.KEY_BURST_CONFIG)
 
-        next_form = SimulatorFinalFragment()
+        simulation_number = self.burst_service.get_available_bursts(common.get_current_project().id).__len__() + 1
```
liadomide: I suggest using here a count query. This would become a performance problem (to load all bursts and just count).
Have you tested what is happening when there are no simulations?

liadomide:
Pls in the future no longer mix fixes, and rather create separate branches for each of the tasks you solve. Now here I see both TVB-2595 and TVB-2578

framework_tvb/tvb/adapters/simulator/simulator_adapter.py
```diff
@@ -360,12 +362,13 @@ def launch(self, view_model):
...
             ts_index.state = 'INTERMEDIATE'
 
-            # state_variable_dimension_name = ts.labels_ordering[1]
-            # if ts_index.user_tag_1:
-            #     ts_index.labels_dimensions[state_variable_dimension_name] = ts.user_tag_1.split(';')
-            # elif m_name in self.HAVE_STATE_VARIABLES:
-            #     selected_vois = [self.algorithm.model.variables_of_interest[idx] for idx in monitor.voi]
-            #     ts.labels_dimensions[state_variable_dimension_name] = selected_vois
+            state_variable_dimension_name = ts.labels_ordering[1]
+            if ts_index.user_tag_1:
```
liadomide: This if branch should be dropped, and remain only the second branch `if m_name in self.HAVE_STATE_VARIABLE`

liadomide: Pls test how this fix works in case one of EEG, MEG or Bold Monitors are being used. I assume the bug will not be fixed, and we should either have some acceptable defaults, or handle in the viewers this case further

PR #31, opened 2020-02-07
remove opencl module

maedoc:
OpenCL isn't on our roadmap anymore, and no one has
really made use or tested so we should let it RIP


maedoc:
@liadomide are you OK with the removal?  This was a GSoC contribution but I prefer to take it out and bring it back later if there is will/desire. wdyt?

liadomide:
I fully agree to remove these. I also meant to ask you about this code, but somehow I forgot.

maedoc:
thanks!

PR #32, opened 2020-02-07
WIP update setup scripts

maedoc:
The existing scripts don't work for me but can't mess
with the CI builds.  At some point it may be better
to write this in Python since Conda is Python and
APIs would be cleaner that various shell & batch
scripts.


liadomide:
Pls feel free to change create_conda_build.* scripts in any way to see fit. Currently the Jenkins CI does not use these. 
We used these as documentation/draft for building the native installations needed for TVB_Distribution builds, but they do not seem to be up to date

liadomide:
As part of task TVB-2523 we managed to have functional Docker files with Anaconda for both Linux and Windows. Those scripts will be maintained.
I propose we remove all `create_conda_build.*` scripts, as they are hardly ever touched, and most of the time left outdated.

liadomide:
I will close this (without merge) for the above reason. 
In #133 also the other similar scripts `create_conda_build` are removed.

PR #33, opened 2020-02-07
WIP symbolic jacobians

maedoc:
symbolic jacobians are in reach for most of our models,
especially with the upcoming DSL.  this is a first
sketch which computes derivatives symbolically with
sympy for Jansen Rit and generates a sequences of
instructions which can be strung together, eval'd and
then jit'd with numba.

one could then also specialize when parameter values
are known by substituting the values in prior to
doing the symbolics, which will help as well.


maedoc:
shelving this for now, new stuff is in the oven

PR #36, opened 2020-02-10
Neotraits compatibility with PostgreSQL

gabiflorea10:


framework_tvb/tvb/adapters/datatypes/db/connectivity.py
```diff
@@ -67,6 +67,12 @@ def fill_from_has_traits(self, datatype):
         self.undirected = datatype.undirected
         self.weights_min, self.weights_max, self.weights_mean = from_ndarray(datatype.weights)
         self.tract_lengths_min, self.tract_lengths_max, self.tract_lengths_mean = from_ndarray(datatype.tract_lengths)
+        self.weights_min = numpy.float64(self.weights_min)
+        self.weights_mean = numpy.float64(self.weights_mean)
+        self.weights_max = numpy.float64(self.weights_max)
+        self.tract_lengths_min = numpy.float64(self.tract_lengths_min)
+        self.tract_lengths_mean = numpy.float64(self.tract_lengths_mean)
+        self.tract_lengths_max = numpy.float64(self.tract_lengths_max)
```
liadomide: Have you considered including the cast inside the from_ndarray?
It seems that all these variables come from it.

framework_tvb/tvb/adapters/datatypes/db/spectral.py
```diff
@@ -50,9 +51,12 @@ def fill_from_has_traits(self, datatype):
         # type: (FourierSpectrum)  -> None
         super(FourierSpectrumIndex, self).fill_from_has_traits(datatype)
         self.segment_length = datatype.segment_length
+        self.segment_length = numpy.float64(self.segment_length)
```
liadomide: I would really want to have these casts in a single place, not spread all over the DB entities

liadomide:
At this point it is not clear to me what is the cause of our type mismatch. How do we get from float to numpy.int64? When is this happening, and what library is responsible for this cast?
So, the current coding solution might be ok, but at the same time might just hide a problem. Thus, we need to understand the root cause.

PR #37, opened 2020-02-10
TVB NEST serial

maedoc:
This is a work in progress to merge TVB NEST cosimulation support into TVB.

~~Tracked in JIRA here https://req.thevirtualbrain.org/browse/TVB-2611~~ This PR covers broad inclusion of cosim/multiscale support from @dionperd work in the the-virtual-brain/tvb-multiscale repository.


maedoc:
First steps here will be to factor the new models which make this PR diff hard to read.  @dionperd the models don't seem to change much besides the non-integrated parameters, is there a reason to copy-paste so mcuh?  I will take a closer look soon.

dionperd:
There are changes only to the:
a. models, which are added so that they don't modify the existing wong-wang, wilson_cowan, and generic_2d_oscillators models (therefore a lot of lines of change since they are new additional models),
b. simulator where the changes for non-state variables are also present together with any TVB-NEST related changes. 

Therefore, I think this branch should be compared against the non-state-variables branch, and not against the master.

As for the models, they currently contain extra parameters to flag whether the models should be simlated by TVB for a node or overwrite some of their (non-)state variables from input coming from NEST. Given that for the moment I don't want to affect the existing TVB models, I copied pasted them as new models.

Alternatively, one could inherit the existing ones and only the modify the parts in question (some parameters, and some logic in the dfun).

I can do this, if you wish.

dionperd:
OK. With these last 2 commits now the ReducedWongWangExcInh model inherits from TVB and only adds the params and logic that related to updating from NEST.

WilsonCowan and generic_2d_oscillator were already inheriting from TVB ones.

maedoc:
> Therefore, I think this branch should be compared against the non-state-variables branch, and not against the master.

Good point, we need the other one merged first.

scientific_library/tvb/simulator/simulator.py
```diff
@@ -50,16 +50,24 @@
 from .history import SparseHistory
 from tvb.basic.neotraits.api import HasTraits, Attr, NArray, List, Float
 
+from tvb.simulator.models.reduced_wong_wang_exc_io_inh_i import ReducedWongWangExcIOInhI
+from tvb_multiscale.config import CONFIGURED
```
liadomide: Adding this import, we have a cyclic dependency here (between tvb-library and tvb-multiscale), unhealthy.

scientific_library/tvb/simulator/simulator.py
```diff
@@ -50,16 +50,24 @@
...
+
 
 # TODO with refactor, this becomes more of a builder, since iterator will account for
 # most of the runtime associated with a simulation.
 class Simulator(HasTraits):
+
+    tvb_spikeNet_interface = None
```
liadomide: This `tvb_spikeNet_interface ` seems to be a complex object, missing from this PR

scientific_library/tvb/simulator/simulator.py
```diff
@@ -50,16 +50,24 @@
...
+    configure_spiking_simulator = None
+    run_spiking_simulator = None
+
     """A Simulator assembles components required to perform simulations."""
 
     connectivity = Attr(
         field_type=connectivity.Connectivity,
         label="Long-range connectivity",
-        default=None,
+        default=CONFIGURED.DEFAULT_SUBJECT["connectivity"],
```
liadomide: This could be left to None, and then perhaps we can easier remove the dependency towards  CONFIGURED

scientific_library/tvb/simulator/simulator.py
maedoc: Reading through that config module, it looks like some boilerplate.  The import here is a convenience to be removed, and those classes could move into tvb's core config as runtime configuration.  This could move into a PR to merge before this one.

scientific_library/tvb/simulator/simulator.py
maedoc: yep there is a rather large refactor to complete first.

scientific_library/tvb/simulator/simulator.py
```diff
@@ -363,6 +422,42 @@ def _loop_monitor_output(self, step, state):
...
+        node_coupling = numpy.zeros((history.shape[0], 1, history.shape[2], self.model.number_of_modes))
+        for i_time in range(history.shape[1]):
+            update_initial_conditions(history[:, i_time], node_coupling[:, 0], 0.0)
+        self.bound_and_clamp(history)
+
+    def update_state(self, state, node_coupling, local_coupling=0.0, use_numba=True):
+        # If there are non-state variables, they need to be updated for the initial condition:
+        try:
+            self.model.update_non_state_variables
+        except:
```
maedoc: What exception would be raised here?  A bare `except` is poor style.

maedoc:
hi @dionperd a second round of reading through after your recent commits.  It mostly looks good to me, though I would like to the tic/toc stuff refactored out.  do you have any docs on how to setup a working environment for this?  (preferably a `Dockerfile` or `Vagrantfile` :sweat_smile: )

liadomide:
> hi @dionperd a second round of reading through after your recent commits. It mostly looks good to me, though I would like to the tic/toc stuff refactored out. do you have any docs on how to setup a working environment for this? (preferably a `Dockerfile` or `Vagrantfile` 😅 )

Hi. I think you can use this Dockerfile https://github.com/the-virtual-brain/tvb-multiscale/blob/master/docker/Dockerfile. It still needs a bit of cleaning, but should be functional, and give you an jupyter notebook testbed with some examples in.

scientific_library/tvb/simulator/simulator.py
dionperd: I removed it.

scientific_library/tvb/simulator/simulator.py
dionperd: If this is None, it meens no multiscale simulation is going to take place, and it is just ignored.

scientific_library/tvb/simulator/simulator.py
dionperd: Done

scientific_library/tvb/simulator/simulator.py
dionperd: No exception is to be raised here.
The try ... except... can be substituted by 
if there is an "model.update_non_state_variables" method, run it,
else, do nothing.

framework_tvb/tvb/core/neocom/_h5loader.py
```diff
@@ -147,6 +147,8 @@ def store(self, datatype):
 
             if self.recursive:
                 sub_dt_refs = f.gather_references()
+                if self.recursive != "deep":  # usually we don't want to store twice the same references
```
liadomide, line 150: @dionperd could you pls separate these last commits and put them on a separate branch and PR?
I do not think we can efficiently merge all these together.

scientific_library/tvb/simulator/cosimulator.py
```diff
@@ -0,0 +1,268 @@
...
+#
+
+"""
+This is the module responsible for co-simulation of TVB with spiking simulators.
+It inherits the Simulator class.
+
+.. moduleauthor:: Stuart A. Knock <Stuart@tvb.invalid>
+.. moduleauthor:: Marmaduke Woodman <marmaduke.woodman@univ-amu.fr>
+.. moduleauthor:: Paula Sanz Leon <Paula@tvb.invalid>
+.. moduleauthor:: Dionysios Perdikis <dionysios.perdikis@charite.de>
```
liadomide, line 38: Maybe put the real author here would be good, for contribution acknowledgment, but also for servicing

scientific_library/tvb/simulator/models/generic_2d_oscillator_multiscale.py
```diff
@@ -0,0 +1,293 @@
...
+# When using The Virtual Brain for scientific publications, please cite it as follows:
+#
+#   Paula Sanz Leon, Stuart A. Knock, M. Marmaduke Woodman, Lia Domide,
+#   Jochen Mersmann, Anthony R. McIntosh, Viktor Jirsa (2013)
+#       The Virtual Brain: a simulator of primate brain network dynamics.
+#   Frontiers in Neuroinformatics (7:10. doi: 10.3389/fninf.2013.00010)
+
+"""
+Oscillator builders.
+
```
liadomide, line 31: Denis should be author here

scientific_library/tvb/simulator/simulator.py
```diff
@@ -253,29 +277,38 @@ def configure(self, full_configure=True):
             The configured Simulator instance.
 
         """
+        # TODO: temporary hack because numba dfuns fail for multiple modes
+        if self.use_numba and self.model.number_of_modes == 1:
```
liadomide, line 281: Shouldn't we better fix those numba dfuns, and not commit this `temporary hack` ?

dionperd: It is not that easy for me at least. I don't understand 100% how the numba arrays work.

But, even if we fix this, we cannot expect that all users who want to add models, or have their own models sitting somewhere, they know how to write numba code.

I think that for some time, we should keep the option for the user to choose the numpy or the numba funs, somehow.

Setting the spatial_reshape shape correctly would follow then, depending on whether we have written the numba funs for multiple modes or not.

Still, I don't know if this solution of having a use_numba flag is the best way to deal with this...

liadomide: I think there is already a mecanism to Avoid numba, by not inheriting numba based model class, and not Having a numba function.

scientific_library/tvb/simulator/simulator.py
```diff
@@ -425,8 +425,9 @@ def _update_and_bound_history(self, history):
 
     def update_state(self, state, node_coupling, local_coupling=0.0):
         # If there are non-state variables, they need to be updated for the initial condition:
-        state = \
-                self.model.update_non_state_variables(state, node_coupling, local_coupling, use_numba=self.use_numba)
+        if self.model._update_non_state_variables:
+            state = \
```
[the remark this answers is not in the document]
dionperd: Yes, this could be the case. The default implementation is to do nothing. In order to avoid a  redundant function call, I thought to have an if statement inside the loop that checks about whether there is need for any specific model to call this method...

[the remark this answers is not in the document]
dionperd: As for avoiding numba, I meant at runtime.

scientific_library/tvb/simulator/simulator.py
dionperd: As for avoiding numba, I meant at runtime.

scientific_library/tvb/simulator/simulator.py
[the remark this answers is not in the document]
dionperd: For the moment, I removed the redundant if statement (there is one already in the __call__ method

scientific_library/tvb/simulator/models/generic_2d_oscillator_multiscale.py
dionperd: I put my name where I think it should be!

scientific_library/tvb/simulator/simulator.py
```diff
@@ -278,12 +278,13 @@ def configure(self, full_configure=True):
...
         # TODO: temporary hack because numba dfuns fail for multiple modes
-        if self.use_numba and self.model.number_of_modes == 1:
-            self._dfun = self.model.dfun
-        else:
+        self._dfun = self.model.dfun
+        if not self.use_numba or self.model.number_of_modes > 1:
             self.use_numba = False
-            self._dfun = self.model._numpy_dfun
+            if hasattr(self.model, "_numpy_dfun"):
+                self._dfun = self.model._numpy_dfun
```
liadomide, line 285: This type of code is ugly, especially as we are having both `Model` and `ModelNumbaDfun` base classes very easy to use/switch.

tvb_contrib/tvb/contrib/scripts/datatypes/time_series.py
```diff
@@ -421,7 +431,10 @@ def sample_rate(self):
 
     @property
     def space_labels(self):
-         return numpy.array(self.labels_dimensions.get(self.labels_ordering[2], []))
+        try:
+            return numpy.array(self.get_space_labels())
+        except:
```
bvalean, line 436: What exception are we expecting here?

scientific_library/tvb/simulator/simulator.py
```diff
@@ -432,9 +432,9 @@ def _apply_spike_stimulus(self, step):
             setattr(self.model, target_parameter, self._spike_stimulus_fun(target_parameter, step-1))
 
     def update_non_state_variables(self, state, node_coupling, local_coupling):
-        if self.model._update_non_state_variables:
-            state = self.model.update_non_state_variables(state, node_coupling, local_coupling,
-                                                          use_numba=self.use_numba)
+        state = self.model.update_non_state_variables(state, node_coupling, local_coupling,
```
bvalean, line 435: Did these changes in order to remove the has attr checking from the model configure method.

tvb_contrib/tvb/contrib/scripts/datatypes/time_series.py
dionperd: I corrected this now. try except is not needed and there is no get_space_labels method anymore

scientific_library/tvb/simulator/simulator.py
```diff
@@ -557,7 +559,8 @@ def _configure_history(self, initial_conditions):
                 self.current_step += ic_shape[0] - 1
 
         # Make sure that history values are bounded
-        self.bound_and_clamp(history)
+        for it in range(history.shape[0]):
+            history[it] = self.bound_and_clamp(history[it])
```
bvalean, line 563: The bound_and_clamp method doesn't have a return statement so after this loop all history elements will be nan. There is also the test_propagation in history_test.py which is failing after this change. Call only the bound and clamp method without assigning the method's result to the history[it] because the history[it] values will be changed inside of this method.

bvalean:
> Wouldn't be better to create a separate PR for the Sensors and Projections ENUMS ?
> We could merge that faster in master, and then rebase here..
> BTW: good idea to introduce these enums. I had considered it in the past, but never found the courage and time to identify all usage

Yes it would. I will create another PR only with this change.

maedoc:
lots of conflicts here, is anyone working on resolving those now or should I get started?

maedoc:
merge conflicts should be resolved now

scientific_library/tvb/simulator/simulator.py
maedoc: I would agree type dispatch is better, though numba gufuncs support modes, it's really the dfun method wrapping it which assumes otherwise.  You rather need to think about which stage you're coupling the modes.

scientific_library/tvb/simulator/simulator.py
```diff
@@ -363,6 +399,46 @@ def _loop_monitor_output(self, step, state):
...
+        # If there is a state boundary...
+        if self.integrator.state_variable_boundaries is not None:
+            # ...use the integrator's bound_state
+            self.integrator.bound_state(state)
+        # If there is a state clamping...
+        if self.integrator.clamped_state_variable_values is not None:
+            # ...use the integrator's clamp_state
+            self.integrator.clamp_state(state)
+
+    def _print_progression_message(self, step, n_steps):
```
maedoc, line 412: I already asked for this to be refactored the progress monitor.  If you want to hardcode a default monitor in the cosimulator, that's fine, but no sys.stdout here please.

maedoc:
@dionperd a lot of this looks quite good now, interesting to see the modes abused for spiking neurons :)  but there are broad questions:

Making a cosimulator class subclassing the base simulator is a good strategy, but why do you then introduce a number of changes in the base class?  You largely rewrite `Simulator.__call__` and then have much the same code in `Cosimulator.__call__`.  I would prefer you leave the base `Simulator` alone and keep your high-level changes in `Cosimulator`.

It's a little hard to understand what's going on in `time_series_xarray.py`.  My guess is that it is a bunch of support code for your overall workflow, but it would be good to hear more about what's happening there.

maedoc:
Also @dionperd if you can start thinking how to setup short unit tests for the cosimulation, it would be helpful.  Since you already have a Dockerfile which produces a usable image, adding that here would be a good step, and I can get the tests running.

framework_tvb/tvb/adapters/simulator/monitor_forms.py
```diff
@@ -32,9 +32,11 @@
 from tvb.core.adapters.abcadapter import ABCAdapter
 from tvb.core.entities.file.simulator.view_model import *
 from tvb.core.entities.filters.chain import FilterChain
+from tvb.datatypes.projections import ProjectionsType
+from tvb.adapters.simulator.equation_forms import get_ui_name_to_monitor_equation_dict, HRFKernelEquation
 from tvb.core.neotraits.forms import Form, ScalarField, ArrayField, MultiSelectField, SelectField, \
     TraitDataTypeSelectField
-from tvb.datatypes.projections import ProjectionsType
+from tvb.basic.neotraits.api import List
```
liadomide, line 39: These changed imports seem odd here. Do we need this file changed?

scientific_library/tvb/datatypes/sensors.py
```diff
@@ -49,6 +49,7 @@ class SensorTypes(Enum):
     TYPE_EEG = "EEG"
     TYPE_MEG = "MEG"
     TYPE_INTERNAL = "Internal"
+    TYPE_SEEG = "SEEG"
```
liadomide, line 52: Do we really need this extra enum entry?

scientific_library/tvb/simulator/models/reduced_wong_wang_exc_io.py
```diff
@@ -0,0 +1,320 @@
...
+    else:
+        # TVB computation
+        dx[1] = 0.0
+    dx[0] = - (S[0] / t[0]) + (1.0 - S[0]) * S[2] * g[0]   # S
+    dx[2] = 0.0  # R
+    dx[3] = 0.0  # Rin
+    dx[4] = 0.0  # I
+
+
+class ReducedWongWangExcIO(TVBReducedWongWang):
```
liadomide, line 86: This new model, if remaining here should be included in the normal tvb-library unit-tests round (where all the other unit-tests are covered)

scientific_library/tvb/simulator/simulator.py
```diff
@@ -363,6 +373,16 @@ def _loop_monitor_output(self, step, state):
         if any(outputi is not None for outputi in output):
             return output
 
+    def bound_and_clamp(self, state):
```
liadomide, line 376: Could this clamping of the history values be separated and introduced into tvb-library as a separate PR with dedicated unit-tests ?
I think it is related to #13 and #6 Isn't it ?

framework_tvb/tvb/adapters/simulator/monitor_forms.py
bvalean: Yes, we need the _**List**_ import because it seems that in master this import does not exist but it is used. From what I can see there were some conflicts in the branch where this import was added and at rebase this import was missed.

liadomide: oK. i fix this import in master, as it is relevant for others also.

maedoc:
This has been superseded by the other PR recently merged.

PR #38, opened 2020-02-10
WIP dynamic non-state variables

maedoc:
These are elsewhere referred to as auxiliary variables, but here used to accommodate NEST I/O.  They are similar to what is already present as `variables_of_interest` which are used for monitoring arithmetic expressions of state variables, e.g. `x1 - x2` for Epileptor LFP, but weren't well documented.

Following also at https://req.thevirtualbrain.org/browse/TVB-2611


scientific_library/tvb/simulator/simulator.py
```diff
@@ -38,7 +38,7 @@
 .. moduleauthor:: Paula Sanz Leon <Paula@tvb.invalid>
 
 """
-
+import sys
```
maedoc: shouldn't need `sys` in this module

scientific_library/tvb/simulator/simulator.py
```diff
@@ -363,6 +363,38 @@ def _loop_monitor_output(self, step, state):
...
+            self.bound_and_clamp(history)
+        except:
+            pass
+
+    def update_state(self, state, node_coupling, local_coupling=0.0):
+        # If there are non-state variables, they need to be updated for the initial condition:
+        try:
+            self.model.update_non_state_variables(state, node_coupling, local_coupling)
+            self.bound_and_clamp(state)
+        except:
```
maedoc: bare except is not OK, should prefer to handle this more gracefully or obviate it entirely

scientific_library/tvb/simulator/simulator.py
```diff
@@ -376,9 +408,9 @@ def __call__(self, simulation_length=None, random_state=None):
 
         self.calls += 1
         if simulation_length is not None:
-            self.simulation_length = float(simulation_length)
+            self.simulation_length = simulation_length
```
maedoc: should keep the conversion to float here

scientific_library/tvb/simulator/simulator.py
```diff
@@ -387,20 +419,50 @@ def __call__(self, simulation_length=None, random_state=None):
...
+                    if toc > 7200:
+                        time_string = "%0.1f hours" % (toc / 3600)
+                    else:
+                        time_string = "%0.1f min" % (toc / 60)
+                else:
+                    time_string = "%0.1f sec" % toc
+                print_this = "\r...%0.1f%% done in %s" % \
+                             (100.0 * (step - self.current_step) / n_steps, time_string)
+                sys.stdout.write(print_this)
+                sys.stdout.flush()
```
maedoc: This would be good to integrate into the [ProgressLogger](https://github.com/the-virtual-brain/tvb-root/blob/master/scientific_library/tvb/simulator/monitors.py#L911) monitor maybe taking cues from [tqdm](https://github.com/tqdm/tqdm)? but not here

scientific_library/tvb/simulator/simulator.py
```diff
@@ -387,20 +419,50 @@ def __call__(self, simulation_length=None, random_state=None):
...
+                tic_point += tic_ratio * n_steps
 
         self.current_state = state
-        self.current_step = self.current_step + n_steps
+        self.current_step = self.current_step + n_steps - 1  # -1 : don't repeat last point
```
maedoc: this is wrong

scientific_library/tvb/simulator/simulator.py
```diff
@@ -450,13 +512,15 @@ def _configure_history(self, initial_conditions):
                     history = numpy.roll(history, shift, axis=0)
                 self.current_step += ic_shape[0] - 1
 
-        if self.integrator.state_variable_boundaries is not None:
-            self.integrator.bound_state(numpy.swapaxes(history, 0, 1))
+        # Make sure that history values are bounded,
+        # and any possible non-state variables are initialized
+        # based on state variable ones (but with no coupling yet...)
+        self._update_and_bound_history(numpy.swapaxes(history, 0, 1))
```
maedoc: `numpy.swapaxes` should return a view but its version dependent, so I'd prefer to avoid it. Why can't this method operate directly on the history buffer as-is?

scientific_library/tvb/simulator/simulator.py
```diff
@@ -363,6 +363,38 @@ def _loop_monitor_output(self, step, state):
         if any(outputi is not None for outputi in output):
             return output
 
+    def bound_and_clamp(self, state):
```
maedoc: These operations should be done via code-generation (i.e. numba) to avoid this sort of dynamism at runtime.  The bounding & clamping isn't changing during a simulation.

maedoc:
@dionperd looks like a good start to me.  I will review it more and come back with some more comments.  I think the you non-state vars can be merged with the existing "variables of interest" in the current TVB.

scientific_library/tvb/simulator/simulator.py
dionperd: Yes. It is needed only for the progression messages.

scientific_library/tvb/simulator/simulator.py
dionperd: I just put it there so that you see it and move it somewhere else. It is something lacking currently.

scientific_library/tvb/simulator/simulator.py
dionperd: I don't remember removing it. Probably I just copied-pasted an older version.

scientific_library/tvb/simulator/simulator.py
dionperd: Again, I don't think it is my making. How far in the past is this an error? Does this mean that we used to simulate one step less?

scientific_library/tvb/simulator/simulator.py
dionperd: The shape is not correct for the state otherwise. We could also loop across time and modes also...

dionperd:
I am not sure they can be merged with the "variables of interest". First of all, they have to be computed for EVERY time step, unlike monitors that might have a different resolution. Second, they have to be computed several times for integration schemes of order >1..

The computation of these variables require dfun to be broken into two parts:
dfun that computes the derivative based on state variables and coupling at time t, and update_fun that computes non-state variables based on state variables and coupling, also at time t.
Usually dfun would also call the update_fun.

In other words, the logic is 
state(t), delayed_coupling(t) ---update_fun---> non-state(t) -----dfun----> dstate(t) ------> 
state(t+dt), delayed_coupling(t+dt) ---update_fun---> non-state(t+dt) -----dfun---->dstate(t+dt) -----> 
state(t+2dt),, delayed_coupling(t+2dt) etc

The above loop runs e.g., 4 times for Runge-Kutta 4th order considering delayed_coupling(t) as fixed, but non-state variables changing, since they would anyway be computed inside dfun.

Instead, whatever monitor is applied comes after integration of time step t -> t+dt is concluded.

The try... except... above tries to call such an update_fun, if it exists in the model. If not, it just goes on without doing anything.

scientific_library/tvb/simulator/simulator.py
maedoc: you can ask GitHub or pycharm for the "git blame" to see who changed those lines.

maedoc:
You are right about computing these at every dfun evaluation.  However, I do not think that precludes a general notion of auxiliary variables which are computing before, during or after dfun and then used through coupling or monitors.

scientific_library/tvb/simulator/models/reduced_wong_wang_exc_io_inh_i.py
```diff
@@ -0,0 +1,357 @@
+# -*- coding: utf-8 -*-
```
liadomide: Could this be merged/linked somehow with wong_wang_exc_inh.ReducedWongWangExcInh model class, or should it stay as a complete separate implementation ? They seem to have a lot of common code

dionperd: Sure. I have done so already in the tvb-mulstiscale-cosim branch. Not needed to repeat here since anyway tvb-mulstiscale-cosim should be merged at some point to this branch, after this branch is merged to the master.

scientific_library/tvb/simulator/models/reduced_wong_wang_exc_io_inh_i.py
```diff
@@ -0,0 +1,357 @@
...
+                 x_{ek}       &=   w_p\,J_N \, S_{ek} - J_iS_{ik} + W_eI_o + GJ_N \mathbf\Gamma(S_{ek}, S_{ej}, u_{kj}),\\
+                 H(x_{ek})    &=  \dfrac{a_ex_{ek}- b_e}{1 - \exp(-d_e(a_ex_{ek} -b_e))},\\
+                 \dot{S}_{ek} &= -\dfrac{S_{ek}}{\tau_e} + (1 - S_{ek}) \, \gammaH(x_{ek}) \,
+
+                 x_{ik}       &=   J_N \, S_{ek} - S_{ik} + W_iI_o + \lambdaGJ_N \mathbf\Gamma(S_{ik}, S_{ej}, u_{kj}),\\
+                 H(x_{ik})    &=  \dfrac{a_ix_{ik} - b_i}{1 - \exp(-d_i(a_ix_{ik} -b_i))},\\
+                 \dot{S}_{ik} &= -\dfrac{S_{ik}}{\tau_i} + \gamma_iH(x_{ik}) \,
+
+    """
+    _ui_name = "Reduced Wong-Wang"
```
liadomide: no ui_* attributes should exist in tvb-library 2.0

maedoc:
@dionperd I think there are some changes you add to the cosim branch which aren't update here? can you confirm? since we want to merge this first, I want to make sure it's all up to date here first.  Then I will try to write some basic tests checking the logic as you stated above

> In other words, the logic is
state(t), delayed_coupling(t) ---update_fun---> non-state(t) -----dfun----> dstate(t) ------>
state(t+dt), delayed_coupling(t+dt) ---update_fun---> non-state(t+dt) -----dfun---->dstate(t+dt) ----->
state(t+2dt),, delayed_coupling(t+2dt) etc

scientific_library/tvb/simulator/models/reduced_wong_wang_exc_io_inh_i.py
dionperd: Done. Removed.

dionperd:
Yes. There were additional changes in the tvb-multiscale-cosim branch, which I have just transferred to this branch as well. However, this is not good for me since I have to work on the other branch and keep updating this one too (I guess there is some git - solution to this problem...). 

We lack right now a way to formalise all use cases of such dynamic non-state variables.

Should they be updated only before (as it is now) or after the large-scale coupling computation?

If we have both options, should the user define two separate functions for each different update?
etc...

maedoc:
> git - solution to this problem

if these are commits that you transfer, yes; if you are copying pasting then let's just try to merge this quickly.. 

> We lack right now a way to formalise all use cases of such dynamic non-state variables.

true I will think about this while testing the code

PR #40, opened 2020-02-12
TVB-2587 Solved: Added support for uploaders to launch_operation.

VinczeRobert:
Discussed with Paula before pushing these. Please let me know if there is something that could be done in a better way.

### DIFF
--- a/framework_tvb/tvb/adapters/uploaders/brco_importer.py
+++ b/framework_tvb/tvb/adapters/uploaders/brco_importer.py
@@ -56,13 +56,21 @@ class BRCOImporterModel(UploaderViewModel):
         doc='The Connectivity for which these annotations were made'
     )
 
+    @staticmethod
+    def get_files_types():
+        return ['.xml']
+
+    @staticmethod
+    def get_upload_files_names():
+        return ['data_file']
+
 
 class BRCOImporterForm(ABCUploaderForm):
 
     def __init__(self, prefix='', project_id=None):
         super(BRCOImporterForm, self).__init__(prefix, project_id)
 
-        self.data_file = TraitUploadField(BRCOImporterModel.data_file, '.xml', self, name='data_file')
+        self.data_file = TraitUploadField(BRCOImporterModel.data_file, self.get_view_model().get_files_types()[0], self, name='data_file')
         self.connectivity = TraitDataTypeSelectField(BRCOImporterModel.connectivity, self, name='connectivity')
 
     @staticmethod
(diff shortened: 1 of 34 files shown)


framework_tvb/tvb/adapters/uploaders/brco_importer.py
```diff
@@ -56,13 +56,21 @@ class BRCOImporterModel(UploaderViewModel):
         doc='The Connectivity for which these annotations were made'
     )
 
+    @staticmethod
+    def get_files_types():
+        return ['.xml']
```
liadomide: why return a list and always use it with .get_files_types()[0] ?

framework_tvb/tvb/interfaces/rest/server/resources/operation/operation_resource.py
```diff
@@ -117,10 +124,20 @@ def post(self, project_gid, algorithm_module, algorithm_classname):
                                                                  {})
             storage_path = self.files_helper.get_project_folder(project, str(operation.id))
 
+            if isinstance(adapter_instance, ABCUploader):
+                data_file_1 = self.extract_file_from_request(file_name='data_file_1', file_extension=view_model.get_files_types()[0])
+                handle_data_file(data_file_1, destination_folder, view_model_h5, storage_path, 0)
+
+                if isinstance(adapter_instance, (CSVConnectivityImporter, NIFTIImporter)) or \
```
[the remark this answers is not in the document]
VinczeRobert: > Is there any unit-test for these changes?
> Then we could easily refactor it

There is no unit test yet, but I wrote an example similar to the fire_simulation one, where I uploaded a connectivity, a surface, a region mapping and a connectivity csv and then download them to see if they where actually uploaded.

framework_tvb/tvb/interfaces/rest/server/resources/user/user_resource.py
```diff
@@ -40,7 +40,7 @@ def get(self):
         """
         :return a list of system's users
         """
-        users = UserService.fetch_all_users()
+        users = UserService.fetch_all_users(page_size=1000)
```
liadomide: This constant should be defined some place else

framework_tvb/tvb/interfaces/rest/server/resources/util.py
```diff
@@ -29,16 +29,33 @@
...
 
 import os
+import shutil
 import tempfile
 from tvb.basic.profile import TvbProfile
 from werkzeug.utils import secure_filename
 
 
-def save_temporary_file(file):
+def save_temporary_file(file, destination_folder):
```
liadomide: This `utils.py` module is not very nice and OOP-friendly. Can we consider putting these functions in another place? e.g. in the base `RestResource` class?

PR #43, opened 2020-02-13
WIP Auto-gradients

maedoc:
This is some initial support for automatic gradients with autograd.  We will use a hybrid approach with this for generic numpy stuff & the gradients2 branch for models.


maedoc:
This should get folded into a backend as part of the sim-refactor branch.

maedoc:
#303

maedoc:
We will re-evaluate this after working out the main backends in as part of TVB-2600.

PR #45, opened 2020-02-20
TVB-2619 DT Overlay not showing in web GUI when Adapter has errors in its filter

adrian-dordea:
Add default on DataTypeMatrix.ndim field 
Add warning to be displayed in the web GUI

### DIFF
--- a/framework_tvb/tvb/core/services/flow_service.py
+++ b/framework_tvb/tvb/core/services/flow_service.py
@@ -280,6 +280,9 @@ class FlowService:
             if not filter_chain or filter_chain.get_python_filter_equivalent(datatype):
                 filtered_adapters.append(stored_adapter)
 
+        if filtered_adapters.__len__() == 0:
+            raise OperationException("There are no valid adapters")
+
         return datatype, filtered_adapters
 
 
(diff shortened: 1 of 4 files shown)


framework_tvb/tvb/core/services/flow_service.py
```diff
@@ -280,6 +280,9 @@ def get_launchable_algorithms_for_datatype(self, datatype, categories):
             if not filter_chain or filter_chain.get_python_filter_equivalent(datatype):
                 filtered_adapters.append(stored_adapter)
 
+        if filtered_adapters.__len__() == 0:
+            raise OperationException("There are no valid adapters")
+
```
liadomide: I do not think this exception should be thrown. It is possible that we have no operation compatible with the current DT

framework_tvb/tvb/interfaces/web/controllers/project/project_controller.py
```diff
@@ -338,8 +338,12 @@ def get_datatype_details(self, entity_gid, back_page='null', exclude_tabs=None):
         current_type = datatype_details.data_type
         datatype_gid = datatype_details.gid
         categories = {}
+        no_visualisers = None
         if not entity.invalid:
-            categories = self.flow_service.get_launchable_algorithms(datatype_gid)
+            try:
+                categories = self.flow_service.get_launchable_algorithms(datatype_gid)
+            except Exception as excep:
+                no_visualisers = 'Not all operations could be loaded for this input DataType. Contact the admin to check the logs!'
```
liadomide: Now we actually catch too late the exception, thus we will never show any viewer, when we should only omit the ones with problems

PR #48, opened 2020-02-21
WIP TVB-2620 Secure REST api using Keycloak (OpenID Connect)

bvalean:


framework_tvb/tvb/interfaces/rest/client/main_api.py
```diff
@@ -27,19 +27,35 @@
...
-    def __init__(self, server_url):
+    def __init__(self, server_url, auth_token=''):
+        # type: (str, str) -> None
         """
         Rest server url, where all rest calls will be made
+        :param server_url: REST server base URL
+        :param auth_token: Keycloak authorization token
         """
         self.server_url = server_url + "/" + Strings.BASE_PATH.value
+        self.authorization_token = auth_token
```
liadomide: What is happening with long running operation?
Where can we check /configure the token attributes (expiration date) ?

framework_tvb/tvb/interfaces/rest/client/user/user_api.py
```diff
@@ -28,23 +28,36 @@
...
-        return response, UserDto
+    def login(self, username, password):
+        response = self.secured_request().post(self.build_request_url(RestLink.LOGIN.compute_url(True)), json={
+            "username": username,
+            "password": password
+        })
+        return response
+
+    @handle_response
+    def refresh(self, refresh_token):
```
liadomide: Could you document on this method when and how should be used?
is BrainX3 expected to call it, or we do it automatically?

bvalean: I'm refactoring this part so refreshing token will be done automatically before a request if it is expired

framework_tvb/tvb/interfaces/rest/client/main_api.py
bvalean: From keycloak server we receive also the number of seconds left before the access token expires.

liadomide: then it is configured in Keycloak, right?
What is the current/default ?

bvalean: Yes, it is configured in keycloak. Default value is 5 minutes

framework_tvb/tvb/interfaces/rest/client/examples/launch_operation.py
```diff
@@ -125,26 +120,33 @@
...
+    logger.info(
+        "Preparing a connectivity csv H5 file (this one requires two datatype files: one containing weights and the other one containing tracts)...")
     csv_view_model = CSVConnectivityImporterModel()
 
-    csv_weights_path = os.path.join(os.path.dirname(tvb_data.__file__), 'dti_pipeline_toronto', 'output_ConnectionCapacityMatrix.csv')
-    csv_tracts_path = os.path.join(os.path.dirname(tvb_data.__file__), 'dti_pipeline_toronto', 'output_ConnectionDistanceMatrix.csv')
+    csv_weights_path = os.path.join(os.path.dirname(tvb_data.__file__), 'dti_pipeline_toronto',
+                                    'output_ConnectionCapacityMatrix.csv')
+    csv_tracts_path = os.path.join(os.path.dirname(tvb_data.__file__), 'dti_pipeline_toronto',
+                                   'output_ConnectionDistanceMatrix.csv')
```
liadomide: This example has become too complex and very difficult to read.
What is the purpose of enumerating all these importers?
If we need all for some reason, at least we should create separate functions for each of the calls

framework_tvb/tvb/interfaces/rest/client/main_api.py
```diff
@@ -27,19 +27,64 @@
...
+
+    def secured_request(self):
+        """
+        If we have an expiration date and the token is expired we make a refresh call then we build a secured request
+        based on the refreshed token.
+        :return: secured request session
+        """
+        if self.token_expiry_date is not None and datetime.now() >= self.token_expiry_date \
+                and self.refresh_token is not None:
+            refresh_token_response = self._refresh_token()
```
liadomide: This refresh seems welcome to have.

framework_tvb/tvb/interfaces/rest/client/user/user_api.py
liadomide: I see this had been done, and it is happening as long as one of out TVB Rest API methods is called. I think this is the only appropriate way to handle this.

PR #49, opened 2020-02-24
TVB-2591 completed! (a proposed solution for TVB-2546 is also here)

VinczeRobert:
I added using_template as a decorator in flow_controller, thus the wrong dependency in abc_adapter is no more! Also, I changed the names from 'form' to 'adapter_form' as it is more suggestive.


framework_tvb/tvb/adapters/creators/local_connectivity_creator.py
```diff
@@ -68,7 +68,7 @@ def get_filters():
 
     @using_template('spatial/spatial_fragment')
```
liadomide: Here the same wrong dependency was maintained. 
tvb.interfaces should depend on adapters, not viceversa.
See tvb-root/framework_tvb/README.rst for the full tree of dependencies

framework_tvb/tvb/adapters/creators/stimulus_creator.py
```diff
@@ -60,7 +60,7 @@ def __init__(self, project_id):
 
     @using_template('spatial/spatial_fragment')
     def __str__(self):
-        return {'form': self, 'legend': 'Loaded stimulus'}
+        return {'adapter_form': self, 'legend': 'Loaded stimulus'}
```
liadomide: adapters module should not depend on interface

framework_tvb/tvb/interfaces/web/controllers/spatial/local_connectivity_controller.py
```diff
@@ -140,7 +140,7 @@ def set_equation(self, equation):
         current_lconn.equation = eq_class()
 
         eq_params_form = get_form_for_equation(eq_class)(prefix=LocalConnectivityCreatorForm.NAME_EQUATION_PARAMS_DIV)
-        return {'form': eq_params_form, 'equationsPrefixes': self.plotted_equation_prefixes}
+        return {'adapter_form': eq_params_form, 'equationsPrefixes': self.plotted_equation_prefixes}
```
liadomide: Here you should repeat the solution for the generic form, and have an own controller method to fill a jinja template from the form, instead of relying on __str__

framework_tvb/tvb/interfaces/web/controllers/spatial/region_stimulus_controller.py
```diff
@@ -117,7 +117,7 @@ def set_temporal(self, temporal_equation):
 
         eq_params_form = get_form_for_equation(eq_class)(prefix=RegionStimulusCreatorForm.NAME_TEMPORAL_PARAMS_DIV)
         #TODO: check eqPrefixes
-        return {'form': eq_params_form, 'equationsPrefixes': self.plotted_equation_prefixes}
+        return {'adapter_form': eq_params_form, 'equationsPrefixes': self.plotted_equation_prefixes}
```
liadomide: Or put it into the base class... if possible

PR #50, opened 2020-02-24
TVB-2531 - Neotraits compatibility with PostgreSQL  

gabiflorea10:


framework_tvb/tvb/tests/framework/core/neotraits/db_test.py
```diff
@@ -64,7 +64,7 @@ class FooIndex(HasTraitsIndex):
 
     array_int_max = Column(Integer)
     array_int_min = Column(Integer)
-    array_int_mean = Column(Integer)
+    array_int_mean = Column(Float)
```
liadomide: It makes sense. Shouldn't we also have the `ensure_float` where this field is computed ?

gabiflorea10: I think it is not necessary, because according to numpy.mean official documentation ( https://docs.scipy.org/doc/numpy/reference/generated/numpy.mean.html ), the result type for integer inputs is float64.
Yet, if it is preferred, we can add this ensure_float in the place where this field is computed, for extra safety.

gabiflorea10:
> Can you reproduce the test failures as it is currently happening on Jenkins?

Yes, I can.

framework_tvb/tvb/tests/framework/core/neotraits/db_test.py
liadomide: Good. Then, it is not necessary to put `ensure_float`. Leave it as you initially proposed.

liadomide:
> > Can you reproduce the test failures as it is currently happening on Jenkins?
> 
> Yes, I can.

And, is the current change enough to fix the unit-tests? or is just one step forward ?

gabiflorea10:
> > > Can you reproduce the test failures as it is currently happening on Jenkins?
> > 
> > 
> > Yes, I can.
> 
> And, is the current change enough to fix the unit-tests? or is just one step forward ?

This is just one step forward. Complete fix will be available here today.

PR #51, opened 2020-02-24
TVB-2594 (part) - Fix how the errors from forms are reported

gabiflorea10:


framework_tvb/tvb/interfaces/rest/server/run.py
```diff
@@ -78,6 +79,7 @@ def initialize_flask():
     # creating the flask app
     app = Flask(__name__)
     app.json_encoder = CustomFlaskEncoder
+    app.jinja_env.add_extension(formencode_jinja2.formfill)
```
liadomide: I do not think for the rest api formencode-jinja will be really usefull.
I was mostly thinking of Cherrypy server when serving Jinja templates.

gabiflorea10: Ok. I revert configurations in this file.

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
```diff
@@ -489,7 +489,7 @@ def execute_post(self, project_id, submit_url, step_key, algorithm, **data):
             common.set_important_message(str(result))
 
         except formencode.Invalid as excep:
-            errors = excep.unpack_errors()
+            errors = excep.unpack_errors
```
liadomide: I do not think this solves the problem. Look at the printed message.
Like this, we say that variable `errors` is the reference towards a function. It might "work", meaning that it prints something and does not throw an error, like it did before, but it does not say what is the 'required' missing attribute (or whatever the specific exception might be in the form validation).

gabiflorea10: The printed message looks like that:
2020-02-26 14:05:20,413 - WARNING - tvb.interfaces.web.controllers.flow_controller - Invalid form inputs <bound method Invalid.unpack_errors of Invalid('Could not fill algorithm from the given inputs!', {}, None, None, {'_time_series': ValueError('Field required'), '_dt': ValueError('Field required'), '_neural_input_transformation': ValueError('Field required'), '_bold_model': ValueError('Field required')})>

liadomide: yep. I am right, it is the __str__ on the function `<bound method .... of ...[class]...`
It shows un indeed the problem, but it is a hack

framework_tvb/tvb/interfaces/web/controllers/exceptions.py
```diff
@@ -0,0 +1,53 @@
...
+#
+#
+"""
+Exceptions for the web interface.
+"""
+
+from tvb.basic.exceptions import TVBException
+
+
+class InvalidFormValues(TVBException):
```
liadomide: tvb.interfaces.web.controllers.common.py has other exceptions for controllers. 
Add this new class there also ?

framework_tvb/tvb/interfaces/web/controllers/exceptions.py
```diff
@@ -0,0 +1,53 @@
...
+    def __init__(self, message, error_dict=None):
+        TVBException.__init__(self, message)
+        self.error_dict = error_dict
+
+    def display_full_errors(self):
+        if self.error_dict:
+            result = {}
+            for name, item in self.error_dict.items():
+                result[name] = str(item)
+            return result
```
liadomide: We should display both `self.message` and the concatenated `error_dict.`
Maybe put every entry in the dict on a separate line, and show both the key and the value from the dictionary

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
```diff
@@ -463,8 +464,8 @@ def execute_post(self, project_id, submit_url, step_key, algorithm, **data):
                     raise formencode.Invalid("Could not find a model for this form!", {}, None,
                                              error_dict=form.get_errors_dict())
```
liadomide: Here we should also use the new `InvalidFormValues `

framework_tvb/tvb/interfaces/web/controllers/exceptions.py
liadomide: We should display both `self.message` and the concatenated `error_dict.`
Maybe put every entry in the dict on a separate line, and show both the key and the value from the dictionary

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
```diff
@@ -495,6 +496,10 @@ def execute_post(self, project_id, submit_url, step_key, algorithm, **data):
         except OperationException as excep1:
             self.logger.exception("Error while executing a Launch procedure:" + excep1.message)
             common.set_error_message(excep1.message)
+        except InvalidFormValues as excep2:
+            errors = excep2.display_full_errors()
+            common.set_error_message("Invalid form inputs")
```
liadomide: Here maybe use directly the `InvalidFormValues .message`

framework_tvb/tvb/interfaces/web/controllers/exceptions.py
gabiflorea10: Yes

PR #55, opened 2020-03-03
TVB-2631 Support authorization also using a keycloak server 

bvalean:
Keycloak  & web GUI integration


framework_tvb/tvb/interfaces/web/static/js/keycloak.js
```diff
@@ -0,0 +1,1686 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates
```
liadomide: We should add this a a 3rd party dependency in TVB_distribution and copy its original license next to the others

framework_tvb/tvb/interfaces/web/controllers/users_controller.py
```diff
@@ -36,27 +36,29 @@
...
-from tvb.core.services.project_service import ProjectService
+from tvb.core.entities.file.files_update_manager import FilesUpdateManager
 from tvb.core.services.exceptions import UsernameException
+from tvb.core.services.project_service import ProjectService
+from tvb.core.services.texture_to_json import color_texture_to_list
+from tvb.core.services.user_service import UserService, KEY_PASSWORD, KEY_EMAIL, KEY_USERNAME, KEY_COMMENT, \
+    KEY_AUTH_TOKEN
 from tvb.core.utils import format_bytes_human, hash_password
-import tvb.interfaces.web
+from tvb.interfaces.rest.server.security.authorization import AuthorizationManager
```
liadomide: This dependency of tvb.interface.web on tvb.interfaces.rest is odd.
We should avoid it. Can we move the AuthorizationManager down under tvb.core? Maybe even tvb.core.services?

framework_tvb/tvb/interfaces/web/controllers/users_controller.py
```diff
@@ -146,14 +154,18 @@ def profile(self, logout=False, save=False, **data):
...
             user = self.user_service.get_user_by_id(user.id)
             common.add2session(common.KEY_USER, user)
 
         template_specification['user_used_disk_human'] = format_bytes_human(
             self.user_service.compute_user_generated_disk_size(user.id))
         return self.fill_default_attributes(template_specification)
 
+    @cherrypy.expose
+    @using_template('user/silent_check_sso')
+    def check_sso(self):
```
liadomide: What exactly will this check do?
Will we automatically be logged in TVB after integration with HBP Collab, if the current person is already logged in HBP Collab, for example?

framework_tvb/tvb/interfaces/web/static/js/keycloak.js
liadomide: Adding the original license file and one entry in packages_accepted.xml is generally enough for our dependencies.
But for the JS libraries, we do not have the "smart" auto detect feature, as we currently do with Python dependencies. So, you need to explicitly list keycloack JS into the EXTRA_MODULES dict (line 100 in tvb_build.third_party_licenses.package_finder.py

PR #57, opened 2020-03-04
Fixed most of the rest tests and deleted __str__ with using_template.

VinczeRobert:
About the tests:

1. I added a fixture which creates a flask app and used it with test_context to solve the out of context problem which appeared in several tests.
2. test_server_get_operations_for_datatype failes because result has only 2 elements and it should have more than 3 
3. If you debug test_server_fire_simulation it says that the path is not a zip file even though to me it seems like it is. The file actually exists at the specified path and it has .zip extension.
4. test_get_projects doesn't fail if you run it separately. (in group it says that it doesn't have permission to do the teardown method)
5. I deleted the test test_get_project_invalid_username because the retrieving of projects is not done by username anymore. We should probably write a test that replaces this.
6. test_get_users says that it has empty keycloak config path.


framework_tvb/tvb/core/neotraits/forms.py
```diff
@@ -747,7 +747,7 @@ def validate(self):
         return self.form.validate()
 
     def __str__(self):
-        return jinja_env.get_template(self.template).render(form=self.form)
+        return jinja_env.get_template(self.template).render(adapter_form=self.form)
```
liadomide: This is better hidden, but somehow the same backwards dependency that we tried to get rid in your previous PR. I will mark with a TODO for later

PR #58, opened 2020-03-04
Dsl

DeLaVlag:
This PR add the DSL functionality that generates python TVB models according to an extended version of the LEMS library. This PR also includes documentation and a demo. Tests to validate will be added soon.


maedoc:
It seems like a lot of this PR is "just" a copy of the PyLEMS library vendored in.  Is there a rationale for this? If so we need perhaps to merge PyLEMS itself in at a particular point in their project history along with a license.

scientific_library/tvb/dsl/LEMS2python.py
```diff
@@ -0,0 +1,68 @@
+from mako.template import Template
+import tvb
+import sys
+import os
+sys.path.append("{}{}".format(os.path.dirname(tvb.__file__),"/dsl/NeuroML/lems"))
```
maedoc: what is happening here? you need to import code from that folder?

DeLaVlag:
The DSL is based on an adapted version of LEMS locally added to the PR. Its not just a copy. Simply merging it in is not going to work. Alterations were made on datastructures, parsing and the naming is fitted to TVB models.

scientific_library/tvb/dsl/LEMS2python.py
DeLaVlag: yes, this is the local altered copy of LEMS

maedoc:
> adapted version of LEMS locally added to the PR

ok can you just mention the LEMS version or Git commit you got it from so I can check the license and eventually later update the base LEMS version if required?

DeLaVlag:
Its version 0.4.9.3. and commit 4eeb719d2f23650fe16c38626663b69b5c83818b

maedoc:
Thanks so the [license there is LGPL](https://github.com/LEMS/pylems/blob/4eeb719d2f23650fe16c38626663b69b5c83818b/LICENSE.lesser)

maedoc:
with #62 merged we should be OK to proceed @DeLaVlag @sdiazpier thanks!

maedoc:
@DeLaVlag can we close this PR now or there is outstanding commits to keep/merge?

DeLaVlag:
@maedoc no outstanding commits, thanks!

PR #59, opened 2020-03-05
WIP TVB-2634 Fix gain order

maedoc:
[TVB-2634](https://req.thevirtualbrain.org/browse/TVB-2634)

### DIFF
--- a/scientific_library/pytest.ini
+++ b/scientific_library/pytest.ini
@@ -10,3 +10,5 @@ markers =
 
 testpaths =
     tvb/tests/library
+
+looponfailroots = tvb/tests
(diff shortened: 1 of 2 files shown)


maedoc:
@i-Zaak I can't reproduce the problem with config, can you check the test & specify a little more what problem you saw?

maedoc:
welp, I shoulda run the full test suite before pushing..

i-Zaak:
The condition before reducing the lead field for regional sim is not correct. It assumes that the region mapping is of the same size as the cotex-only gain from the projection.

https://github.com/the-virtual-brain/tvb-root/blob/504b4169cc46d71500347a05098d69f7b0954f57/scientific_library/tvb/simulator/monitors.py#L476

i-Zaak:
When that is fixed, we actually enter the bug, where the subcortical regions are added on top, because this is the only place, where the gain is initialized for regional simulation with surface-based projection.

PR #60, opened 2020-03-05
WIP TVB-2626

VinczeRobert:
So far I got rid of some if's by replacing them with methods for the sensor monitors that specify which class is needed for which sensors and surface type.


liadomide:
We should not modify `monitors.py`
To me it seems that the information in method `get_sensors_and_projection_surface_classes` is redundant anyway. Can't we get these types from Monitor.projection field directly?

PR #62, opened 2020-03-06
WIP LEMS DSL

maedoc:
hi @DeLaVlag I have rebase your PR on TVB master (which required fixing a silent merge non-conflict which made for a bug) and massaged the modules a bit so there's no unwiedly path manipulation, as well as add unit tests for loading and rendering models.
If you are OK with me managing the PR and merging then let's move forward this way.  Otherwise, I can PR to your branch and then you need to update your PR to the main repo.  What do you think?


DeLaVlag:
Hello @maedoc. No comments on the changes. Please proceed merging the PR.

maedoc:
I'm still thinking about user workflow for pulling in XML files, so I don't want to complete the functionality in the `models/__init__.py`.  Instead, let's go ahead with this merge, so that we can do the next PR on CUDA support.

tvb_documentation/demos/model_generation_using_dsl.ipynb
```diff
@@ -0,0 +1,285 @@
...
+    "## First let us load the necessary functions to generate code"
+   ]
+  },
+  {
+   "cell_type": "code",
+<<<<<<< Updated upstream
+   "execution_count": 16,
+=======
+   "execution_count": 4,
+>>>>>>> Stashed changes
```
liadomide, line 21: This notebook can not be opened like this `not a valid JSON` (due to merging errors, I think).

Then, the content has untested code: `self is not defined`, variable `model` does not exist....!

Also, function `regTVB_templating` can not work in case tvb-framework is also installed in the same env as tvb-library.

I fixed as much as I could in this commit: https://github.com/the-virtual-brain/tvb-root/pull/211/commits/db44239d9152315a38269d19bcc03c977ad9252a

PR #65, opened 2020-03-09
TVB-2624 Improve SimulatorAdapter area

VinczeRobert:
I have successfully merged the simulation length selection step with the simulation launching / pse setup and all the button changes related to PSE are now displayed correctly as well (that means that all next, previous etc. button are displayed and hidden correctly according to the rendering rules).

### DIFF
--- a/framework_tvb/tvb/adapters/simulator/simulator_fragments.py
+++ b/framework_tvb/tvb/adapters/simulator/simulator_fragments.py
@@ -159,18 +159,12 @@ class SimulatorMonitorFragment(ABCAdapterForm):
         self.monitor.data = trait.monitors[0].__class__
 
 
-class SimulatorLengthFragment(ABCAdapterForm):
-
-    def __init__(self, prefix='', project_id=None):
-        super(SimulatorLengthFragment, self).__init__(prefix, project_id)
-        self.length = ScalarField(Simulator.simulation_length, self)
-
-
 class SimulatorFinalFragment(ABCAdapterForm):
 
     def __init__(self, prefix='', project_id=None, simulation_number=1):
         super(SimulatorFinalFragment, self).__init__(prefix, project_id)
         default_simulation_name = "simulation_" + str(simulation_number)
+        self.length = ScalarField(Simulator.simulation_length, self)
         self.simulation_name = ScalarField(Attr(str, doc='Name for the current simulation configuration', default=default_simulation_name,
                                                 label='Simulation name'), self, name='input-simulation-name-id')
 
--- a/framework_tvb/tvb/adapters/simulator/simulator_fragments.py
+++ b/framework_tvb/tvb/adapters/simulator/simulator_fragments.py
@@ -166,7 +166,7 @@ class SimulatorFinalFragment(ABCAdapterForm):
         default_simulation_name = "simulation_" + str(simulation_number)
         self.length = ScalarField(Simulator.simulation_length, self)
         self.simulation_name = ScalarField(Attr(str, doc='Name for the current simulation configuration', default=default_simulation_name,
-                                                label='Simulation name'), self, name='input-simulation-name-id')
+                                                label='Simulation name'), self, name='input_simulation_name_id')
 
     def fill_from_post(self, form_data):
         super(SimulatorFinalFragment, self).fill_from_post(form_data)
@@ -246,19 +246,19 @@ class SimulatorPSEParamRangeFragment(ABCAdapterForm):
 
     @staticmethod
     def fill_from_post(all_range_parameters, **data):
-        pse_param1_name = data.get('_pse_param1_name')
-        pse_param2_name = data.get('_pse_param2_name')
+        pse_param1_name = data.get('pse_param1_name')
+        pse_param2_name = data.get('pse_param2_name')
 
         pse_param1 = all_range_parameters.get(pse_param1_name)
         if pse_param1.type is float:
-            pse_param1_lo = data.get('_pse_param1_lo')
-            pse_param1_hi = data.get('_pse_param1_hi')
-            pse_param1_step = data.get('_pse_param1_step')
+            pse_param1_lo = data.get('pse_param1_lo')
+            pse_param1_hi = data.get('pse_param1_hi')
+            pse_param1_step = data.get('pse_param1_step')
             param1_range = RangeParameter(pse_param1_name, pse_param1.type,
                                           Range(float(pse_param1_lo), float(pse_param1_hi), float(pse_param1_step)),
                                           is_array=pse_param1.is_array)
         else:
-            param1_range_str = data.get('_pse_param1')
+            param1_range_str = data.get('pse_param1')
             pse_param1_name += '.gid'
             param1_range = RangeParameter(pse_param1_name, pse_param1.type,
                                           SimulatorPSEParamRangeFragment._prepare_pse_uuid_list(param1_range_str))
@@ -267,14 +267,14 @@ class SimulatorPSEParamRangeFragment(ABCAdapterForm):
         if pse_param2_name:
             pse_param2 = all_range_parameters.get(pse_param2_name)
             if pse_param2.type is float:
-                pse_param2_lo = data.get('_pse_param2_lo')
-                pse_param2_hi = data.get('_pse_param2_hi')
-                pse_param2_step = data.get('_pse_param2_step')
+                pse_param2_lo = data.get('pse_param2_lo')
+                pse_param2_hi = data.get('pse_param2_hi')
+                pse_param2_step = data.get('pse_param2_step')
                 param2_range = RangeParameter(pse_param2_name, float,
                                               Range(float(pse_param2_lo), float(pse_param2_hi), float(pse_param2_step)),
                                               is_array=pse_param2.is_array)
             else:
-                param2_range_str = data.get('_pse_param2')
+                param2_range_str = data.get('pse_param2')
                 pse_param2_name += '.gid'
                 param2_range = RangeParameter(pse_param2_name, pse_param1.type,
                                               SimulatorPSEParamRangeFragment._prepare_pse_uuid_list(param2_range_str))
--- a/framework_tvb/tvb/adapters/simulator/simulator_fragments.py
+++ b/framework_tvb/tvb/adapters/simulator/simulator_fragments.py
@@ -164,7 +164,7 @@ class SimulatorFinalFragment(ABCAdapterForm):
     def __init__(self, prefix='', project_id=None, simulation_number=1):
         super(SimulatorFinalFragment, self).__init__(prefix, project_id)
         default_simulation_name = "simulation_" + str(simulation_number)
-        self.length = ScalarField(Simulator.simulation_length, self)
+        self.simulation_length = ScalarField(Simulator.simulation_length, self)
         self.simulation_name = ScalarField(Attr(str, doc='Name for the current simulation configuration', default=default_simulation_name,
                                                 label='Simulation name'), self, name='input_simulation_name_id')
 
--- a/framework_tvb/tvb/adapters/simulator/simulator_fragments.py
+++ b/framework_tvb/tvb/adapters/simulator/simulator_fragments.py
@@ -280,3 +280,10 @@ class SimulatorPSEParamRangeFragment(ABCAdapterForm):
                                               SimulatorPSEParamRangeFragment._prepare_pse_uuid_list(param2_range_str))
 
         return param1_range, param2_range
+
+
+class SimulatorPSEFinalFragment(ABCAdapterForm):
+    def __init__(self, prefix='', project_id=None):
+        super(SimulatorPSEFinalFragment, self).__init__(prefix, project_id)
+        self.simulation_name = ScalarField(Attr(str, doc='Name for the current simulation configuration', label='Number of simulations'),
+                                           self, name='number_of_simulations', disabled=True)
--- a/framework_tvb/tvb/adapters/simulator/simulator_fragments.py
+++ b/framework_tvb/tvb/adapters/simulator/simulator_fragments.py
@@ -281,9 +281,3 @@ class SimulatorPSEParamRangeFragment(ABCAdapterForm):
 
         return param1_range, param2_range
 
-
-class SimulatorPSEFinalFragment(ABCAdapterForm):
-    def __init__(self, prefix='', project_id=None):
-        super(SimulatorPSEFinalFragment, self).__init__(prefix, project_id)
-        self.simulation_name = ScalarField(Attr(str, doc='Name for the current simulation configuration', label='Number of simulations'),
-                                           self, name='number_of_simulations', disabled=True)
--- a/framework_tvb/tvb/core/entities/file/simulator/monitor_h5.py
+++ b/framework_tvb/tvb/core/entities/file/simulator/monitor_h5.py
@@ -132,7 +132,7 @@ class BoldH5(MonitorH5):
     def __init__(self, path):
         super(BoldH5, self).__init__(path)
         self.period = Scalar(Bold.period, self)
-        self.equation = EquationScalar(Bold.equation, self)
+        self.hrf_kernel = EquationScalar(Bold.hrf_kernel, self)
         self.hrf_length = Scalar(Bold.hrf_length, self)
 
 
--- a/framework_tvb/tvb/core/entities/file/simulator/monitor_h5.py
+++ b/framework_tvb/tvb/core/entities/file/simulator/monitor_h5.py
@@ -132,7 +132,7 @@ class BoldH5(MonitorH5):
     def __init__(self, path):
         super(BoldH5, self).__init__(path)
         self.period = Scalar(Bold.period, self)
-        self.hrf_kernel = EquationScalar(Bold.hrf_kernel, self)
+        self.equation = EquationScalar(Bold.equation, self)
         self.hrf_length = Scalar(Bold.hrf_length, self)
 
 
--- a/framework_tvb/tvb/core/services/simulator_service.py
+++ b/framework_tvb/tvb/core/services/simulator_service.py
@@ -37,6 +37,7 @@ import json
 import os
 import shutil
 import uuid
+import numpy
 from tvb.basic.logger.builder import get_logger
 from tvb.core.entities.file.files_helper import FilesHelper
 from tvb.core.entities.file.simulator.simulator_h5 import SimulatorH5
@@ -166,11 +167,11 @@ class SimulatorService(object):
                     simulator = copy.deepcopy(session_stored_simulator)
                     simulator.gid = uuid.uuid4()
                     self._set_simulator_range_parameter(simulator, range_param1.name, param1_value)
-                    ranges = {range_param1.name: param1_value[0]}
+                    ranges = {range_param1.name: param1_value[0] if type(param1_value) is numpy.ndarray else param1_value}
 
                     if param2_value:
                         self._set_simulator_range_parameter(simulator, range_param2.name, param2_value)
-                        ranges[range_param2.name] = param2_value[0]
+                        ranges[range_param2.name] = param2_value[0] if type(param2_value) is numpy.ndarray else param2_value
 
                     ranges = json.dumps(ranges)
 
--- a/framework_tvb/tvb/core/services/simulator_service.py
+++ b/framework_tvb/tvb/core/services/simulator_service.py
@@ -155,19 +155,24 @@ class SimulatorService(object):
             operation_group = burst_config.operation_group
             metric_operation_group = burst_config.metric_operation_group
             operations = []
-            range_param2_values = []
+            range_param2_values = [None]
             if range_param2:
                 range_param2_values = range_param2.get_range_values()
             first_simulator = None
-            for param1_value in range_param1.get_range_values():
-                for param2_value in range_param2_values:
+
+            for param2_value in range_param2_values:
+                for param1_value in range_param1.get_range_values():
                     # Copy, but generate a new GUID for every Simulator in PSE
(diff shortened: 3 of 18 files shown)


framework_tvb/tvb/interfaces/web/controllers/common.py
```diff
@@ -89,6 +89,8 @@
 KEY_SECTION_TITLES = "section_titles"
 KEY_SUBSECTION_TITLES = "sub_section_titles"
 KEY_SIMULATION_NAME = "input_simulation_name_id"
+KEY_PSE_PARAM_1 = "pse_param1"
```
liadomide: We have used before `tvb.core.entities.model.model_burst.RANGE_PARAMETER_1` and `RANGE_PARAMETER_2` Aren't these the same thing?

liadomide:
I see conflicts on this PR. A rebase with upstream master would be in order.

VinczeRobert:
> I see conflicts on this PR. A rebase with upstream master would be in order.

I know, I will do that.

VinczeRobert:
I think this PR should be reviewed now because there are many changes and they cover more than what this task was supposed to be:
- I made PSE Simulation run with one and two range params (excluding the case where the range param have GID, for example a connectivity)
- Discrete PSE Visualizer works now
- Loading works correctly now for PSE
- I made the simulation_name and simulation_length be stored and loaded correctly for both the normal simulation and the one with PSE
- The ranges of the params are in a different method now than launch_pse and launch_pse uses a fragment where there is a message about the number of the simulations that are about to be launched
- I made launch_simulation and launch_pse not being called when we just want to load the data
- I added three private methods in simulator_controller to avoid some duplicated code

scientific_library/tvb/simulator/monitors.py
```diff
@@ -796,7 +796,7 @@ class Bold(Monitor):
         an integral multiple of 500. Typical measurment interval (repetition
         time TR) is between 1-3 s. If TR is 2s, then Bold period is 2000ms.""")
 
-    hrf_kernel = Attr(
+    equation = Attr(
```
liadomide: Why this rename? 
The API in the Simulator should stay as stable as possible. This was named `hrf_kernel ` even from before neotraits

tvb_documentation/sim_doc/img/classes_monitors.svg
```diff
@@ -17,7 +17,7 @@
 <text text-anchor="start" x="1344.5" y="-371.3" font-family="Times,serif" font-size="14.00">hemodynamic_response_function</text>
 <text text-anchor="start" x="1344.5" y="-356.3" font-family="Times,serif" font-size="14.00">hrf_length : Float</text>
 <text text-anchor="start" x="1344.5" y="-341.3" font-family="Times,serif" font-size="14.00">period : Float</text>
-<text text-anchor="start" x="1344.5" y="-326.3" font-family="Times,serif" font-size="14.00">hrf_kernel : HRFKernelEquation</text>
+<text text-anchor="start" x="1344.5" y="-326.3" font-family="Times,serif" font-size="14.00">equation : HRFKernelEquation</text>
```
liadomide: This is not a file to change manually. Did you run `class_diagrams.sh` ?

framework_tvb/tvb/interfaces/web/controllers/common.py
liadomide: I am still not convinced why we need these?
Can't we use Forms ?

framework_tvb/tvb/core/services/simulator_service.py
```diff
@@ -155,19 +156,24 @@ def async_launch_and_prepare_pse(self, burst_config, user, project, simulator_al
...
-            range_param2_values = []
+            range_param2_values = [None]
             if range_param2:
                 range_param2_values = range_param2.get_range_values()
             first_simulator = None
-            for param1_value in range_param1.get_range_values():
-                for param2_value in range_param2_values:
+
+            for param2_value in range_param2_values:
+                for param1_value in range_param1.get_range_values():
```
liadomide: Why was this switch of the loops necessary?
Theoretically it gives the same set of iterations, just in a different order, and it is not very easy to digest when reading the code.

framework_tvb/tvb/core/entities/file/simulator/monitor_h5.py
```diff
@@ -132,7 +132,7 @@ class BoldH5(MonitorH5):
     def __init__(self, path):
         super(BoldH5, self).__init__(path)
         self.period = Scalar(Bold.period, self)
-        self.hrf_kernel = EquationScalar(Bold.hrf_kernel, self)
+        self.equation = EquationScalar(Bold.equation, self)
```
liadomide: See the comments bellow. Why is this rename necessary?

framework_tvb/tvb/adapters/simulator/simulator_fragments.py
```diff
@@ -273,16 +267,23 @@ def fill_from_post(all_range_parameters, **data):
...
-                param2_range_str = data.get('_pse_param2')
+                param2_range_str = data.get('pse_param2')
                 pse_param2_name += '.gid'
                 param2_range = RangeParameter(pse_param2_name, pse_param1.type,
                                               SimulatorPSEParamRangeFragment._prepare_pse_uuid_list(param2_range_str))
 
         return param1_range, param2_range
+
+
+class SimulatorPSEFinalFragment(ABCAdapterForm):
```
liadomide: Can you pls explain why for PSE we need a different final form for the name ?
Can't we use the one with the length as on regular single simulations ?

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -748,118 +789,171 @@ def set_monitor_equation(self, **data):
...
+                                                          last_form_url=self.last_loaded_form_url,
+                                                          is_launch_pse_fragment=True)
+        return rendering_rules.to_dict()
+
+    @cherrypy.expose
+    @handle_error(redirect=False)
+    @check_user
+    def launch_pse(self, **data):
         session_stored_simulator = common.get_from_session(common.KEY_SIMULATOR_CONFIG)
+        session_stored_simulator.simulation_length = float(common.get_from_session(common.KEY_SIMULATION_LENGTH))
```
liadomide: why keep length as a separate field in session? 
It should already be set on session_stored_simulator. Or why not?

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -748,118 +789,171 @@ def set_monitor_equation(self, **data):
...
 
         project = common.get_current_project()
         user = common.get_logged_user()
 
         burst_config = common.get_from_session(common.KEY_BURST_CONFIG)
         burst_config.start_time = datetime.now()
-        # if burst_name != 'none_undefined':
-        #     burst_config.name = burst_name
+        simulation_name = common.get_from_session(common.KEY_SIMULATION_NAME)
```
liadomide: Why this is kept separately in session, why not set it already on the object `burst_config ` in session?

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -893,6 +990,7 @@ def launch_simulation(self, launch_mode, **data):
 
         burst_name = current_form.simulation_name.value
         session_stored_simulator = common.get_from_session(common.KEY_SIMULATOR_CONFIG)
+        session_stored_simulator.simulation_length = float(data['simulation_length'])
```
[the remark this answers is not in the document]
VinczeRobert: Yes, it can, I fixed this. The problem was the in the SimulatorFinalFragment, the name of the field was length, not simulation_length, so it wasn't added by the fill_from_post. Now it is okay, I renamed it to simulation_length.

tvb_documentation/sim_doc/img/classes_monitors.svg
VinczeRobert: No, I just renamed it everywhere.

scientific_library/tvb/simulator/monitors.py
VinczeRobert: I wrote the reason for this in one of my commits: " Loading with Bold monitor wasn't working because in some places the field for equation was named 'equation' and in other places 'hrf_kernel' (so basically the monitor had two attributes with the exact same information), so that's why I renamed all of them to 'equation' to make it work and because it is clearer. " However, if it is important for the name to to stay the same in those places, I will make the necessary changes while keeping the name.

framework_tvb/tvb/core/services/simulator_service.py
VinczeRobert: Yeah, I realized it is not necessary... The important change there was that in case of a single parameter to make range_param2_values be a list = [None] so it can still enter in the fors and compute the array of operation. Original order can stay.

framework_tvb/tvb/core/entities/file/simulator/monitor_h5.py
VinczeRobert: Answered this on an above comment.

framework_tvb/tvb/core/services/simulator_service.py
VinczeRobert: And about the PSE_RANGE_PARAM_1 and 2 (I dunno why it doesn't let me directly reply to that comment), the thing is that those range params are needed in launch_pse, but they don't get it in the data, because that final fragment doesn't have them and range params are also not part of session_stored_simulator. So how can I do it differently without having to store them in the session?

framework_tvb/tvb/adapters/simulator/simulator_fragments.py
VinczeRobert: That field should be simulation_number, not name... I dunno why I wrote simulation_name. I added that final form because there was a TO-DO which required to separate the setting of range params from the pse_launching and the only thing the last fragment should do is dispay the number of simulations that are about to be launched.

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
VinczeRobert: simulation_length is actually set in launch_simulation method. However, if you wish to run a PSE simulation, you don't go through launch_simulation. I actually had an idea that simulation name and length should be stored in a separate fragment from the last one with launch simulation and then in launch_simulation we just launch it, we don't send any parameter. That change would solve this problem too. Should I do this?

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
VinczeRobert: Same reason as for the question about the simulation_length setting on line 935.

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
liadomide: Can't we stay with the same form (containing name and length) regardless if it is PSE or single op?

tvb_documentation/demos/exploring_the_bold_monitor.ipynb
```diff
@@ -241,7 +241,7 @@
     }
    ],
    "source": [
-    "bold.hrf_kernel"
+    "bold.equation"
```
liadomide: this file should be reverted also

framework_tvb/tvb/core/services/simulator_service.py
liadomide: I would expect that we use now forms everywhere.
Indeed, these can not stay on a seesion_stored_sim, but there should be a form displaying the ranges. Isn't it?

framework_tvb/tvb/adapters/simulator/simulator_fragments.py
VinczeRobert: Actually Paula suggested me to use a form for that and I think the idea was that the user could see the number of simulations to be launched before actually launching them (like a last chance to change his mind if he thinks that there are too many or not enough). But if that's not important, I think I could display it from JS and in that case the launch_pse method will just get a None parameter for the fragment.

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
VinczeRobert: In the way that it is now, launch_simulation gets name and simulation_length as parameters, launch_pse doesn't. If I make them both get these parameters (i.e. using the same form), then in case of launch_pse the form will be included twice in the flow, which we don't want. Something that I can think of is to make the user choose between normal and pse simulation BEFORE that form appears.

tvb_documentation/demos/exploring_the_bold_monitor.ipynb
VinczeRobert: Okay.

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -200,6 +202,12 @@ def include_setup_pse(self):
             return True
         return False
 
+    @property
+    def include_launch_pse_button(self):
+        if self.is_launch_pse_fragment and (not self.load_readonly or self.is_simulation_copy):
```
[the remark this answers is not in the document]
VinczeRobert: I haven't checked that yet. I will do it very soon.

framework_tvb/tvb/core/services/simulator_service.py
VinczeRobert: There is. SimulatorPSEConfigurationFragment is for choosing the params, and SimulatorPSEParamRangeFragment for choosing the ranges. But the params are needed not just in their respective setting method, but in launch_pse as well.

framework_tvb/tvb/interfaces/web/controllers/common.py
liadomide: Understood today that KEY_PSE_PARAM_1 and KEY_PSE_PARAM_2 should stay in session as separate objects.

framework_tvb/tvb/adapters/simulator/simulator_fragments.py
liadomide: As discussed today, this SimulatorPSEFinalFragment could disappear, and we shall use the JS code for displaying and computing the message with PSE no of ops.

framework_tvb/tvb/interfaces/web/controllers/burst/exploration_controller.py
```diff
@@ -135,11 +136,11 @@ def get_series_array_discrete(self, datatype_group_gid, backPage, color_metric=N
             color_metric = None
         if size_metric == 'None':
             size_metric = None
-
+        datatype_group_gid = uuid.UUID(datatype_group_gid)
```
liadomide: If you would put this conversion after the `if self._is_compatible(` the .hex won't be necessary

framework_tvb/tvb/interfaces/web/static/js/bursts.js
```diff
@@ -936,6 +936,54 @@ function displayBurstTree(selectedHref, selectedProjectID, baseURL) {
     $("#div-burst-tree").show();
 }
 
+function displayPseSimulationMessage(simulation_mesage){
```
liadomide: Have you searched for the previous JS code doing this count?
It should be either reused, or removed.

framework_tvb/tvb/interfaces/web/static/js/bursts.js
```diff
@@ -936,6 +936,54 @@ function displayBurstTree(selectedHref, selectedProjectID, baseURL) {
...
+    param1_difference = pse_param1_hi - pse_param1_lo;
+    pse_param1_number = Math.floor( param1_difference / pse_param1_step);
+    remainder_param1 = param1_difference % pse_param1_step;
+    if(remainder_param1 !== 0){
+        pse_param1_number = pse_param1_number + 1;
+    }
+
+    pse_param2_lo = $("#pse_param2_lo");
+
+    if(pse_param2_lo.length !== 0) {
```
liadomide: This looks like duplicate (what is done for both param1 and param2. 
Can't you extract a reusable function?

tvb_documentation/demos/exploring_the_bold_monitor.ipynb
```diff
@@ -1945,7 +1945,7 @@
     "\n",
     "figure()\n",
     "for hrf in hrf_kernels: \n",
-    "    bold_monitor = monitors.Bold(hrf_kernel=hrf)\n",
+    "    bold_monitor = monitors.Bold(equation=hrf)\n",
```
liadomide: This should be reverted, right?

VinczeRobert: Yes, I reverted it.

framework_tvb/tvb/interfaces/web/controllers/burst/exploration_controller.py
VinczeRobert: Okay

framework_tvb/tvb/interfaces/web/static/js/bursts.js
VinczeRobert: I did now... I searched for 'pse' in js. files, but I couldn't find anything that looks like some kind of count of pse. Maybe it has been already deleted.

liadomide: Have you looked into range.js (or ranges.js)?

framework_tvb/tvb/interfaces/web/controllers/burst/exploration_controller.py
liadomide: Can you still change as I suggest and test to report back what error you get?

VinczeRobert: Okay, I will make it like you said. The error is: 
2020-03-30 18:57:56,027 - WARNING - tvb.interface.web.controllers.decorators - Redirect converted to error: (['http://127.0.0.1:8080/burst/explore/pse_error?adapter_name=Discrete%20Parameter%20Space%20Exploration&message=Discrete%20PSE%20is%20incompatible%20%28most%20probably%20due%20to%20result%20size%20being%20too%20large%29.'], 303).

And then if I click on that link it takes me to a web page, with the message:
Incompatible visualizer: Discrete Parameter Space Exploration
Reason: Discrete PSE is incompatible (most probably due to result size being too large)

liadomide: It looks like either a deeper problem, or there must be another stack trace in another place. 
I see nothing about the exploration_controller in tis message.
Have you tried a debug ?

Do you understand why it looks "fishy" when reading the code ?

framework_tvb/tvb/interfaces/web/static/js/bursts.js
liadomide: See range.js after line 331.
Also, check all functions in that JS file for their usages, please.

framework_tvb/tvb/interfaces/web/controllers/burst/exploration_controller.py
liadomide: Also, I can not see any recent change by Paula in the history of that file neither on master nor in this PS list of commits. Can you pls point me to a commit Diff ?

liadomide: Also, it looks strange the way the 2 methods `get_series_array_discrete` and `draw_discrete_exploration` are written in this controller now. I think doth should have no UUId, nor .hex, and we should also change line 162 from pse_discreet.py into `datatype_group = dao.get_datatype_group_by_gid(datatype_group_gid)`. Bet let us have a call tomorrow with screen share and discuss these better live

VinczeRobert: Oh sorry, I have confused that with another commit. Okay, let's discuss it live.

framework_tvb/tvb/interfaces/web/static/js/bursts.js
VinczeRobert: Yeah, I thought that those methods could be used in our case, but there is a problem. They use an id named "range_1" and "range_2", but the inputs for the ranges in the controller have "pse_param1_hi" and so on. So I strongly believe that those methods are used somewhere else, if used at all. I will check if they are used anywhere.

VinczeRobert: A list of usages of all the functions in range.js:
- toggleRange - only used in number2component.html
- initRangeComponent - only used in the previous toggleRange
- _validateInput - used in the _updateRangeInterval function of this same file
- _updateRangeInterval - used only in initRangeComponent of this same file
- updateRangeValues - used only in bursts.js, function tryExpandRangers, line 374
- _getDataSelectRangeComponent - this seems not to be used anywhere, should I delete it?
- _getValues - used only in _displayRangeLabels of this same file
- _displayRangeLabels - used in multiple places in this same file: in updateRangeValues, initRangeComponent, slide, updateRangeInterval
- _prepareDataForValues - used in the same places as above 
- _getUnnormalizedValue - used two times in _displayRangeLabels, in _prepareDataForSubmit and _refreshFromToValues in this same file
- getNormalizedValue - used more times in updateRangeValues, in initRangeComponent and in updateRangeInterval in this same file
- _refreshFromToValues - used in initRangeComponent and slide in this same file
- displayRangeComponent - used in toggleRange of this same file and also used in stepsUpdate.js, line 29
- _getOpsForRanger - only used in RANGE_computeNrOfOps in this same file
- RANGE_computeNrOfOps - used in _prepareDataForSubmit and in disableRangeComponent of this same file, also used in stepsUpdate.js, line 68
-

liadomide: Ok, then this cleanup will be done in a separate task

PR #66, opened 2020-03-10
TVB-2628 Check if user can access a resource (project, operation, datatype etc.)

bvalean:


framework_tvb/tvb/interfaces/rest/server/access_permissions/permissions.py
```diff
@@ -0,0 +1,84 @@
...
+
+    @abstractmethod
+    def get_resource_owners(self):
+        """
+        :return: a list of users id who can access the requested resource
+        """
+        raise RuntimeError("Not implemented.")
+
+
+class OperationAccessPermission(ResourceAccessPermission):
```
liadomide: This is a bit too restrictive.
A user able to access a Project, should be allowed to see all Operation under that project, as well as all DataTypes in the project.

framework_tvb/tvb/interfaces/rest/server/access_permissions/permissions.py
```diff
@@ -0,0 +1,84 @@
...
+        project_members = self.project_dao.get_members_of_project(project.id)
+        return [project_member.id for project_member in project_members]
+
+
+class DataTypeAccessPermission(ResourceAccessPermission):
+    def __init__(self, datatype_gid):
+        super(DataTypeAccessPermission, self).__init__(datatype_gid)
+        self.datatype_dao = DatatypeDAO()
+
+    def get_resource_owners(self):
```
liadomide: This seems again much more restrictive than envisioned under tvb-web GUI.
If a user has access to a Project, the he can see all DataTypes under that Project.
Plus, we have the DataType links ;-) which are a way to expose a DataType from a Project1 into another Project2, without exposing the entire Project1.

framework_tvb/tvb/interfaces/rest/server/access_permissions/permissions.py
```diff
@@ -0,0 +1,84 @@
...
+        super(ProjectAccessPermission, self).__init__(project_gid)
+        self.project_dao = CaseDAO()
+
+    def get_resource_owners(self):
+        try:
+            project = self.project_dao.get_project_lazy_by_gid(self.resource_identifier)
+        except ProjectServiceException:
+            return []
+        project_members = self.project_dao.get_members_of_project(project.id)
+        return [project_member.id for project_member in project_members]
```
liadomide: This is ok, as a check.
But can we add Project members from tvb-rest api ?

bvalean: No, we cannot add project members from the REST api. Only when we create a new db user from a keycloak user, we add this user as a member of the default project.

liadomide: it should be possible to add:
- get_tvb_users
- add_project_members(project_gid, user_names_list)

PR #68, opened 2020-03-12
DSL related licences

liadomide:
I see Mako and NeuroML as new dependencies brought by the DSL. Something else?

These should be listed under accepted TVB dependencies.
For mako it was easy to identify the license. 
Could you point to what from NeuroML we should cite ?

### DIFF
--- a/tvb_build/tvb_build/third_party_licenses/license/LICENSE_lems.txt
+++ b/tvb_build/tvb_build/third_party_licenses/license/LICENSE_lems.txt
@@ -1,0 +1,166 @@
+                   GNU LESSER GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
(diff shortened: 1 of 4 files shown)


liadomide:
@DeLaVlag Could you pls help us with the dependencies licenses also?

maedoc:
LEMS is the relevant repo (https://github.com/LEMS/pylems) and is [LGPL licensed](https://github.com/LEMS/pylems/blob/master/LICENSE.lesser)

maedoc:
thx @liadomide for this; I had verified the licesnse prior to merge but didn't think to put them in there.

liadomide:
I see now the relevant comments about the licenses on PR #58 . 
I missed that conversation before.

DeLaVlag:
Thanks for the merge. The title of the readme is not correct. Title should be 'TVB model generation using LEMS format'. Could this be altered?

PR #69, opened 2020-03-12
TVB-2636 Enhance User entity

bvalean:
For security reasons, we do not want to expose the username


framework_tvb/tvb/interfaces/rest/commons/dtos.py
```diff
@@ -56,10 +56,8 @@ def __init__(self, user=None, **kwargs):
         """
         self.update(kwargs)
         if user is not None:
-            self.username = user.username
+            self.gid = user.gid
```
liadomide: We should also have display_name here, right?

liadomide:
This branch is a bit outdated now compared with master. This is the reason why tvb-library unit tests will not be correctly discovered here, and few other tests are failing.

framework_tvb/tvb/interfaces/rest/commons/dtos.py
bvalean: Yes. We will have also the display_name field in the DTO

PR #70, opened 2020-03-12
WIP - Enable run of TVB Distribution with HPC backends

liadomide:
- We envision a TVB web deploy, plus HPC access for running large async operations 
- compared with the cluster deployment, we can not share the same Postgres DB and storage folder between nodes
- thus steps for Stage-in and Stage-Out are needed
- we plan for an HBP deploy in this formula soon, but this solution should be generic enough for any HPC backend.


liadomide:
I close this, as we have #140 after rebase and solve of conflicts

PR #75, opened 2020-03-16
HBP-64

popaula937:
Added the Dockerfile, Jenkinsfile and .tvb.configuration necessary for HBP openshift deploy, in a separate folder. Working on adding a configuration file for the openshift deploy, so we don't have to do it manually.
Should we keep this Dockerfile and Jenkinsfile near the others or is it more clear to have them together under the hbp-openshift folder?


popaula937:
@liadomide we use the same Dockerfile, Jenkinsfile and .tvb.configuration files for both deployments, so we thought about a different folder structure. What do you think?

tvb_build/deploy_with_keycloak/Jenkinsfile
```diff
@@ -0,0 +1,34 @@
+pipeline {
+    agent any
+
+    environment {
+        FULL_DOCKER_IMAGE_NAME = 'thevirtualbrain/tvb-run'
+        DEPLOY_TAG='deploy_with_keycloak'
```
liadomide: maybe use a shorter tag, just 'keycloak'

PR #78, opened 2020-03-23
HBP-69 Request profile and email scopes from keycloak server

bvalean:


framework_tvb/tvb/interfaces/web/templates/jinja2/user/login.html
```diff
@@ -31,7 +31,9 @@ <h3>Login User</h3>
 										.appendTo("#user_login_form");
 								$('#user_login_form').submit();
 							} else {
-								keycloak.login()
+								keycloak.login({
+									scope: 'profile email'
```
liadomide: will this work for all the 3 instances of keycloak we are currently testing with ?

bvalean: yes

PR #82, opened 2020-03-30
TVB-2641 Use Fraunhofen REST url by default

bvalean:
Use Fraunhofen REST url by default in examples. 
There is also an option to pass the REST url to the examples:

python -m tvb.interfaces.rest.client.examples.fire_simulation --rest-url=http://localhost:8080


framework_tvb/tvb/interfaces/rest/client/examples/__init__.py
```diff
@@ -0,0 +1,11 @@
+import sys
```
liadomide: Missing license, and the use of __init__.py is not ideal, but ok as an idea to accept input params

PR #83, opened 2020-03-30
Prepare release

popaula937:
Prepare release for version 2.0.3a1 of tvb-rest-client and 2.0.3 of tvb-library

### DIFF
--- a/dev_resources/docs/ReleaseProcedure
+++ b/dev_resources/docs/ReleaseProcedure
@@ -58,4 +58,22 @@ When creating a new TVB set of release packages you should:
     
     
      
- 
+ TVB REST CLIENT release procedure:
+    Usually, we need to:
+        1. Release the tvb-rest-client package on Pypi:
+            - it is mandatory to take the package from a Jenkins build
+            - make sure you keep 'forever' that build and give it a significant name (eg: Release ​tvb-rest-client ​2.0.2)
+            - download on your local machine the package from Jenkins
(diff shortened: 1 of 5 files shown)


dev_resources/docs/ReleaseProcedure
```diff
@@ -58,4 +58,22 @@ When creating a new TVB set of release packages you should:
...
+            - make sure you keep 'forever' that build and give it a significant name (eg: Release ​tvb-rest-client ​2.0.2)
+            - download on your local machine the package from Jenkins
+            - install it in a new environment and run the examples (eg: fire_simulation.py) to ensure it works correctly
+            - upload it on Pypi
+
+        2. Update the REST server docker image:
+            Currently, we have 2 images:
+                - thevirtualbrain/tvb-run:keycloak - it is built and uploaded by Jenkins. We have to mount the proper
+                                                     Keycloak configuration files inside the container.
+                - thevirtualbrain/tvb-run:tvb-rest - it comes with a Keycloak client configuration. For deploy, we have
```
liadomide: This image we can already declare deprecated, right?
We might still offer it in 1-2 sprints for IBEC, but will become obsolete, as we do not want to be obliged to keep the Codemart Keycloack forever.

popaula937: Yes, I just want to keep it now for easy sharing with IBEC

PR #84, opened 2020-03-31
HBP-72 Add static help pages

bvalean:


tvb_build/deploy_with_keycloak/Jenkinsfile
```diff
@@ -23,7 +41,7 @@ pipeline {
     }
     post {
         changed {
-            mail to: 'paula.popa@codemart.ro',
+            mail to: 'paula.popa@codemart.ro bogdan.valean@codemart.ro',
```
liadomide: me too, pls :-)

framework_tvb/tvb/interfaces/web/templates/jinja2/user/profile.html
```diff
@@ -32,7 +32,7 @@ <h3>User Profile</h3>
 
     <section class="profile">
         <h2><mark>User</mark> details</h2>
-		{% if showOnlineHelp %}
+		{% if showOnlineHelp and keycloakConfigWebFile is not defined %}
```
liadomide: I don't understand what is the change doing here. Pls explain

tvb_build/deploy_with_keycloak/Jenkinsfile
```diff
@@ -7,8 +7,26 @@ pipeline {
     }
 
     stages {
+        stage('Copy Step 1') {
```
liadomide: This is very good, and will update/maintain on Docker hub thevirtualbrain/tvb-run:keycloak
We should have also thevirtualbrain/tvb-run:latest updated. 
Could it be ok in the same Jenkins job ? 
This part is not part of the current task, rather of TVB-2523, but might be easy to fix here or soon after these changes.

framework_tvb/tvb/interfaces/web/templates/jinja2/user/profile.html
bvalean: This change is hiding the online help question mark near the User Details in the user profile page when keycloak is enabled.

tvb_build/deploy_with_keycloak/Jenkinsfile
bvalean: I would remove the keycloak tag an keep only the latest one because the differences will be that in the keycloak tag keycloak is enabled by default, but in our deploys (VBC and HBP) we override the tvb config file anyway and we can set this flag to True here.

framework_tvb/tvb/interfaces/web/templates/jinja2/user/profile.html
liadomide: ... because the help on this section mentions TVB users and that is no longer relevant in this deployment. Long term we should update the help to contain info about the keycloak login. But that is not something which we should fix now.

tvb_build/deploy_with_keycloak/Jenkinsfile
liadomide: good.

PR #85, opened 2020-04-01
TVB-2463: Simulation Launch to work with GUID ranges

VinczeRobert:
I managed to succesfully launch a simulation with connectivity as range param and I was able to see if with the Discrete Visualizer. I also adapted the js code which calculates the operations.

A question: Should we add a filter which only lets you run with connectivities that have the same number of regions as the original connectivity chosen for simulation?

### DIFF
--- a/framework_tvb/tvb/adapters/simulator/simulator_fragments.py
+++ b/framework_tvb/tvb/adapters/simulator/simulator_fragments.py
@@ -218,7 +218,7 @@ class SimulatorPSEParamRangeFragment(ABCAdapterForm):
 
         else:
             self.pse_param1_dt = DataTypeSelectField(h5.REGISTRY.get_index_for_datatype(pse_param1.type), self,
-                                                     name='pse_param1', required=True, label='pse_param1',
+                                                     name='pse_param1_guid', required=True, label='pse_param1_guid',
                                                      dynamic_conditions=pse_param1.range_definition,
                                                      has_all_option=True)
 
@@ -234,7 +234,7 @@ class SimulatorPSEParamRangeFragment(ABCAdapterForm):
                                                         default=pse_param2.range_definition.step)
             else:
                 self.pse_param2_dt = DataTypeSelectField(h5.REGISTRY.get_index_for_datatype(pse_param2.type), self,
-                                                         name='pse_param2', required=True, label='pse_param2',
+                                                         name='pse_param2_guid', required=True, label='pse_param2_guid',
                                                          dynamic_conditions=pse_param2.range_definition,
                                                          has_all_option=True)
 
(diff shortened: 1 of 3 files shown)


framework_tvb/tvb/adapters/simulator/simulator_fragments.py
```diff
@@ -218,7 +218,7 @@ def __init__(self, pse_param1, pse_param2, prefix='', project_id=None):
 
         else:
             self.pse_param1_dt = DataTypeSelectField(h5.REGISTRY.get_index_for_datatype(pse_param1.type), self,
-                                                     name='pse_param1', required=True, label='pse_param1',
+                                                     name='pse_param1_guid', required=True, label='pse_param1_guid',
```
liadomide, line 221: Is this the label text as it appears in GUI?
Then it should be composed of separate words, not _

framework_tvb/tvb/core/services/simulator_service.py
```diff
@@ -167,11 +167,22 @@ def async_launch_and_prepare_pse(self, burst_config, user, project, simulator_al
                     simulator = copy.deepcopy(session_stored_simulator)
                     simulator.gid = uuid.uuid4()
                     self._set_simulator_range_parameter(simulator, range_param1.name, param1_value)
-                    ranges = {range_param1.name: param1_value[0] if type(param1_value) is numpy.ndarray else param1_value}
+                    if type(param1_value) is numpy.ndarray:
```
liadomide, line 170: As this check is repeated twice, it calls for a private method use

framework_tvb/tvb/interfaces/web/static/js/bursts.js
```diff
@@ -950,18 +950,38 @@ function displayPseSimulationMessage() {
     const THREASHOLD_WARNING = 500;
     const THREASHOLD_ERROR = 50000;
 
-    const pse_param1_lo = $("#pse_param1_lo")[0].valueAsNumber;
-    const pse_param1_hi = $("#pse_param1_hi")[0].valueAsNumber;
-    const pse_param1_step = $("#pse_param1_step")[0].valueAsNumber;
-    const pse_param1_number = calculateValuesInRage(pse_param1_lo, pse_param1_hi, pse_param1_step);
+    // Check if we have first range parameter with guid
```
liadomide, line 953: Again, this might go in a reusable function, once the IDs of the HTML elements is given as params.

framework_tvb/tvb/interfaces/web/static/js/bursts.js
```diff
@@ -946,7 +946,17 @@ function calculateValuesInRage(pse_param_lo, pse_param_hi, pse_param_step){
     return pse_param_number;
 }
 
-function displayPseSimulationMessage() {
+function _getRangeValueForGuidParameter(pse_param_guid){
```
liadomide, line 949: I was thinking to get more in this function.
Including the parts where values are read from HTML $("#pse_param1_hi")[0].valueAsNumber;...

VinczeRobert: So should I read those params dynamically as well? Like sending "pse_param1" as parameter and then compute the rest in a separate function from there?

liadomide: yes, something like `range_no = compute_range_no_for_param_prefix('pse_param1')`

VinczeRobert:
Oh sorry, yes it

> Ok from my side. Should we wait for some other change, or it can be merged?
> I am asking as I still see WIP in the title.

Oh sorry, yes it can be merged.

liadomide:
> A question: Should we add a filter which only lets you run with connectivities that have the same number of regions as the original connectivity chosen for simulation?

Yes, but there is a bigger task to be solved regarding UI Filters

PR #88, opened 2020-04-02
TVB 2550: Added Filter to only show Cortical Surfaces

VinczeRobert:


liadomide:
This would be ok to merge, of only one commit is separated.
Currently I see also the previous PSE over GUIDS state, which is not really the latest in master.
Could you rebase?

PR #89, opened 2020-04-03
TVB-2650 Review baseUrl

bvalean:
As a first change I take the base URL from the request and set it to the BASE URL on the TVBProfile if it is not set.


framework_tvb/tvb/interfaces/web/controllers/users_controller.py
```diff
@@ -433,6 +434,11 @@ def _populate_version(self, template_dictionary):
         template_dictionary[KEY_CURRENT_VERSION_FULL] = TvbProfile.current.version.CURRENT_VERSION
         return template_dictionary
 
+    @staticmethod
+    def _set_base_url():
+        if not TvbProfile.current.web.BASE_URL:
+            TvbProfile.current.web.BASE_URL = cherrypy.request.base
```
liadomide, line 440: We should test this behind a Proxy.
Maybe OPenShift dev would be a good testing bed ?

bvalean:
After these changes the baseUrl still exists in model_parameters, stimulus_surface and stimulus_region but here it is used with another scope.

framework_tvb/tvb/interfaces/web/controllers/users_controller.py
liadomide: we leave this here, as it is used now only when sending emails

PR #91, opened 2020-04-03
WIP: TVB-2555 Making sure that every parameter in the Simulator Loads correctly

VinczeRobert:


scientific_library/tvb/datatypes/cortex.py
```diff
@@ -52,7 +52,7 @@ class Cortex(HasTraits):
 
     region_mapping_data = Attr(
         field_type=region_mapping.RegionMapping,
-        label="region mapping",
+        label="Region mapping",
```
liadomide, line 55: No big deal to change this here, but I think it will be safer if we capitalize all labels when rendering the Jinja template.

VinczeRobert: Okay, I will keep that in mind. I capitalized it because every other parameter from the Simulation Controller is written with capital letters.

VinczeRobert:
Loading the page at refreshing or when you go out of it and back still has two issues:

1. I will explain this one with an example as it is easier for me: let's say you kept going with 'Next' until the setting of the monitor (at this point your last_loaded_form_url will be 'set_monitors'), but then you go back with 'Previous' cause you want to edit some parameters. After this, if you refresh the page or go out of it and then back, you will be pointed at the monitors fragment. This is because last_loaded_form_url is never updated if you go back. For this one, I think we need an Ajax Call from JS to a Python method in the previousWizzardStep function. Or is there any way I can do that directly from JS?

2. After starting a new simulation and refreshing the page (or going out and back in), you can't go the setup PSE part, because it will automatically try to launch the simulation, no matter which of the two buttons you press. I guess here is only a matter of updating the url's in some more cases that I haven't done yet.

Besides this, I still have to work on the copy_configuration part. If both these cases + the copy case is handled, I think this task would be done, but correct me if I'm wrong.

liadomide:
1. If we want the `last_loaded_form_url` to change, yes, we will need to communicate to the server  (cherrypy backend) that the back button was pressed (ajax call from JS to Python). But do we really care for this? as long as there is no error, I think it is also ok to keep the current behavior. At most, the user will need to press `Back` again a few times. If the user would have pressed `back` and changed something, then `Next`, I suspect the correct step is shown, right?
2. I am not sure I understand this point. Maybe in a call tomorrow.

VinczeRobert:
About that method in the BurstDAO:
I believe that line with "not" is not needed. It gave an error, and it seems like without it, we have the wished behaviour. I am not sure, but to me it seems like it doesn't actually count the the exact names, but the names that contain a certain name given as argument. Anyway, normal simulation in branch works. I still have to make refreshing and branching work with pse params as well.

liadomide:
> About that method in the BurstDAO:
> I believe that line with "not" is not needed. It gave an error, and it seems like without it, we have the wished behaviour. I am not sure, but to me it seems like it doesn't actually count the the exact names, but the names that contain a certain name given as argument. Anyway, normal simulation in branch works. I still have to make refreshing and branching work with pse params as well.

Have you tried to use `.notlike` ?

framework_tvb/tvb/core/entities/storage/burst_dao.py
```diff
@@ -81,8 +81,7 @@ def count_bursts_with_name(self, burst_name, project_id):
             count = self.session.query(BurstConfiguration
                                        ).filter_by(project_id=project_id
                                        ).filter(BurstConfiguration.name.like(burst_name + '%')
-                                                ).filter(not_(BurstConfiguration.name.like(burst_name + '/_%/_%', escape='/'))
-                                       ).count()
+                                                ).count()
```
liadomide, line 84: We could use probably `BurstConfiguration.name.notlike`, but you also might be right to remove it.
Have you tried to do a branch from a branch?
A branch from a copy ?
Branch of a branch or a copy?

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -689,6 +693,8 @@ def set_monitors(self, **data):
 
             if cherrypy.request.method != 'POST':
```
liadomide, line 694: Don't we have here 6 lines repeated at least 3 times ?

framework_tvb/tvb/core/entities/storage/burst_dao.py
VinczeRobert: I will try that today.

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
VinczeRobert: Which lines are you thinking about?

In any case, at the end when I have no errors at loading, I will review the code and extract duplicated code in functions.

VinczeRobert:
I want to talk a little bit about the "refreshing the page problem". It works correctly if you are not trying to launch simulation with PSE, but if you want there are some problems the way launch_simulation works now.

The problem starts with the fact that when you load a simulation (refreshing is a kind of loading too) and you are at the fragment before the last one with the simulation name and length, you have to know the form_url to continue. For this purpose, at normal loading, I have done the _get_form_url_after_monitors which returns 'launch_simulation' URL if the current burst_config from the database does not have an operation group and it returns the 'launch_pse' if it has. However, if you refreshed the page, you don't have a burst_config in the database yet so you can't do that. My idea was to go back to store the range params or something in the session and check if they exist or not. Basically, if you reached the setup_pse part, you can add something in the session which lets the program know when you refresh that it should take you back there and not try to launch a normal simulation. 

However, there is still at least one problem that this won't solve and that is, if you reached the last fragment (The one with name and length) and you refresh it there. Because you did not reach the pse part yet (Even though you might want to do it) AND you are past the first call to _get_form_url_after_monitors, the form_url will be launch_simulation and you even if you try to press on 'Setup PSE', it will still try to launch a normal simulation and fail.

I think we should discuss about this cause even besides this problem, I think that the way this logic of choosing between normal simulations and pse simulations is a little complicated. 

I am sorry if I couldn't explain the problem well enough, maybe we could have a talk about it tomorrow.

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -645,14 +646,17 @@ def set_noise_equation_params(self, **data):
...
         if operation_group is not None:
-            return SimulatorWizzardURLs.SETUP_PSE_URL
-        else:
-            return SimulatorWizzardURLs.LAUNCH_SIMULATION_URL
+            if loading is False:
+                return SimulatorWizzardURLs.SETUP_PSE_URL
+            else:
+                return SimulatorWizzardURLs.LAUNCH_PSE_URL
+
+        return SimulatorWizzardURLs.LAUNCH_SIMULATION_URL
```
popaula937, line 659: I think this is were your problems come from. LAUNCH_PSE_URL should not be used as an URL to go to from the NEXT button/action. And by NEXT action I refer to an action that gives us the next fragment in our wizzard.
In your case, the fragment with buttons Launch or SetupPSE should have as action the SETUP_PSE_URL, because that one represents the NEXT action, that gives us the next fragment.
Launch action is called in another manner from JS part.

framework_tvb/tvb/core/entities/storage/burst_dao.py
VinczeRobert: Okay, I checked this now. It does work with notlike in the sense that it doesn't give errors, but that would mean we are getting the number of bursts which don't have a certain name, when in actuality we woud need the number of bursts that do have for example "simulation_1" in their name, so for multiple branches, they would be named "simulation_1_branch1", "simulation_1_branch2" and so on.

liadomide: Let us imagine that we have `simlia` as the first BC.
We branch from it, we want automatic name `simlia_branch1` to be generated.
We branch again from the initial BC `simlia`, and we want automatic name `simlia_branch2` to be generated.
Now, we branch from `simlia_branch1` and the name generated should be `simlia_branch1_branch1`.
If we now branch from `simlia` a 3rd time!!, we want the name `simlia_branch3` to be generated. If you cut that `notlike`, instead it will become the wrong `simlia_branch4`, right?

Thus, pls just correct the previous query to the latest SQL syntax, and do not cut logic.

VinczeRobert: It's true that with "like" instead of "notlike" the behaviour is the one you described, but if I use notlike, then it's not good either: I made a simulation simlia, and then I launched two branches from that and both of them had the name "simlia_branch0" and then I launched a branched from "simlia_branch0" and it was named "simlia_branch0_branch1".

VinczeRobert:
I tested the changes that Paula made and they are all okay. Also, normal branching works, branch from a branch can also be launched succesfully. There are 3 things that still need to be done in this task:
1. The buttons that appear when we copy a simulation are not okay, there are are extra buttons that shouldn't be there.
2. When we copied a simulation and we are in the setup pse part and we hit previous, only the Previous button will be shown(Launch and SetupPSE won't be there).
3. We should make sure that branching and branching from a branch is also possible in the case of PSE simulations.

framework_tvb/tvb/core/entities/storage/burst_dao.py
liadomide: Then this should be fixed. 
In https://www.thevirtualbrain.org/user/ Default Project it woks as I describe it.

liadomide:
> 3. We should make sure that branching and branching from a branch is also possible in the case of PSE simulations.
I do not think this use-case makes much sense. We should be able from a Simple Single Simulation to branch into a PSE, but not to continue a PSE with a branch

framework_tvb/tvb/core/entities/storage/burst_dao.py
liadomide: with the current code (so without the second condition), it can not work as I described it.
See the `simlia_branch4` vs `simlia_branch3` last part

VinczeRobert:
> > 1. We should make sure that branching and branching from a branch is also possible in the case of PSE simulations.
> 
> I do not think this use-case makes much sense. We should be able from a Simple Single Simulation to branch into a PSE, but not to continue a PSE with a branch

Okay.

popaula937:
We will merge this PR because it is blocking PR#109. The remaining TODOs will be addressed in a follow-up PR.

PR #93, opened 2020-04-06
Pre-Gsoc20 prototype: TVB: Reusable visualization tools for Jupyter

umarbrowser:
Added the class of event handler for TVB: Reusable visualization tools for Jupyter


liadomide:
This is a bit against the spirit of tvb.simulator.plot, and I do not see its use-fullness, thus I close it.

PR #95, opened 2020-04-07
TVB-2649 Run REST client tests on jenkins

bvalean:


bvalean:
I added 2 new stages to the step1 jenkinsfile. First stage is starting the REST server and the second one is executing the rest client tests.

PR #97, opened 2020-04-09
Fix model variables selection step

VinczeRobert:
I have done this without using a class.

I made a private method in SimulatorController which builds a list of the chosen variable names associated with their indexes and I give the names to the forms and the indexes to the monitor. The if condition in the fill_from_post in MonitorForm is because we only want that to be executed at loading, when we actually have variables of interest in the monitor, we don't need that when setting the parameters.


popaula937:
Ok from my side, too. We could even make variables_of_interest required for MonitorForm init. 
Any thoughts?

PR #101, opened 2020-04-10
Suggested enhancements for neotraits H5 from DP

liadomide:


framework_tvb/tvb/core/neocom/_h5loader.py
```diff
@@ -83,8 +83,12 @@ class DirLoader(object):
 
     def __init__(self, base_dir, registry, recursive=False):
         # type: (str, Registry, bool) -> None
+        self.base_file = None
         if not os.path.isdir(base_dir):
-            raise IOError('not a directory {}'.format(base_dir))
+            self.base_file = base_dir
+            base_dir = os.path.dirname(base_dir)
```
liadomide, line 89: This code is miss-leading. 
We name `base_dir` the parameter, and then we cut one folder up. 
I suspect that it might be a file_path given here, but the it should be the responsibility of who is calling to API o adapt.

framework_tvb/tvb/core/neocom/_h5loader.py
```diff
@@ -128,19 +132,26 @@ def load(self, gid):
     def store(self, datatype):
         # type: (HasTraits) -> None
         h5file_cls = self.registry.get_h5file_for_datatype(type(datatype))
-        path = self.path_for(h5file_cls, datatype.gid)
+        if self.base_file is None:
+            path = self.path_for(h5file_cls, datatype.gid)
+        else:
+            path = self.base_file
+            self.base_file = None
```
liadomide, line 139: This code is again miss-leading. 
We might have at some point set `self.base_file`, to a not None value, and here, we use that value to store the current file, and regardless if we succeed or not, we set `self.base_file` to None. What are you trying to achieve?
Can't we better extract new methods on this Loader class? Or even create a different Loader subclass?

framework_tvb/tvb/core/neocom/h5.py
```diff
@@ -134,7 +137,7 @@ def load_from_dir(base_dir, gid, recursive=False):
     return loader.load(gid)
 
 
-def store_to_dir(base_dir, datatype, recursive=False):
+def store_to_dir(datatype, base_dir, recursive=False):
```
liadomide, line 140: If you change such signatures, than you should also change the `type` bellow, and the unit-test

framework_tvb/tvb/core/neocom/_h5loader.py
```diff
@@ -105,9 +105,12 @@ def find_file_name(self, gid):
...
         return fname
 
-    def load(self, gid):
+    def load(self, gid=None, fname=None):
         # type: (typing.Union[uuid.UUID, str]) -> HasTraits
-        fname = self.find_file_name(gid)
+        if fname is None:
+            if gid is None:
+                raise ValueError("Neither gid nor filename is provided to load!")
+            fname = self.find_file_name(gid)
```
liadomide: I suggest here to use the same paradign as in load, because we are in DirLoader, and we already have self.base_dir
`path = os.path.join(self.base_dir, fname)`

dionperd: I don't get it because both input fname and self.find_file_name(gid) return a file name, not a path
therefore it is enough to have join in here:
 with H5File.from_file(os.path.join(self.base_dir, fname)) as f:
Instead, in store() self.path_for(h5file_cls, datatype.gid) returns a full path, not just a file name

PR #102, opened 2020-04-10
TVB-2532 - Fix Creators (Large Conn, LocalConn, Stimulus)

adrian-dordea:


framework_tvb/tvb/interfaces/web/controllers/spatial/region_stimulus_controller.py
```diff
@@ -175,7 +176,7 @@ def step_2(self):
         template_specification = dict(title="Spatio temporal - Region stimulus")
         template_specification['mainContent'] = 'spatial/stimulus_region_step2_main'
         template_specification['next_step_url'] = '/spatial/stimulus/region/step_2_submit'
-        template_specification['regionStimSelectorForm'] = region_stim_selector_form
+        template_specification['regionStimSelectorForm'] = FlowController().get_template_dict(region_stim_selector_form)
```
popaula937, line 179: Maybe would be nicer to move the get_template_dict() method to BaseController, thus we could use it with self.get_template_dict() and won't be necessary to import FlowController in the other controllers.

liadomide: In fact there is one `def get_template_dict(self, adapter_form)` in the base class SpatioTemporalController inherited also here.
What is the difference between the two methods? I see the decorator is different and the Jinja template ... but did not looked further

liadomide: if we need both, then we will need for one a different name.

popaula937: Indeed, we need both for now. The template here is slightly different. 
The main difference is that we need to send more values at rendering time, while in FlowController we send only the form. Mostly because of the dynamic subforms.
This can be simplified and will be part of task TVB-2617.

For now, I propose to use a different name for method get_template_dict in SpatioTemporalController and move get_template_dict from FlowController to BaseController.
Thanks for pointing this out.

liadomide: I just did this change myself, and could not help to rename both methods :-)

framework_tvb/tvb/adapters/creators/connectivity_creator.py
```diff
@@ -114,27 +149,27 @@ def _store_related_region_mappings(self, original_conn_gid, new_connectivity_ht)
...
-        original_conn_ht = h5.load_from_index(original_connectivity)
+        original_conn_ht = h5.load_from_index(view_model.original_connectivity)
         assert isinstance(original_conn_ht, Connectivity)
 
-        if not is_branch:
-            new_conn_ht = self._cut_connectivity(original_conn_ht, numpy.array(new_weights),
-                                                 numpy.array(interest_area_indexes), numpy.array(new_tracts))
+        if not view_model.is_branch:
+            new_conn_ht = self._cut_connectivity(original_conn_ht, numpy.array(view_model.new_weights),
+                                                 numpy.array(view_model.interest_area_indexes), numpy.array(view_model.new_tracts))
```
popaula937, line 163: all these numpy.array() should not be necessary as view_model attributes are already defined as NArray

framework_tvb/tvb/adapters/creators/stimulus_creator.py
```diff
@@ -104,7 +105,8 @@ def get_input_name():
 
     @staticmethod
     def get_filters():
-        return None
+        return FilterChain(fields=[FilterChain.datatype + '.surface_type'], operations=["=="],
+                                 values=[CORTICAL])
```
liadomide, line 109: The filter is ok. In this file there is a TODO related to this filter. The TODO can be removed now, right?

framework_tvb/tvb/interfaces/web/controllers/spatial/region_stimulus_controller.py
```diff
@@ -50,6 +50,7 @@
 from tvb.interfaces.web.controllers import common
 from tvb.interfaces.web.controllers.decorators import handle_error, expose_page, expose_fragment, using_template, \
     check_user
+from tvb.interfaces.web.controllers.flow_controller import FlowController
```
liadomide, line 53: Is this import still necessary?

framework_tvb/tvb/interfaces/web/controllers/spatial/surface_stimulus_controller.py
```diff
@@ -48,6 +48,7 @@
 from tvb.interfaces.web.controllers import common
 from tvb.interfaces.web.controllers.decorators import expose_page, expose_json, expose_fragment, using_template, \
     handle_error, check_user
+from tvb.interfaces.web.controllers.flow_controller import FlowController
```
liadomide, line 51: This import?

popaula937:
Suggestions regarding PRs:
- in the future please use rebase instead of merge in order to synchronize your branch with master. Otherwise, we get all those 'merge commits' in PRs
- you can squeeze more commits into a single one. It would be nice to squeeze your last commits into a 'TVB-2532 Review' for example

PR #107, opened 2020-04-14
TVB-2658: Got rid of disturbing messages on Simulator Page

VinczeRobert:
I deleted a lot of functions that are not used anymore in bursts.js and range.js. I ran the tests and it seems they were not affected by my changes. I also tried all kinds of things in the UI, so I think it's okay to delete all these.

Also:
1. I only left one method in range.js, but could that be deleted too? It's only used in some stepsUpdate file once.
2. Can I delete all JS methods and HTML elements related to portlets as well? (I understood that the Python methods were already deleted).


popaula937:
1. It should be safe to delete also the method left in range.js and the usages of _disableDomForRangeComponent_ByValue from stepUpdate.
2. At least those that do Ajax calls to methods that do not exist anymore can be removed.

framework_tvb/tvb/interfaces/web/templates/jinja2/flow/type2component/number2component.html
```diff
@@ -41,12 +41,6 @@
                 <tr> <td colspan="2" id="{{ containerDivId }}_rangeLabelsTd"> </td> </tr>
             </table>
 
-            <p class="field-adapters">
```
liadomide, line 44: I think all templates under `type2component` folder are no longer in use, right?
If that is the case, they should be removed

VinczeRobert: From my part this task is done. Please let me know if there is anything else I can delete.

liadomide: the rebase seems to be done in the wrong direction. Unfortunately now the changes list is too difficult to read. Could you cleanup history?

VinczeRobert: Done.

PR #110, opened 2020-04-15
Put back tvb.analyzers.info

liadomide:
Last year, when we ensured TVB becomes compatible with Python 3, we removed `tvb.analyzers.info` module, as it was not working with recent numpy and python versions, and we thought it is also not used.
I think we made a mistake in judgment. The module was part of TVB api, and is even referred in one of the Tutorials (true, from the "sandbox" folder).

I am thinking to put the module back. But then, we will need to also adapt the code.
Currently there is a problem at line 87:
`  File "...tvb-root/scientific_library/tvb/analyzers/info.py", line 87, in sampen
    y = y[:y.shape[0] / taus * taus].reshape((-1, taus)).mean(axis=1)
TypeError: slice indices must be integers or None or have an __index__ method`
If I edit that line into:
`y = y[:int(y.shape[0] / taus * taus)].reshape((-1, taus)).mean(axis=1)`
Then the error becomes:
`ValueError: cannot reshape array of size 29999 into shape (7)`

@maedoc Could you pls have a quick look, and advice how to proceed further?

### DIFF
--- a/scientific_library/tvb/analyzers/info.py
+++ b/scientific_library/tvb/analyzers/info.py
@@ -84,7 +84,7 @@ def sampen(y, m=2, r=None, qse=False, taus=1, info=False,
 
     # if we have a scale factor, coarsen time series 
     if taus > 1:
-        y = y[:y.shape[0] / taus * taus].reshape((-1, taus)).mean(axis=1)
+        y = y[:y.shape[0] // taus * taus].reshape((-1, taus)).mean(axis=1)
 
     # compute embedding of signal dims m, m+1, initialize match counts to 0
     Y1 = subseq(y, m)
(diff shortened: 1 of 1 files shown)


scientific_library/tvb/analyzers/info.py
```diff
@@ -0,0 +1,112 @@
...
+    # helper function to reformat arrays for matching
+    subseq = lambda y, n: y[tile(r_[0:n], (y.size - n + 1, 1)) + r_[0:y.size - n + 1][:, na]]
+
+    # default value of r
+    if r is None:
+        r = 0.15 * y.std()
+
+    # if we have a scale factor, coarsen time series 
+    if taus > 1:
+        y = y[:y.shape[0] / taus * taus].reshape((-1, taus)).mean(axis=1)
```
maedoc, line 87: this line is a problem in Py 3 because `/` produces a float and not integer division as in Py 2.  If this is changed to `y.shape[0] // taus`, it should be fine.

maedoc:
@liadomide this looks good to go

PR #114, opened 2020-04-16
TVB-2662. Extend tvb-rest-client with helpers

popaula937:
Toni needs:
- a helper to launch an operation based on datatype_gid and algorithmDTO
- a way to access the algorithmDTO displayname from an OperationDTO

### DIFF
--- a/framework_tvb/tvb/interfaces/rest/client/examples/fire_simulation.py
+++ b/framework_tvb/tvb/interfaces/rest/client/examples/fire_simulation.py
@@ -118,6 +118,26 @@ def fire_simulation_example(tvb_client_instance):
         assert chunk.shape[2] == data_shape[2]
         assert chunk.shape[3] == data_shape[3]
 
+        return project_gid, time_series_gid
+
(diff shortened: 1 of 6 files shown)


framework_tvb/tvb/interfaces/rest/client/examples/fire_simulation.py
```diff
@@ -118,11 +118,32 @@ def fire_simulation_example(tvb_client_instance):
         assert chunk.shape[2] == data_shape[2]
         assert chunk.shape[3] == data_shape[3]
 
+        return project_gid, time_series_gid
+
+
+def quick_launch_an_operation(tvb_client_instance, project_gid, time_series_gid):
```
liadomide, line 124: Maybe rename `time_series_gid ` into something a bit more generic?
It could be any `datatype_gid`, right?

popaula937: yes, sure

PR #115, opened 2020-04-21
TVB-2666 Uniform column names in BurstConfiguration and *Index classes

liadomide:
We want all Foreign Key columns to be named under patter: `fk_...`
The new columns which are Foreign Keys towards a gid will be named `fk_...._gid`.


liadomide:
@popaula937 From my pov this PR is ready for merge. 
Should we do it before the end of the current sprint or after?

PR #116, opened 2020-04-22
TVB-2555: Fix Loading Parameters When Copying A Simulation

VinczeRobert:
About those "is_simulator_copy = False" deletions. I understood that branching has to be possible even if you've done for example Copy -> Previous -> Next -> Branch (and that we will deal with this case in another task). With those there, this won't be possible cause if is_simulator_copy is False, the button won't show up and in case of normal simulation launching that variable will be False regardless of this statement.

The buttons all show up fine, but there is a problem that I receive an error when trying to PSE launch a copied simulation (same error for both simulation types):
Traceback (most recent call last):
  File "d:\tvb\tvb-root\framework_tvb\tvb\interfaces\web\controllers\decorators.py", line 147, in deco
    return func(*a, **b)
  File "d:\tvb\tvb-root\framework_tvb\tvb\interfaces\web\controllers\decorators.py", line 193, in deco
    return func(*a, **b)
  File "d:\tvb\tvb-root\framework_tvb\tvb\interfaces\web\controllers\simulator_controller.py", line 904, in launch_pse
    burst_config.operation_group = operation_group
  File "C:\Users\Robert.Vincze\Anaconda3\envs\tvb-root-p3\lib\site-packages\sqlalchemy\orm\attributes.py", line 224, in __set__
    instance_dict(instance), value, None)
  File "C:\Users\Robert.Vincze\Anaconda3\envs\tvb-root-p3\lib\site-packages\sqlalchemy\orm\attributes.py", line 797, in set
    LOAD_AGAINST_COMMITTED)
  File "C:\Users\Robert.Vincze\Anaconda3\envs\tvb-root-p3\lib\site-packages\sqlalchemy\orm\attributes.py", line 584, in get
    value = self.callable_(state, passive)
  File "C:\Users\Robert.Vincze\Anaconda3\envs\tvb-root-p3\lib\site-packages\sqlalchemy\orm\strategies.py", line 530, in _load_for_state
    (orm_util.state_str(state), self.key)
sqlalchemy.orm.exc.DetachedInstanceError: Parent instance <BurstConfiguration at 0x165e13cfa08> is not bound to a Session; lazy load operation of attribute 'operation_group' cannot proceed.

Should I solve it in this task or is this going to be another one?

Also, a smaller problem regarding the simulation name... for example if I just launch a copied simulation without doing anything else, it will have the same name. Shouldn't we do something like "originalname_copy1"?

### DIFF
--- a/framework_tvb/tvb/core/entities/storage/burst_dao.py
+++ b/framework_tvb/tvb/core/entities/storage/burst_dao.py
@@ -80,8 +80,8 @@ class BurstDAO(RootDAO):
         try:
             count = self.session.query(BurstConfiguration
                                     ).filter_by(fk_project=project_id
-                                    ).filter(BurstConfiguration.name.like(burst_name + '%')
-                                    ).filter(BurstConfiguration.name.notlike(burst_name + '/_%/_%', escape='/')
+                                    ).filter(BurstConfiguration.name.like(burst_name + '_branch%')
+                                    ).filter(BurstConfiguration.name.notlike(burst_name + '_branch%_branch%', escape='/')
                 ).count()
         except SQLAlchemyError as excep:
             self.logger.exception(excep)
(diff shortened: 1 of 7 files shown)


popaula937:
I did not review this completely but I want to say that launching PSE from a copy works in my current branch, so this might be fixed once we merge the PR.
I also noticed the simulation name is not updated at copy time. We should handle that.
And at first sight, I don't think we should remove all those "is_simulator_copy = False", but I will come back with an answer once I manage to do the review.

PR #118, opened 2020-04-22
TVB-2581: Improve LocalConnectivity creator

adrian-dordea:
LocalConnectivity creator: 
- show a nice message, to inform the user what to do when missing Cortical Surfaces from the current project
- cleanup TODOs

### DIFF
--- a/framework_tvb/tvb/interfaces/web/controllers/spatial/local_connectivity_controller.py
+++ b/framework_tvb/tvb/interfaces/web/controllers/spatial/local_connectivity_controller.py
@@ -44,8 +44,8 @@ from tvb.adapters.simulator.equation_forms import GAUSSIAN_EQUATION, DOUBLE_GAUS
 from tvb.core.entities.storage import dao
 from tvb.core.neocom import h5
 from tvb.core.adapters.abcadapter import ABCAdapter
+from tvb.core.services.exceptions import ServicesBaseException
 from tvb.datatypes.surfaces import CORTICAL
-from tvb.interfaces.rest.commons.exceptions import ServiceException
 from tvb.interfaces.web.controllers import common
 from tvb.interfaces.web.controllers.base_controller import BaseController
 from tvb.interfaces.web.controllers.decorators import check_user, handle_error, using_template
@@ -328,7 +328,7 @@ class LocalConnectivityController(SpatioTemporalController):
             surface_gid = current_lconn.surface.hex
             surface = ABCAdapter.load_entity_by_gid(surface_gid)
             if surface is None:
-                raise ServiceException("There is no surface in the current project. Please upload one to continue.")
+                raise ServicesBaseException("There is no surface in the current project. Please upload one to continue.")
             max_x = current_lconn.cutoff
             if max_x <= 0:
                 max_x = 50
--- a/framework_tvb/tvb/interfaces/web/controllers/spatial/local_connectivity_controller.py
+++ b/framework_tvb/tvb/interfaces/web/controllers/spatial/local_connectivity_controller.py
@@ -29,6 +29,7 @@
 #
 
 """
+.. moduleauthor:: Paula Popa <paula.popa@codemart.ro>
 .. moduleauthor:: Bogdan Neacsa <bogdan.neacsa@codemart.ro>
 .. moduleauthor:: Ionel Ortelecan <ionel.ortelecan@codemart.ro>
 """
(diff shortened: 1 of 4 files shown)


framework_tvb/tvb/interfaces/web/controllers/spatial/local_connectivity_controller.py
```diff
@@ -51,6 +51,8 @@
 from tvb.interfaces.web.controllers.decorators import expose_fragment, expose_page, expose_json
 from tvb.interfaces.web.controllers.spatial.base_spatio_temporal_controller import SpatioTemporalController
 
+from framework_tvb.tvb.core.entities.file.exceptions import MissingDataSetException
```
popaula937, line 54: This import is not correct, it should be: from tvb.core.entities....

framework_tvb/tvb/interfaces/web/controllers/spatial/local_connectivity_controller.py
```diff
@@ -44,6 +44,7 @@
 from tvb.core.entities.storage import dao
 from tvb.core.neocom import h5
 from tvb.core.adapters.abcadapter import ABCAdapter
+from tvb.core.services.exceptions import ServicesBaseException
```
liadomide, line 47: This is a service layer exception and we are in a controller.
For sure we have exceptions for controllers. If none appropriate exist, just create one.

framework_tvb/tvb/interfaces/web/controllers/spatial/local_connectivity_controller.py
```diff
@@ -95,7 +96,6 @@ def step_1(self, do_reset=0, **kwargs):
             if default_surface_index:
                 new_lconn.surface = uuid.UUID(default_surface_index.gid)
             else:
-                # TODO: ok to keep a default gid here?
                 new_lconn.surface = uuid.uuid4()
```
liadomide, line 99: This line makes not much sense `new_lconn.surface = uuid.uuid4()`. 
Why not put here `common.set_error_message("There ....` ?

framework_tvb/tvb/interfaces/web/controllers/spatial/local_connectivity_controller.py
```diff
@@ -326,6 +327,8 @@ def get_equation_chart(self):
             current_lconn = common.get_from_session(KEY_LCONN)
             surface_gid = current_lconn.surface.hex
             surface = ABCAdapter.load_entity_by_gid(surface_gid)
+            if surface is None:
```
liadomide, line 330: With my above suggestion then this check will change into `if current_lconn.surface is None:` and should be move upper.

framework_tvb/tvb/interfaces/web/templates/jinja2/spatial/equation_displayer.html
```diff
@@ -22,7 +22,7 @@
     {% else %}
         <script type="text/javascript">
             $(document).ready(function () {
-                displayMessage('{{ errorMsg }}', "warningMessage");
+                displayMessage('{{ errorMsg }}', "errorMessage");
```
liadomide, line 25: Have you checked which of the 2 messages is displayed here?

framework_tvb/tvb/interfaces/web/controllers/spatial/local_connectivity_controller.py
adrian-dordea: that line is used after in common.add2session(KEY_LCONN, new_lconn) which is used after in current_lconn, and is we don't have that we get a error when we do: configure_lcon_form.fill_from_trait(current_lconn) that redirects the user to the generic error page.

framework_tvb/tvb/interfaces/web/controllers/spatial/local_connectivity_controller.py
adrian-dordea: because if we do not remove the new_lconn.surface = uuid.uuid4() from the above comment he still have chere current_lconn.surface, and the surface is the one that is None

framework_tvb/tvb/interfaces/web/templates/jinja2/spatial/equation_displayer.html
adrian-dordea: yes, the one from get_equation_chart

PR #119, opened 2020-04-23
TVB-2657: Simulations with Multiple Monitors

VinczeRobert:
This task is far to be done yet, but I wanted to put my work so far so you can see it.

Currently I was able to launch a simulation with three monitors that are not Bold. I need to write the logic to handle bold as well and to expand the logic for loading. Also, probably changing the names of the UI parameters would be a good idea. It looks confusing to have three 'Sampling Periods' for example, so I think we could have something like 'Sampling Period 1', 'Sampling Period 2', 'Sampling Period 3' etc...

### DIFF
--- a/framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
+++ b/framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
@@ -654,6 +654,16 @@ class SimulatorController(BurstBaseController):
             indexes[variable] = all_variables.index(variable)
         return indexes
 
+    @staticmethod
+    def _build_list_of_monitors(monitor_names, is_surface_simulation):
+        monitor_dict = get_ui_name_to_monitor_dict(is_surface_simulation)
+        monitor_classes = []
+
+        for monitor in monitor_names:
+            monitor_classes.append(monitor_dict[monitor]())
+
+        return monitor_classes
+
     @cherrypy.expose
     @using_template("simulator_fragment")
     @handle_error(redirect=False)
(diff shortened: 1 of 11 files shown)


VinczeRobert:
I am almost done with this task. So far I have managed to handle all special cases, like when you choose both Bold monitors or when you choose Raw. It seems like loading and launching work with any combination (I will test them at the end again and I know I have to adapt the tests to the changes as well).

I have also added legends to the monitor forms, because this way it is clear at which monitor's parameters is the user looking, otherwise it was confusing even to me. I think they look good this way. 

Paula told me that there are already too many methods in SimulatorController and I have just added some more. I can move some of them in other places if it is required. 

There is just one problem remaining, to which I will tell my idea. So far I have stored the index of the current monitor in the session so the program is able to get the next monitor parameters form. The problem with this is that if you hit Previous, the index can't be decreased. My idea is to send the index as a parameter to set_monitor_params and set_monitor_equation from Javascript. We would only need an if there and a place to keep the index. I think this is good, because when you hit previous, the previous_wizzard_submit method is called, so the index could be decreased there.

VinczeRobert:
I am done with this task. The only reason I still left the WIP tag is because I have to fix the tests and I am waiting for a code review, since there are a lot of changes.

Besides the incapability to update the current_monitor_index previously stored in session by me, another problem was that all the monitor param fragments were having the same id. Instead of this index stored in session, each fragment will have it's url updated (for example: 'burst/set_monitor_params/TemporalAverage'). This fixes both problems. The major new thing with this change is that set_monitor_params and set_monitor_equation have now a new parameters: current_monitor (it's only a string with the name of the class).

popaula937:
Will review after code is up to date and conflicts are resolved.

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -660,28 +678,44 @@ def set_monitors(self, **data):
...
             fragment = SimulatorMonitorFragment(is_surface_simulation=session_stored_simulator.is_surface_simulation)
             fragment.fill_from_post(data)
 
-            session_stored_simulator.monitors = [fragment.monitor.value()]
+            session_stored_simulator.monitors = self._build_list_of_monitors(
+                fragment.monitors.value, session_stored_simulator.is_surface_simulation)
+
+
+
+        # if the first monitor is Raw, it must be skipped because it does not have parameters
```
popaula937, line 690: Maybe extract this in a separate method exclude_raw_monitor_fragment()

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -113,7 +109,8 @@ class SimulatorFragmentRenderingRules(object):
     def __init__(self, form=None, form_action_url=None, previous_form_action_url=None, is_simulation_copy=False,
                  is_simulation_readonly_load=False, last_form_url=SimulatorWizzardURLs.SET_CONNECTIVITY_URL,
                  last_request_type='GET', is_first_fragment=False, is_launch_fragment=False, is_model_fragment=False,
-                 is_surface_simulation=False, is_noise_fragment=False, is_launch_pse_fragment=False, is_pse_launch=False):
+                 is_surface_simulation=False, is_noise_fragment=False, is_launch_pse_fragment=False, is_pse_launch=False,
+                 is_monitor_params_fragment=False, monitor_name=None):
```
popaula937, line 113: If the only purpose of flag is_monitor_params_fragment is to display the monitor_name in the template, then we can remove it, and check if monito_name is not None directly.

PR #122, opened 2020-04-24
TVB-2581 Fix behavior of creators in TVB Web when launched in an empty project

liadomide:
LC, Surface Stimulus and Region Stimulus pages throw generic errors when opened in  an empty project.
Instead we should show explicit error messages.
Fix some unit-tests

### DIFF
--- a/framework_tvb/tvb/interfaces/web/controllers/spatial/region_stimulus_controller.py
+++ b/framework_tvb/tvb/interfaces/web/controllers/spatial/region_stimulus_controller.py
@@ -34,20 +34,20 @@
 """
 
 import json
-import uuid
 import cherrypy
 import numpy
-from tvb.adapters.creators.stimulus_creator import RegionStimulusCreatorForm, RegionStimulusCreator, \
-    StimulusRegionSelectorForm, RegionStimulusCreatorModel
+from tvb.adapters.creators.stimulus_creator import *
 from tvb.adapters.datatypes.h5.patterns_h5 import StimuliRegionH5
 from tvb.adapters.simulator.equation_forms import get_ui_name_to_equation_dict, get_form_for_equation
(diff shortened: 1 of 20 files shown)


framework_tvb/tvb/interfaces/web/controllers/spatial/region_stimulus_controller.py
```diff
@@ -133,16 +136,19 @@ def step_1(self):
...
+        project_id = common.get_current_project().id
+        region_stim_selector_form = StimulusRegionSelectorForm(project_id)
         region_stim_selector_form.region_stimulus.data = selected_stimulus_gid
         region_stim_selector_form.display_name.data = common.get_from_session(KEY_REGION_STIMULUS_NAME)
 
-        region_stim_creator_form = RegionStimulusCreatorForm(self.equation_choices, common.get_current_project().id)
+        region_stim_creator_form = RegionStimulusCreatorForm(self.equation_choices, project_id)
         if not hasattr(current_stimuli_region, 'connectivity') or not current_stimuli_region.connectivity:
-            current_connectivity_in_form = region_stim_creator_form.connectivity._get_values_from_db()[0]
-            region_stim_creator_form.connectivity.data = current_connectivity_in_form[2]
```
liadomide, line 143: @popaula937 is it ok to remove this line, considering that we put the GID on the ViewModel, and then we do `form.fiil_from_trait` ?

framework_tvb/tvb/tests/framework/core/factory.py
```diff
@@ -41,22 +41,23 @@
 import os
 import random
 import tvb_data
```
liadomide, line 43: @VinczeRobert From the history of commits you seem to have worked a bit on this Factory class. 
Pls have a quick look at this file refactor ... and the code could be even further improved ;-)

PR #123, opened 2020-04-25
TVB-2674 and TVB-2667 Review users retrieval and project membership

liadomide:
- Allow using different page sizes on retrieval of users 
  * smaller page size for users management
  * more for project members in web GUI and REST api
- when creating a new project (either from GUI or REST) always make the admin also a member

### DIFF
--- a/framework_tvb/tvb/interfaces/rest/server/resources/user/user_resource.py
+++ b/framework_tvb/tvb/interfaces/rest/server/resources/user/user_resource.py
@@ -91,13 +91,7 @@ class GetUsersResource(RestResource):
         """
         :return: a list of TVB users
         """
-        page_number = flask.request.args.get(Strings.PAGE_NUMBER.value)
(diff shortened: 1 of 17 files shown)


framework_tvb/tvb/interfaces/rest/server/resources/user/user_resource.py
```diff
@@ -91,7 +91,15 @@ def get(self):
...
         :return: a list of TVB users
         """
-        user_dto_list, pages_no = UserFacade.get_users(get_current_user().username)
+        page_number = flask.request.args.get(Strings.PAGE_NUMBER.value)
+        if page_number is None:
+            page_number = 1
+        try:
+            page_number = int(page_number)
+        except ValueError:
+            raise InvalidInputException(message="Invalid page number")
```
popaula937, line 100: We could extract this to a validate_page_number method

PR #128, opened 2020-04-28
TVB-2652. Prepare new versions for pypi packages.

popaula937:


liadomide:
Could you pls also change VersioSettings.BASE_VERSION ? in the code?

liadomide:
and maybe manually increase tvb.version number by some integer.... ?

PR #131, opened 2020-04-29
TVB-2680 Fix Histogram Viewer

liadomide:
The fix for the HistogramViewer consist of changes in `histogram.py` and `graph_h5.py` files only.
Rest of the changes in this PR are for similar problems, or reviewing Indexing classes.

### DIFF
--- a/framework_tvb/tvb/adapters/datatypes/h5/connectivity_h5.py
+++ b/framework_tvb/tvb/adapters/datatypes/h5/connectivity_h5.py
@@ -56,7 +56,7 @@ class ConnectivityH5(H5File):
     def get_region_labels(self):
         return self.region_labels.load()
 
-    def store(self, datatype, scalars_only=False, store_references=False):
+    def store(self, datatype, scalars_only=False, store_references=True):
         # type: (Connectivity, bool, bool) -> None
         super(ConnectivityH5, self).store(datatype, scalars_only, store_references)
         self.region_labels.store(datatype.region_labels.astype(STORE_STRING))
--- a/framework_tvb/tvb/core/adapters/abcadapter.py
+++ b/framework_tvb/tvb/core/adapters/abcadapter.py
@@ -468,7 +468,7 @@ class ABCAdapter(object):
         """
         Load a generic HasTraits instance, specified by GID.
         """
-        index = ABCAdapter.load_entity_by_gid(data_gid.hex)
+        index = ABCAdapter.load_entity_by_gid(data_gid)
         return h5.load_from_index(index)
 
     @staticmethod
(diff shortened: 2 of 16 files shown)


framework_tvb/tvb/adapters/datatypes/h5/connectivity_h5.py
```diff
@@ -56,7 +56,7 @@ def get_centres(self):
     def get_region_labels(self):
         return self.region_labels.load()
 
-    def store(self, datatype, scalars_only=False, store_references=False):
+    def store(self, datatype, scalars_only=False, store_references=True):
```
liadomide, line 59: @popaula937 I found the signature from parent wrongly copied in few H5File subclasses.
It seems to prevent in some cases (e.g from store_complete) storage of Reference attributes (e.g. ConnectivityMeasure.connectivity).
In some other classes, like here on the ConnectivityH5 the missmatch in signature has no effect, as this entity contains no reference.
I changes in all places of DataTypes the signature, for uniformity, except for Simulator classes (Sim, Burst, Monitors...). Could you pls consider these classes, and judge if   those need changes or not?

framework_tvb/tvb/core/adapters/abcadapter.py
```diff
@@ -463,17 +464,17 @@ def load_entity_by_gid(data_gid):
 
     @staticmethod
     def load_traited_by_gid(data_gid):
-        # type: (uuid.UUID) -> HasTraits
+        # type: (typing.Union[uuid.UUID, str]) -> HasTraits
         """
         Load a generic HasTraits instance, specified by GID.
         """
-        index = load_entity_by_gid(data_gid.hex)
+        index = ABCAdapter.load_entity_by_gid(data_gid.hex)
```
popaula937, line 471: .hex should be removed here

framework_tvb/tvb/adapters/datatypes/h5/connectivity_h5.py
popaula937: Sure, looking into it. Thanks for this nice review!

popaula937: Overriding is not necessary for Monitors.
For SimulatorH5 and IntegratorH5, I will do a review soon. Since we now generate the GID on HasTraits, we don't need to wait for their dependencies to be written to H5 before being able to get the GID, so store_reference=False will no longer be necessary.

liadomide: Indeed, the `store_reference=False` flag was making sense when we were not having GID on `HasTraits`, and it had to be filled afterwards. Thanks for explaining why this code!

framework_tvb/tvb/core/adapters/abcadapter.py
liadomide: Changing this line (including Paula's suggestion) actually fixed one unit-test, without us actively expecting that 👍

PR #133, opened 2020-05-02
TVB-2648 TVB-2652 Review

liadomide:
- Adjust Readme files as they appear on Pypi pages
- Update Simulator class diagrams
- translate to neotraits the script generating Phase Plane images for models. It works and generates similar images with prev version, except for ReducedSetHindmarshRose


maedoc:
If it's working for models except Reduced* then it is likely related to modes and a missing reshape.  What do you think?  Sorry, I don't currently have time to check everything out run and test..

liadomide:
> If it's working for models except Reduced* then it is likely related to modes and a missing reshape. What do you think? Sorry, I don't currently have time to check everything out run and test..

No worry. I didn't hoped for an immediate testing (that is the reason why I also already merged this), I merely wanted to document the current state of that script. 
I didn't change the Phase Plane images, as they should not have relevant changes anyway (no change in model equations was intended), and re-runing this script helped internally to "prove" that no significant change appeared.
Indeed, if I look closely to the differences they are related to the randomness of the initial conditions selection, plus the variety of min/max on the axes that is linked to that (in the case of not largely set initially in the model).

PR #134, opened 2020-05-02
TVB-2684 TVB-2679 Review FlowService, add ValueWrapper, unit-tests fixes

liadomide:
Starting from `FlowService.get_available_datatypes` method (which is only used in unit-tests, thus it can be removed), I looked into linked unit-tests (and fixed where the unit-test was deprecated), then in the problems revealed by those unit-test still failing (and reported as extra task, or directly fix, where fast)

### DIFF
--- a/framework_tvb/tvb/tests/framework/core/services/links_test.py
+++ b/framework_tvb/tvb/tests/framework/core/services/links_test.py
@@ -203,28 +203,37 @@ class TestImportExportProjectWithLinksTest(_BaseLinksTest):
         assert 1 == len(links)
         assert self.red_datatype.gid == links[0].gid
 
-    def test_create_interlinked_projects(self, initialize_linked_projects, value_wrapper_factory):
-        """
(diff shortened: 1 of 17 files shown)


popaula937:
Looks good to me until now, I will take another look when the PR is finalized.

framework_tvb/tvb/tests/framework/core/services/links_test.py
```diff
@@ -208,39 +204,35 @@ def test_datatypes_and_links_recreated_on_import(self, transactional_setup_fixtu
         assert self.red_datatype.gid == links[0].gid
 
     @pytest.fixture()
-    def create_interlinked_projects(self, region_mapping_factory, time_series_region_index_factory):
```
liadomide, line 211: @VinczeRobert Could you explain how did this work before?

liadomide:
From my side this is done.
Some tests changed in this PR are still failing, but due to documented other Tasks in Jira.

PR #137, opened 2020-05-05
TVB-2676: Solved problem with display_name 

VinczeRobert:
I solved this problem by looking at how it's done in Local Connectivity Creator, which also has a display_name. So I put a display_name parameter on the ViewModel and then assigned it to generic_attributes.user_tag1.

I want to mention that this is not the only problem regarding this creator. The equation parameters are also not updated and saved to the DB (this applies to Local Connectivity Creator as well) and if I go forward to set scaling and then go back, the equation is kept, but the parameters not (let's say I changed it to Alpha, when I go back the equation will still be Alpha but it will have PulseTrain's parameters).


liadomide:
I think the Equations problem is related to TVB-2617

PR #138, opened 2020-05-05
TVB-2617. Refactor dynamic subforms

popaula937:
Redraw trajectories in PhasePlane is not always correct. I suspect that data is not correctly updated in session. Will be fixed as part of TVB-2553.
Removed SimpleSelectField from forms and adjusted usages (part of TVB-2669).


framework_tvb/tvb/interfaces/web/controllers/spatial/region_stimulus_controller.py
```diff
@@ -113,20 +116,20 @@ def set_display_name(self, **param):
...
     @handle_error(redirect=False)
     @check_user
-    def set_temporal(self, temporal_equation):
+    def refresh_subform(self, temporal_equation, mapping_key):
         eq_class = get_ui_name_to_equation_dict().get(temporal_equation)
         current_region_stim = common.get_from_session(KEY_REGION_STIMULUS)
         current_region_stim.temporal = eq_class()
 
-        eq_params_form = get_form_for_equation(eq_class)(prefix=RegionStimulusCreatorForm.NAME_TEMPORAL_PARAMS_DIV)
+        eq_params_form = SubformHelper().get_subform_for_field_value(temporal_equation, mapping_key)
```
liadomide, line 124: Could such methods be static ?

popaula937: yes, changed then to static

popaula937:
Changes the way we choose a template for select field, so we don't need those expressions everywhere.
Will review the base_url before closing this.

framework_tvb/tvb/interfaces/web/controllers/spatial/surface_model_parameters_controller.py
```diff
@@ -187,7 +187,7 @@ def edit_model_parameters(self):
         Main method, to initialize Model-Parameter visual-set.
         """
         model, cortex = self.get_data_from_burst_configuration()
-        surface_gid = cortex.fk_surface_gid
+        surface_gid = cortex.surface_gid
```
liadomide, line 190: Ups. This was me! Sorry :)

popaula937:
Will be merged after @VinczeRobert has tested these.

VinczeRobert:
I have tested the areas that I was requested to and those were: Stimulus and Surface Viewers, Local Connectivity Viewer, SurfaceModelConfigurations and features in SimulatorController. Most of it is alright, but I did find a few minor issues (I double-checked them to be sure):
- For the first one I don't know if it's really a bug or intended behavior, but at the Surface Stimulus Creator if you click to Edit Focal Points and then you want to go back to by clicking on Edit Stimulus Equations, it doesn't allow you if you haven't set focal points. So far I thought that not setting any focal points would only stop you from saving the new stimulus on surface. Please correct me if I'm wrong.
- Setting up a simulation with surface lets you set up the surface model correctly but when you are done with it, it skips a fragment. You should go on with setting the model params, but instead it jumps to the next one.
- Simulation with only one range param doesn't work. This was fixed once, but it looks like something broke it again...
- Lastly I tried to launch some more complicated simulations (changing a lot of params and trying interesting combinations) and among these I tried a surface Simulation with EEG monitor, which did not work, because: "'TimeSeriesEEGH5' object has no attribute 'surface'".

These are the  things I noticed, the rest seems fine, all parameters in the creators are loaded correctly and are kept.

popaula937:
Thanks for testing!
1. That is how it works since TVB with python2. Maybe we could review that behavior.
2. That is the intended behavior since you already configured the model params on the SetupSurfaceModel page.
3. Will look into it, not sure it is broken by current task.
4. Will look into it, maybe it will be a new task in Jira.

PR #139, opened 2020-05-05
TVB-2689: add summary_info to DataType indexes

adrian-dordea:
After the split in multiple layers on datatypes, we no longer had "scientific Details" in a DataType overlay. We need to add something similar with the previous summary_info on DataType indexes.

### DIFF
--- a/framework_tvb/tvb/core/entities/model/model_datatype.py
+++ b/framework_tvb/tvb/core/entities/model/model_datatype.py
@@ -113,19 +113,18 @@ class DataType(HasTraitsIndex):
     fk_from_operation = Column(Integer, ForeignKey('OPERATIONS.id', ondelete="CASCADE"))
     parent_operation = relationship(Operation, backref=backref("DATA_TYPES", order_by=id, cascade="all,delete"))
 
+    @property
     def summary_info(self):
         # type: () -> typing.Dict[str, str]
 
-        cls = type(self)
-        ret = {'Type': cls.__name__}
+        ret = {}
         if self.title:
             ret['title'] = str(self.title)
 
-        for attribute in cls.__dict__:
+        for attribute in self.__mapper__.column_attrs.keys():
             try:
                 attr_field = getattr(self, attribute)
-                if attr_field is not None:
-                    ret[attribute] = attr_field
+                ret[attribute] = str(attr_field)
             except Exception:
                 ""
         return ret
--- a/framework_tvb/tvb/core/entities/model/model_datatype.py
+++ b/framework_tvb/tvb/core/entities/model/model_datatype.py
@@ -121,14 +121,31 @@ class DataType(HasTraitsIndex):
         if self.title:
             ret['Title'] = str(self.title)
 
-        for attribute in self.__table__.columns.keys():
+        columns = self.get_base_table_columns
+        for attribute in columns:
             try:
                 attr_field = getattr(self, attribute)
-                attr_name = attribute.title().replace("_", " ")
-                ret[attr_name] = str(attr_field)
+                attr_name = self.get_attribute_name(attribute)
+                if attr_name is not None:
+                    ret[attr_name] = str(attr_field)
             except Exception:
                 pass
         return ret
+
+    @property
+    def get_base_table_columns(self):
+        columns = self.__table__.columns.keys()
+        if type(self).__bases__[0] is DataType:
+            return columns
+        base_table_columns = type(self).__bases__[0].__table__.columns.keys()
+        columns.extend(base_table_columns)
+        return columns
+
+    def get_attribute_name(self, attr_name):
+        #don't display fk_ attributes or id(dispayed in generic)
+        if "fk_" in attr_name or "id" in attr_name:
+            return None
+        return attr_name.title().replace("_", " ")
 
     def __init__(self, gid=None, **kwargs):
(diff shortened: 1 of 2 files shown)


framework_tvb/tvb/core/entities/model/model_datatype.py
```diff
@@ -113,6 +113,23 @@ class DataType(HasTraitsIndex):
     fk_from_operation = Column(Integer, ForeignKey('OPERATIONS.id', ondelete="CASCADE"))
     parent_operation = relationship(Operation, backref=backref("DATA_TYPES", order_by=id, cascade="all,delete"))
 
+    def summary_info(self):
+        # type: () -> typing.Dict[str, str]
+
+        cls = type(self)
+        ret = {'Type': cls.__name__}
```
liadomide, line 120: I do not think we need the 'type' or 'class', as this is displayed on the top of the overlay page.

framework_tvb/tvb/core/entities/model/model_datatype.py
```diff
@@ -113,6 +113,23 @@ class DataType(HasTraitsIndex):
...
+        if self.title:
+            ret['title'] = str(self.title)
+
+        for attribute in cls.__dict__:
+            try:
+                attr_field = getattr(self, attribute)
+                if attr_field is not None:
```
liadomide, line 127: Why exclude None values? This might reveal some information also.

liadomide, line 127: How do you exclude `reference` attributes ?

framework_tvb/tvb/core/entities/transient/context_overlay.py
```diff
@@ -291,4 +291,4 @@ def fill_from_datatype(self, datatype_result, parent_burst):
 
         ### Populate Scientific attributes
         if hasattr(datatype_result, 'summary_info') and datatype_result.summary_info is not None:
-            self.add_scientific_fields(datatype_result.summary_info)
+            self.add_scientific_fields(datatype_result.summary_info())
```
liadomide, line 294: Maybe make the new method a `@property` and keep the old style of calling it ?

framework_tvb/tvb/core/entities/model/model_datatype.py
```diff
@@ -113,6 +113,23 @@ class DataType(HasTraitsIndex):
     fk_from_operation = Column(Integer, ForeignKey('OPERATIONS.id', ondelete="CASCADE"))
     parent_operation = relationship(Operation, backref=backref("DATA_TYPES", order_by=id, cascade="all,delete"))
 
+    @property
+    def summary_info(self):
+        # type: () -> typing.Dict[str, str]
+
+        ret = {}
+        if self.title:
+            ret['title'] = str(self.title)
```
popaula937, line 122: use upper-case for title, too

framework_tvb/tvb/core/entities/model/model_datatype.py
```diff
@@ -113,6 +113,23 @@ class DataType(HasTraitsIndex):
...
+
+        for attribute in self.__table__.columns.keys():
+            try:
+                attr_field = getattr(self, attribute)
+                attr_name = attribute.title().replace("_", " ")
+                ret[attr_name] = str(attr_field)
+            except Exception:
+                ""
```
popaula937, line 130: use pass instead of ""?

framework_tvb/tvb/core/entities/model/model_datatype.py
```diff
@@ -113,6 +113,40 @@ class DataType(HasTraitsIndex):
...
+                attr_field = getattr(self, attribute)
+                attr_name = self.get_attribute_name(attribute)
+                if attr_name is not None:
+                    ret[attr_name] = str(attr_field)
+            except Exception:
+                pass
+        return ret
+
+    @property
+    def get_base_table_columns(self):
```
liadomide, line 136: if you use name `get_...` then  leave it a normal method and call with `self. get_base_table_columns()`
if you make it a `@property`, then a better name would have been `def base_table_columns(self):`
Particularly in this case I would use neither, but a private method (so not a property) `def _get_base_table_columns(self):`

framework_tvb/tvb/core/entities/model/model_datatype.py
```diff
@@ -113,6 +113,40 @@ class DataType(HasTraitsIndex):
...
+        columns = self.__table__.columns.keys()
+        if type(self).__bases__[0] is DataType:
+            return columns
+        base_table_columns = type(self).__bases__[0].__table__.columns.keys()
```
liadomide, line 140: This will go just one level up in inheritance, and not consider multiple inheritance.
Will this be enough for all our use-cases?

framework_tvb/tvb/core/entities/model/model_datatype.py
```diff
@@ -113,6 +113,40 @@ class DataType(HasTraitsIndex):
...
+        columns.extend(base_table_columns)
+        return columns
+
+    def get_attribute_name(self, attr_name):
```
liadomide, line 144: In general a new method is not a bad idea, but in this case, I suggest to inline the code and  don't put this part in another method, because the `return None` convention is odd and a bit of an antipattern

PR #141, opened 2020-05-08
TVB-2651: Solved Brain Activity Visualizer Problem

VinczeRobert:
The Visualizer works now, the brain appears, but I just want to do two things before I consider it done:
1. It seems as those little visualizers in the right part don't work like in the demo. Sometimes they appear as I enter, sometimes they don't. I have to see what's there.
2. I was trying to make a general url for adapters(those which call invoke_adapter) cause there is already one for those with h5 and as you can see the new url that I added in JS and I would like to do that in case we want to move more methds there.

### DIFF
--- a/framework_tvb/tvb/adapters/visualizers/brain.py
+++ b/framework_tvb/tvb/adapters/visualizers/brain.py
@@ -138,14 +138,16 @@ class BrainViewer(ABCSurfaceDisplayer):
             surface_gid = self.surface_h5.gid.load().hex
             boundary_url = SurfaceURLGenerator.get_url_for_region_boundaries(surface_gid, self.region_map_gid,
                                                                              self.stored_adapter.id)
+            base_adapter_url = SurfaceURLGenerator.get_url_for_data_page_split(surface_gid, self.stored_adapter.id)
         else:
             boundary_url = ''
+            base_adapter_url = ''
 
         params.update(urlVertices=json.dumps(url_vertices), urlTriangles=json.dumps(url_triangles),
                       urlLines=json.dumps(url_lines), urlNormals=json.dumps(url_normals),
                       urlRegionMap=json.dumps(url_region_map), urlRegionBoundaries=boundary_url,
-                      base_activity_url=base_activity_url,
-                      isOneToOneMapping=self.one_to_one_map, minActivity=min_val, maxActivity=max_val)
+                      base_activity_url=base_activity_url, isOneToOneMapping=self.one_to_one_map,
+                      minActivity=min_val, maxActivity=max_val, base_adapter_url=base_adapter_url)
 
         normalization_factor = figure_size[0] / 800
         if figure_size[1] / 600 < normalization_factor:
@@ -230,7 +232,7 @@ class BrainViewer(ABCSurfaceDisplayer):
         :rtype: `dict`
         :raises Exception: when
                     * number of measure points exceeds the maximum allowed
-                    * a Face object cannot be found in database
+                    * a Face object cannot be found in databasef
 
         """
         self.populate_surface_fields(time_series)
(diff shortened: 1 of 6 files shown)


framework_tvb/tvb/adapters/visualizers/brain.py
```diff
@@ -230,7 +232,7 @@ def compute_parameters(self, time_series, shell_surface=None):
         :rtype: `dict`
         :raises Exception: when
                     * number of measure points exceeds the maximum allowed
-                    * a Face object cannot be found in database
+                    * a Face object cannot be found in databasef
```
liadomide, line 235: typo

framework_tvb/tvb/adapters/visualizers/surface_view.py
```diff
@@ -129,6 +129,10 @@ def get_url_for_region_boundaries(surface_gid, region_mapping_gid, adapter_id):
         return URLGenerator.build_url(adapter_id, 'generate_region_boundaries', surface_gid,
                                       parameter='region_mapping_gid=' + region_mapping_gid)
 
+    @staticmethod
+    def get_url_for_data_page_split(surface_gid, adapter_id):
```
liadomide, line 133: I suggest drop this new method, and use directly `URLGenerator .build_url` in `brain.py`

framework_tvb/tvb/adapters/visualizers/brain.py
```diff
@@ -138,14 +138,16 @@ def generate_preview(self, view_model, figure_size=None):
             surface_gid = self.surface_h5.gid.load().hex
             boundary_url = SurfaceURLGenerator.get_url_for_region_boundaries(surface_gid, self.region_map_gid,
                                                                              self.stored_adapter.id)
+            base_adapter_url = SurfaceURLGenerator.get_url_for_data_page_split(surface_gid, self.stored_adapter.id)
```
liadomide, line 141: Can this be correct? 
I see method `def read_data_page_split(self, time_series_gid...` thus expecting a TS GID and not a Surface.gid.

VinczeRobert:
I corrected the few errors that I had and I reviewed the code, it looks good to me to have a general url to adapters in case we want to move more in the future like I did here. Also, I renamed that base url so we know which baseUrl is it refering too.

About those little visualizers in the right part. Sometimes they are displayed, sometimes not. I tried to find a logic as to what makes them appear or not, but I couldn't find an answer. I even recorded my screen and went in slow motion through the frames to see if it related to the order that the elements appear, but it looks like not... Also, if they don't appear and I open Inspect Element and click on the place where they should appear, then they appear every time...

liadomide:
For the section displays on the right side:
- have you tested with version 1.5? I remember it was working there, but wouldn't hurt to check
- click on those sections should move the "navigator" to a certain position and thus redraw those parts
- Do you have any JS error when those sections are not displayed ?
We could make a short call in the next days to clarify this, and see if it is a bug in the current version to be reported, or not. Anyway. could be a separate issue, and this PR can be merged, from my pov.

VinczeRobert:
> For the section displays on the right side:
> 
> * have you tested with version 1.5? I remember it was working there, but wouldn't hurt to check
> * click on those sections should move the "navigator" to a certain position and thus redraw those parts
> * Do you have any JS error when those sections are not displayed ?
>   We could make a short call in the next days to clarify this, and see if it is a bug in the current version to be reported, or not. Anyway. could be a separate issue, and this PR can be merged, from my pov.

It can be merged from my part as well. I will answer those questions tomorrow.

framework_tvb/tvb/adapters/visualizers/brain.py
liadomide: You solved this, but it is still a problem. This method does not exist on `SurfaceURLGenerator`.

There is one unit-test failing here.  Pls look on all unit-tests when you say a task is complete!

PR #142, opened 2020-05-08
TVB-2696 Overlay on RM and Surface. Wrong Filters

liadomide:


popaula937:
So, this filter was defined in the get_filters() method and got to the DB as a condition for the SurfaceViewers to be displayed. When you click on RegionMapping/Surface, the overlay won't be shown because this filter would be applied to them and fail because none of them has the ndim attribute.

PR #144, opened 2020-05-11
TVB-2651 Review BrainActivityViewer

liadomide:
We seem to have a broken functionality in TSRegion BrainActivity viewer (after recent fix for TSSurface same viewer).
Also some slow or complex code can be observed in these parts.

### DIFF
--- a/framework_tvb/tvb/adapters/datatypes/h5/time_series_h5.py
+++ b/framework_tvb/tvb/adapters/datatypes/h5/time_series_h5.py
@@ -191,6 +191,7 @@ class TimeSeriesH5(H5File):
         metadata = self.data.get_cached_metadata()
         return metadata.min, metadata.max
 
+    #TODO Fix for TSRegion
     def read_data_page_split(self, from_idx, to_idx, step=None, specific_slices=None):
         """
         No Split needed in case of basic TS (sensors and region level)
--- a/framework_tvb/tvb/adapters/datatypes/h5/time_series_h5.py
+++ b/framework_tvb/tvb/adapters/datatypes/h5/time_series_h5.py
@@ -191,13 +191,6 @@ class TimeSeriesH5(H5File):
         metadata = self.data.get_cached_metadata()
         return metadata.min, metadata.max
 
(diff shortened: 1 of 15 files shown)


VinczeRobert:
Firstly, I apologize for the problems that I created.

Secondly, the only thing that I changed was adding that question mark instead of the slash, because the parameters were not sent correctly and I was getting this error: Traceback (most recent call last):
  File "C:\Users\Robert.Vincze\Desktop\TVB_ORIG\tvb-root\framework_tvb\tvb\interfaces\web\controllers\decorators.py", line 147, in deco
    return func(*a, **b)
  File "C:\Users\Robert.Vincze\Desktop\TVB_ORIG\tvb-root\framework_tvb\tvb\interfaces\web\controllers\decorators.py", line 96, in deco
    result = func(*a, **b)
  File "C:\Users\Robert.Vincze\Desktop\TVB_ORIG\tvb-root\framework_tvb\tvb\interfaces\web\controllers\decorators.py", line 193, in deco
    return func(*a, **b)
  File "C:\Users\Robert.Vincze\Desktop\TVB_ORIG\tvb-root\framework_tvb\tvb\interfaces\web\controllers\flow_controller.py", line 610, in read_from_h5_file
    result = result()
TypeError: read_channels_page() missing 2 required positional arguments: 'from_idx' and 'to_idx'

I checked the tests and I also tested the visualizers manually, now I believe everything is fine.

liadomide:
No need to apologize.
I should not have merged so soon.

I still think the current form in this branch has problems with some viewers (not actually tested, but I can  anticipate from the code). Maybe dual viewer, or TSEEG...

liadomide:
It looks ok to me, too.

The only thing I found broken is the usage of `ts_h5.get_grouped_space_labels` and especially `ts_h5.get_space_labels`. That part is deprecated, and should be replaced with `ABCSpaceDisplayer._get_grouped_space_labels` and `ABCSpaceDisplayer.get_space_labels`

liadomide:
Unit and Manual Tests are ok in my local env. I will merge this.

PR #145, opened 2020-05-12
TVB-2688: Put back Remover classes for datatypes

adrian-dordea:
On SqLite, the FK constraints still do not work on remove.
We will need, for example, in case of a Connectivity referred by a RM, the remove on the Connectivity to not happen, unless the user removes first the linked RM.


framework_tvb/tvb/config/init/introspector_registry.py
```diff
@@ -120,6 +126,16 @@ class IntrospectionRegistry(object):
         CreateAlgorithmCategoryConfig: import_adapters(tvb.adapters.creators, ALL_CREATORS),
     }
 
+    DATATYPE_REMOVERS = {
```
liadomide, line 129: How about moving this constant next to `from tvb.adapters.datatypes.db import ALL_DATATYPES` ?

framework_tvb/tvb/datatype_removers/remover_connectivity.py
```diff
@@ -0,0 +1,37 @@
+from tvb.adapters.datatypes.db.graph import ConnectivityMeasureIndex
+from tvb.adapters.datatypes.db.patterns import StimuliSurfaceIndex
```
liadomide, line 2: I suggest moving this folder under `tvb.adapters.datatypes.db.removers`

framework_tvb/tvb/datatype_removers/remover_connectivity.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+
+    def remove_datatype(self, skip_validation=False):
+        """
+        Called when a Connectivity is to be removed.
+        """
+        if not skip_validation:
+            associated_ts = dao.get_generic_entity(TimeSeriesRegionIndex, self.handled_datatype.gid, 'fk_connectivity_gid')
+            associated_rm = dao.get_generic_entity(RegionMappingIndex, self.handled_datatype.gid, 'fk_connectivity_gid')
+            associated_stim = dao.get_generic_entity(StimuliSurfaceIndex, self.handled_datatype.gid, 'fk_connectivity_gid')
```
liadomide, line 22: Shouldn't here be `StimuliRegionIndex` instead ?

liadomide:
Please do a detailed text search in the DataTypes for all FKs, and make sure they are handled by the removers.

The current form looks decent, except for the modules position, as I commented above.

liadomide:
Besides my suggestions above regarding comments/todos, I think a rebase with master would be necessary, to check the unit-tests

PR #146, opened 2020-05-12
TVB-2698: Fixed launching simulations with stimuli

VinczeRobert:
I fixed simulation launching with both kinds of stimuli. I went with the assumption that if you have a surface chosen, then only surface stimuli can be chosen, otherwise only region stimuli, based on the code in SimulatorStimulusFragment.

I suggest to merge TVB-2651 first, so I can make a rebase and test each visualizer with these type of surfaces to make sure they are okay.

### DIFF
--- a/framework_tvb/tvb/adapters/datatypes/db/patterns.py
+++ b/framework_tvb/tvb/adapters/datatypes/db/patterns.py
@@ -30,19 +30,31 @@
 import json
 from sqlalchemy import Column, Integer, ForeignKey, String
 from sqlalchemy.orm import relationship
-from tvb.datatypes.patterns import StimuliRegion, StimuliSurface
+from tvb.datatypes.patterns import StimuliRegion, StimuliSurface, SpatioTemporalPattern
 from tvb.adapters.datatypes.db.connectivity import ConnectivityIndex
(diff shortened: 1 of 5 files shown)


liadomide:
I think in the previous version StimuliRegion can be selected both for a region and a surface simulation.
StimuliSurface only in case a surface is selected

VinczeRobert:
> I think in the previous version StimuliRegion can be selected both for a region and a surface simulation.
> StimuliSurface only in case a surface is selected

In that case I think we need a new superclass for the two StimuliIndexes, because currently we have a DataTypeSelectField which can get StimuliRegionIndex or StimuliSurfaceIndex and we would need a way to get both when we have a surface.

liadomide:
I just checked. It does work as I said in TVB 1.5 (both sensor types in case a surface is selected).
We used to have a parent class indeed in that version, for usage on the Simulator field.
`stimulus = patterns.SpatioTemporalPattern( ...)`
The equivalent would be `SpatioTemporalPatternIndex` now.

VinczeRobert:
I made launching with stimuli work as it did in 1.5.6. I have also found out that surface wasn't loading so I fixed that as well.

PR #148, opened 2020-05-13
TVB-2692: Fixed problem with Kuramoto model

VinczeRobert:
Luckily, the issue was just the fact that a line `"of=str"` was missing from this model, sadly it requires a change in the library and I know we don't really like that.

As a so called bonus, I have noticed the Janset Rit and Zerlaut were having the same problem so I fixed those too and the Janset Rit model were also not having its variables of interest as a MultiSelectField so I changed that as well.

### DIFF
--- a/scientific_library/tvb/simulator/models/jansen_rit.py
+++ b/scientific_library/tvb/simulator/models/jansen_rit.py
@@ -459,6 +459,7 @@ class ZetterbergJansen(Model):
         it is also provides the default range of phase-plane plots.""")
 
     variables_of_interest = List(
+        of=str,
         label="Variables watched by Monitors",
         choices=("v1", "y1", "v2", "y2", "v3", "y3", "v4", "y4", "v5", "y5", "v6", "v7"),
         default=("v6", "v7", "v2", "v3", "v4", "v5"),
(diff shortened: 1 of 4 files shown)


scientific_library/tvb/simulator/models/jansen_rit.py
```diff
@@ -459,6 +459,7 @@ class ZetterbergJansen(Model):
         it is also provides the default range of phase-plane plots.""")
 
     variables_of_interest = List(
+        of=str,
```
liadomide, line 462: These are changes ok to be made even if they are in the library. 
They are left-ver bugs from our neotraits migration.

VinczeRobert:
> I am ok with the changes in the models, as long as all the previous running unit-tests will pass.
> 
> But I right to say that changes unrelated to this bug are also included in this branch?
> Could pls cleanup the history ?

Done.

liadomide:
EpileptorCodim3SlowModModelForm
`self.variables_of_interest = ArrayField(ModelsEnum.EPILEPTOR_CODIM_3_SLOW`
Isn't this also a problem ?

VinczeRobert:
It is. I'll fix it immediately.

VinczeRobert:
I am done. The fields are good now for each model.

PR #150, opened 2020-05-14
TVB-2695: Wavelet GUI input configuration

adrian-dordea:
- make range a dict and saving it as a json in the h5 so it can be saved

### DIFF
--- a/framework_tvb/tvb/adapters/analyzers/wavelet_adapter.py
+++ b/framework_tvb/tvb/adapters/analyzers/wavelet_adapter.py
@@ -66,12 +66,12 @@ class WaveletAdapterModel(ViewModel, ContinuousWaveletTransform):
 class RangeForm(Form):
     def __init__(self, prefix=''):
         super(RangeForm, self).__init__(prefix)
-        self.lo = SimpleFloatField(self, name='lo', required=True, label='Lo', doc='start of range')
-        # default=ContinuousWaveletTransform.frequencies.lo)
-        self.step = SimpleFloatField(self, name='step', required=True, label='Step', doc='step of range')
-        # default=ContinuousWaveletTransform.frequencies.step)
-        self.hi = SimpleFloatField(self, name='hi', required=True, label='Hi', doc='end of range')
-        # default=ContinuousWaveletTransform.frequencies.hi)
+        self.lo = SimpleFloatField(self, name='lo', required=True, label='Lo', doc='start of range',
+        default=ContinuousWaveletTransform.frequencies.default.lo)
+        self.step = SimpleFloatField(self, name='step', required=True, label='Step', doc='step of range',
+        default=ContinuousWaveletTransform.frequencies.default.step)
+        self.hi = SimpleFloatField(self, name='hi', required=True, label='Hi', doc='end of range',
+        default=ContinuousWaveletTransform.frequencies.default.hi)
 
 
 class ContinuousWaveletTransformAdapterForm(ABCAdapterForm):
@@ -147,12 +147,6 @@ class ContinuousWaveletTransformAdapter(ABCAsynchronous):
 
         if view_model.sample_period is not None:
             algorithm.sample_period = view_model.sample_period
-
-        # TODO range form is not correctly populated, some work is still needed there
-        # if (view_model.frequencies is not None):
-        #     and 'lo' in frequencies_parameters
-        #         and 'hi' in frequencies_parameters and frequencies_parameters['hi'] != frequencies_parameters['lo']):
-        #     algorithm.frequencies = Range(**frequencies_parameters)
 
         if view_model.normalisation is not None:
             algorithm.normalisation = view_model.normalisation
--- a/framework_tvb/tvb/adapters/analyzers/wavelet_adapter.py
+++ b/framework_tvb/tvb/adapters/analyzers/wavelet_adapter.py
@@ -40,7 +40,7 @@ ContinuousWaveletTransform Analyzer.
 import uuid
 import numpy
 from tvb.adapters.datatypes.h5.time_series_h5 import TimeSeriesH5
-from tvb.basic.neotraits._attr import Float
+from tvb.basic.neotraits.api import Float
 from tvb.core.neotraits.view_model import ViewModel, DataTypeGidAttr
 from tvb.analyzers.wavelet import ContinuousWaveletTransform
 from tvb.datatypes.time_series import TimeSeries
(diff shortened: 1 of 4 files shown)


liadomide:
I expect `TestAdapters.test_wavelet_adapter` to pass before we can merge

framework_tvb/tvb/adapters/analyzers/wavelet_adapter.py
```diff
@@ -148,12 +148,6 @@ def configure(self, view_model):
         if view_model.sample_period is not None:
             algorithm.sample_period = view_model.sample_period
 
-        # TODO range form is not correctly populated, some work is still needed there
-        # if (view_model.frequencies is not None):
-        #     and 'lo' in frequencies_parameters
-        #         and 'hi' in frequencies_parameters and frequencies_parameters['hi'] != frequencies_parameters['lo']):
-        #     algorithm.frequencies = Range(**frequencies_parameters)
-
```
liadomide, line 156: Are you certain this part simply needs to be removed?
Are the values from UI being passed to the ViewModel and then into the algorithm at launch time?

adrian-dordea: yes, at launch frequencies is in both ViewModel and algorithm at launch that is why I simply removed it.

framework_tvb/tvb/adapters/analyzers/wavelet_adapter.py
```diff
@@ -66,12 +66,12 @@ class WaveletAdapterModel(ViewModel, ContinuousWaveletTransform):
 class RangeForm(Form):
     def __init__(self, prefix=''):
         super(RangeForm, self).__init__(prefix)
-        self.lo = SimpleFloatField(self, name='lo', required=True, label='Lo', doc='start of range')
-        # default=ContinuousWaveletTransform.frequencies.lo)
-        self.step = SimpleFloatField(self, name='step', required=True, label='Step', doc='step of range')
-        # default=ContinuousWaveletTransform.frequencies.step)
-        self.hi = SimpleFloatField(self, name='hi', required=True, label='Hi', doc='end of range')
-        # default=ContinuousWaveletTransform.frequencies.hi)
+        self.lo = SimpleFloatField(self, name='lo', required=True, label='Lo', doc='start of range',
```
popaula937, line 69: We should replace the usage of SimpleFloatField with FloatField and adapt the definition.

framework_tvb/tvb/adapters/analyzers/wavelet_adapter.py
popaula937: Please do a debug and choose other values than the defaults. I think it always uses the defaults if we do not explicitly do something. I think you should extend the fill_trait method on ContinuousWaveletTransformAdapterForm and explicitly prepare the Range object with values from RangeForm.

framework_tvb/tvb/adapters/analyzers/wavelet_adapter.py
```diff
@@ -62,16 +62,37 @@ class WaveletAdapterModel(ViewModel, ContinuousWaveletTransform):
...
+    )
+
+    f_step = Attr(
+        field_type=float,
+        required=True,
+        label='Step',
+        default=ContinuousWaveletTransform.frequencies.default.step,
+        doc='step of range'
+    )
+
```
popaula937, line 88: I think these new fields are not needed on the model class, we can define them directly in the Form where they are used. Otherwise, they will be written separately in the H5 file, while we already write the Range.

popaula937: Also, you can use Float instead of Attr so you don't need to define Attr(field_type=float).
required is True by default so that is not necessary either,

framework_tvb/tvb/adapters/analyzers/wavelet_adapter.py
```diff
@@ -62,16 +62,37 @@ class WaveletAdapterModel(ViewModel, ContinuousWaveletTransform):
...
         super(RangeForm, self).__init__(prefix)
-        self.lo = SimpleFloatField(self, name='lo', required=True, label='Lo', doc='start of range')
-        # default=ContinuousWaveletTransform.frequencies.lo)
-        self.step = SimpleFloatField(self, name='step', required=True, label='Step', doc='step of range')
-        # default=ContinuousWaveletTransform.frequencies.step)
-        self.hi = SimpleFloatField(self, name='hi', required=True, label='Hi', doc='end of range')
-        # default=ContinuousWaveletTransform.frequencies.hi)
+        self.lo = FloatField(WaveletAdapterModel.f_lo, self, name='Lo')
+        self.hi = FloatField(WaveletAdapterModel.f_hi, self, name='Hi')
+        self.step = FloatField(WaveletAdapterModel.f_step, self, name='Step')
```
popaula937, line 95: So, related to the comment above, put directly here:
self.lo = FloatField(Float(label='Lo', default=ContinuousWaveletTransform.frequencies.default.lo, doc='start of range'), self, name='Lo')

framework_tvb/tvb/adapters/analyzers/wavelet_adapter.py
```diff
@@ -97,6 +118,12 @@ def get_view_model():
     def get_required_datatype():
         return TimeSeriesIndex
 
+    def fill_trait(self, datatype):
+        super(ContinuousWaveletTransformAdapterForm, self).fill_trait(datatype)
+        datatype.frequencies.lo = self.frequencies.form.lo.data
+        datatype.frequencies.step = self.frequencies.form.step.data
+        datatype.frequencies.hi = self.frequencies.form.hi.data
```
popaula937, line 125: Usually, .value should return the proper validated value of the field

framework_tvb/tvb/core/neotraits/_h5accessors.py
```diff
@@ -439,6 +439,28 @@ def load(self):
...
 
+class JsonRange(Scalar):
+    """
+    Stores and loads a Range in the form of a json in h5.
+    """
+
+    def __init__(self, trait_attribute, h5file, name=None, json_encoder=None, json_decoder=None):
+        super(JsonRange, self).__init__(trait_attribute, h5file, name)
+        self.json_encoder = json_encoder
+        self.json_decoder = json_decoder
```
popaula937, line 450: I don't think encoder and decoder are necessary here since we serialize a dictionary.

framework_tvb/tvb/adapters/analyzers/wavelet_adapter.py
```diff
@@ -40,16 +40,16 @@
 import uuid
 import numpy
 from tvb.adapters.datatypes.h5.time_series_h5 import TimeSeriesH5
+from tvb.basic.neotraits._attr import Float
```
popaula937, line 43: from tvb.basic.neotraits.api import Float

popaula937:
Don't forget to fix test TestAdapters.test_wavelet_adapter before we can merge

liadomide:
a rebase with master seems necessary here

PR #151, opened 2020-05-15
TVB-2700: Fixed noise configuration in case of Stochastic Integrators + Tests

VinczeRobert:
Did some additional cleanup on unit-tests around this area

### DIFF
--- a/framework_tvb/tvb/tests/framework/interfaces/web/controllers/noise_configuration_controller_test.py
+++ b/framework_tvb/tvb/tests/framework/interfaces/web/controllers/noise_configuration_controller_test.py
@@ -34,45 +34,36 @@
 
 import json
 import cherrypy
-import pytest
 from tvb.interfaces.web.controllers.simulator_controller import SimulatorController
 from tvb.tests.framework.interfaces.web.controllers.base_controller_test import BaseTransactionalControllerTest
-from tvb.tests.framework.adapters.simulator.simulator_adapter_test import SIMULATOR_PARAMETERS
-from tvb.core.entities.model.model_burst import PARAM_INTEGRATOR, PARAM_MODEL
 from tvb.interfaces.web.controllers import common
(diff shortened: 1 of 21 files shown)


VinczeRobert:
> These look good to me, I did some tests and everything seems to work correctly.
> Please fix related test and we can merge.

Ok, but this test is failing in master as well...

popaula937:
> > These look good to me, I did some tests and everything seems to work correctly.
> > Please fix related test and we can merge.
> 
> Ok, but this test is failing in master as well...

Yes, but we have to fix all tests eventually, and would probably be easier if we do it incrementally,  by fixing the ones that are related to our tasks.

VinczeRobert:
I fixed the test and I know that something's not good as there are a lot more failures on Jenkins, I will try to see why...

VinczeRobert:
Please don't merge this yet, because I see there are more tests failed on Jenkins, I think I have added clean_database() where I should have not, I will check again...

framework_tvb/tvb/tests/framework/interfaces/web/controllers/noise_configuration_controller_test.py
```diff
@@ -34,67 +34,35 @@
...
+        self.noise_controller = NoiseConfigurationController()
+        SimulatorController().index()
+        self.simulator = cherrypy.session[common.KEY_SIMULATOR_CONFIG]
+        self.connectivity = connectivity_factory()
+        self.simulator.connectivity = self.connectivity.gid
+        self.simulator.integrator = HeunStochastic()
 
     def transactional_teardown_method(self):
         """ Cleans the testing environment """
+        self.clean_database()
```
liadomide, line 62: Why this cleanup? considering that the test in transactional ?

VinczeRobert:
I deleted transactional_setup_fixtures and I might have found the problem, or at least one of them, but I am not quite sure, I have to investigate some more and I am curious how many tests fail now on Jenkins.

PR #152, opened 2020-05-15
TVB-2701: Fix Spatial Average Monitor Launching 

VinczeRobert:
So far I have only managed to fix launching with the default mask (Hemispheres), but there are other problems as well:

1. Launching with Cortical default mask gives this error: Areas in the spatial_mask must be specified as a contiguous set of indices starting from zero.
2. Brain Activity 2D and 3D Visualizer as well as Brain Activity Visualizer doesn't work with this monitor, because `NO_OF_MEASURE_POINTS !== activitiesData[0].length ( virtualBrain.js, line 404).`
3. This one is basically the same problem as described in https://req.thevirtualbrain.org/browse/TVB-2707 and that is the fact that the surface attributes only exists in TimeSeriesSurface, not on the TimeSeries superclass, so launching with surface and `SpatialAverage` or with surface and a `Projection Monitor` will try to set the surface from TimeSeriesRegion (in the case of SpatialAverage) or from TimeSeriesEEG (in the case of EEG) but won't have it. Should we add the surface on the superclass?

### DIFF
--- a/framework_tvb/tvb/adapters/datatypes/db/connectivity.py
+++ b/framework_tvb/tvb/adapters/datatypes/db/connectivity.py
@@ -49,6 +49,7 @@ class ConnectivityIndex(DataType):
     tract_lengths_mean = Column(Float)
 
     cortical_mask = Column(Boolean)
+    hemispheres_mask = Column(Boolean)
 
     # TODO: keep these metadata?
     # weights_non_zero_id = Column(Integer, ForeignKey('narrays.id'), nullable=False)
@@ -67,6 +68,10 @@ class ConnectivityIndex(DataType):
         self.cortical_mask = True
         if datatype.cortical is None or len(set(datatype.cortical)) != 2:
             self.cortical_mask = False
+
+        self.hemispheres_mask = True
+        if datatype.hemispheres is None or len(set(datatype.hemispheres)) != 2:
+            self.hemispheres_mask = False
 
         self.number_of_regions = datatype.number_of_regions
         self.number_of_connections = datatype.number_of_connections
--- a/framework_tvb/tvb/adapters/datatypes/db/connectivity.py
+++ b/framework_tvb/tvb/adapters/datatypes/db/connectivity.py
@@ -48,8 +48,8 @@ class ConnectivityIndex(DataType):
     tract_lengths_max = Column(Float)
     tract_lengths_mean = Column(Float)
 
-    cortical_mask = Column(Boolean)
-    hemispheres_mask = Column(Boolean)
+    has_cortical_mask = Column(Boolean)
+    has_hemispheres_mask = Column(Boolean)
 
     # TODO: keep these metadata?
     # weights_non_zero_id = Column(Integer, ForeignKey('narrays.id'), nullable=False)
@@ -65,13 +65,13 @@ class ConnectivityIndex(DataType):
         # type: (Connectivity)  -> None
         super(ConnectivityIndex, self).fill_from_has_traits(datatype)
 
-        self.cortical_mask = True
+        self.has_cortical_mask = True
         if datatype.cortical is None or len(set(datatype.cortical)) != 2:
-            self.cortical_mask = False
+            self.has_cortical_mask = False
 
-        self.hemispheres_mask = True
+        self.has_hemispheres_mask = True
         if datatype.hemispheres is None or len(set(datatype.hemispheres)) != 2:
-            self.hemispheres_mask = False
+            self.has_hemispheres_mask = False
 
         self.number_of_regions = datatype.number_of_regions
         self.number_of_connections = datatype.number_of_connections
--- a/framework_tvb/tvb/adapters/datatypes/db/connectivity.py
+++ b/framework_tvb/tvb/adapters/datatypes/db/connectivity.py
@@ -65,14 +65,8 @@ class ConnectivityIndex(DataType):
         # type: (Connectivity)  -> None
         super(ConnectivityIndex, self).fill_from_has_traits(datatype)
 
-        self.has_cortical_mask = True
-        if datatype.cortical is None or len(set(datatype.cortical)) != 2:
-            self.has_cortical_mask = False
-
-        self.has_hemispheres_mask = True
-        if datatype.hemispheres is None or len(set(datatype.hemispheres)) != 2:
-            self.has_hemispheres_mask = False
-
+        self.has_cortical_mask = datatype.cortical is not None
+        self.has_hemispheres_mask = datatype.hemispheres is not None
         self.number_of_regions = datatype.number_of_regions
         self.number_of_connections = datatype.number_of_connections
         self.undirected = datatype.undirected
--- a/framework_tvb/tvb/adapters/datatypes/db/connectivity.py
+++ b/framework_tvb/tvb/adapters/datatypes/db/connectivity.py
@@ -48,6 +48,8 @@ class ConnectivityIndex(DataType):
     tract_lengths_max = Column(Float)
     tract_lengths_mean = Column(Float)
 
+    cortical_mask = Column(Boolean)
+
     # TODO: keep these metadata?
     # weights_non_zero_id = Column(Integer, ForeignKey('narrays.id'), nullable=False)
     # weights_non_zero = relationship(NArrayIndex, foreign_keys=weights_non_zero_id, primaryjoin=NArrayIndex.id == weights_non_zero_id)
@@ -61,11 +63,17 @@ class ConnectivityIndex(DataType):
     def fill_from_has_traits(self, datatype):
         # type: (Connectivity)  -> None
         super(ConnectivityIndex, self).fill_from_has_traits(datatype)
+
+        self.cortical_mask = True
+        if datatype.cortical is None or len(set(datatype.cortical)) != 2:
+            self.cortical_mask = False
+
         self.number_of_regions = datatype.number_of_regions
         self.number_of_connections = datatype.number_of_connections
         self.undirected = datatype.undirected
         self.weights_min, self.weights_max, self.weights_mean = from_ndarray(datatype.weights)
         self.tract_lengths_min, self.tract_lengths_max, self.tract_lengths_mean = from_ndarray(datatype.tract_lengths)
+
         # self.weights_non_zero = NArrayIndex.from_ndarray(datatype.weights[datatype.weights.nonzero()])
         # self.tract_lengths_non_zero = NArrayIndex.from_ndarray(datatype.tract_lengths[datatype.tract_lengths.nonzero()])
         # self.tract_lengths_connections = NArrayIndex.from_ndarray(datatype.tract_lengths[datatype.weights.nonzero()])
--- a/framework_tvb/tvb/adapters/simulator/simulator_fragments.py
+++ b/framework_tvb/tvb/adapters/simulator/simulator_fragments.py
@@ -69,17 +69,31 @@ class SimulatorSurfaceFragment(ABCAdapterForm):
 
 
 class SimulatorRMFragment(ABCAdapterForm):
-    def __init__(self, prefix='', project_id=None, surface_index=None):
+    def __init__(self, prefix='', project_id=None, surface_index=None, connectivity_gid=None):
         super(SimulatorRMFragment, self).__init__(prefix, project_id)
         conditions = None
         if surface_index:
-            conditions = FilterChain(fields=[FilterChain.datatype + '.fk_surface_gid'], operations=["=="],
-                                     values=[str(surface_index.gid)])
+            conditions = FilterChain(fields=[FilterChain.datatype + '.fk_surface_gid',
+                                             FilterChain.datatype + '.fk_connectivity_gid'],
+                                     operations=["==", "=="],
+                                     values=[str(surface_index.gid), connectivity_gid.hex])
         self.rm = TraitDataTypeSelectField(CortexViewModel.region_mapping_data, self, name='region_mapping',
                                            conditions=conditions)
         self.lc = TraitDataTypeSelectField(CortexViewModel.local_connectivity, self, name='local_connectivity',
                                            conditions=conditions)
         self.coupling_strength = ArrayField(CortexViewModel.coupling_strength, self)
+
+    def fill_from_trait(self, trait):
+        # type: (Simulator) -> None
+        self.coupling_strength.data = trait.surface.coupling_strength
+        if hasattr(trait.surface, 'region_mapping_data'):
+            self.rm.data = trait.surface.region_mapping_data.hex
+        else:
+            self.rm.data = None
+        if trait.surface.local_connectivity:
+            self.lc.data = trait.surface.local_connectivity.hex
+        else:
+            self.lc.data = None
 
 
 class SimulatorStimulusFragment(ABCAdapterForm):
--- a/framework_tvb/tvb/adapters/simulator/simulator_fragments.py
+++ b/framework_tvb/tvb/adapters/simulator/simulator_fragments.py
@@ -87,18 +87,6 @@ class SimulatorRMFragment(ABCAdapterForm):
                                            conditions=lc_conditions)
         self.coupling_strength = ArrayField(CortexViewModel.coupling_strength, self)
 
-    def fill_from_trait(self, trait):
-        # type: (Simulator) -> None
-        self.coupling_strength.data = trait.coupling_strength
-        if hasattr(trait, 'region_mapping_data'):
-            self.rm.data = trait.region_mapping_data.hex
-        else:
-            self.rm.data = None
-        if trait.local_connectivity:
-            self.lc.data = trait.local_connectivity.hex
-        else:
-            self.lc.data = None
-
 
 class SimulatorStimulusFragment(ABCAdapterForm):
     def __init__(self, prefix='', project_id=None, is_surface_simulation=False):
--- a/framework_tvb/tvb/adapters/simulator/simulator_fragments.py
+++ b/framework_tvb/tvb/adapters/simulator/simulator_fragments.py
@@ -71,27 +71,31 @@ class SimulatorSurfaceFragment(ABCAdapterForm):
 class SimulatorRMFragment(ABCAdapterForm):
     def __init__(self, prefix='', project_id=None, surface_index=None, connectivity_gid=None):
         super(SimulatorRMFragment, self).__init__(prefix, project_id)
-        conditions = None
+        rm_conditions = None
+        lc_conditions = None
         if surface_index:
-            conditions = FilterChain(fields=[FilterChain.datatype + '.fk_surface_gid',
+            rm_conditions = FilterChain(fields=[FilterChain.datatype + '.fk_surface_gid',
                                              FilterChain.datatype + '.fk_connectivity_gid'],
                                      operations=["==", "=="],
-                                     values=[str(surface_index.gid), connectivity_gid.hex])
+                                     values=[str(surface_index.gid), str(connectivity_gid.hex)])
+            lc_conditions = FilterChain(fields=[rm_conditions.fields[0]], operations=[rm_conditions.operations[0]],
+                                        values=[rm_conditions.values[0]])
         self.rm = TraitDataTypeSelectField(CortexViewModel.region_mapping_data, self, name='region_mapping',
-                                           conditions=conditions)
+                                           conditions=rm_conditions)
+
         self.lc = TraitDataTypeSelectField(CortexViewModel.local_connectivity, self, name='local_connectivity',
-                                           conditions=conditions)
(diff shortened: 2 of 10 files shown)


liadomide:
I suggest setting attribute `surface` only on the TS instances that need it (I think it is only `TimeSeriesSurface`), instead of setting it when we have it in the Simulator.
See my comment on TVB-2707

VinczeRobert:
I did further investigations and I realized that there is actually nothing wrong with the code itself.

Until yesterday I have always tried to run simulations with this monitor using the default connectivity from the Default Project. 

Then I tried to upload new connectivities and I realized that Brain Activity Viewers are not even shown with other connectivities, unless you specify a surface. I don't know why the two brain viewers are shown only for the default connectivity but if these viewers should only be available with surfaces, then everything is allright with the code. 

Also, connectivity 66 from tvb_data has no cortical, so launching that with cortical default mask will fail. Failure will also happen if you choose a surface and a region mapping which don't match in numbers (I think some filtering could be done to not allow that to happen).

VinczeRobert:
I disabled the default mask field when cortical is None or only has True values. This way simulation with this monitor should never fail.

liadomide:
Please rebase with master after recent changes there

VinczeRobert:
> Please rebase with master after recent changes there

I did. I also noticed that the part which handled Projection monitors in set_monitor_params was deleted in master and I believe it was by accident because launching with projection monitors was failing... so I readded that part.

framework_tvb/tvb/adapters/simulator/simulator_fragments.py
```diff
@@ -71,18 +71,32 @@ def fill_trait(self, datatype):
...
-            conditions = FilterChain(fields=[FilterChain.datatype + '.fk_surface_gid'], operations=["=="],
-                                     values=[str(surface_index.gid)])
+            conditions = FilterChain(fields=[FilterChain.datatype + '.fk_surface_gid',
+                                             FilterChain.datatype + '.fk_connectivity_gid'],
+                                     operations=["==", "=="],
+                                     values=[str(surface_index.gid), connectivity_gid.hex])
         self.rm = TraitDataTypeSelectField(CortexViewModel.region_mapping_data, self, name='region_mapping',
                                            conditions=conditions)
         self.lc = TraitDataTypeSelectField(CortexViewModel.local_connectivity, self, name='local_connectivity',
                                            conditions=conditions)
```
liadomide, line 85: I do not see field  `LocalConnectivityIndex. fk_connectivity_gid`. Will this filter ever work for LocalConnectivity?

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -697,6 +697,14 @@ def _skip_raw_monitor(self, monitors):
                                                           type(monitors[first_monitor_index]).__name__)
         return first_monitor_index, last_loaded_fragment_url
 
+    @staticmethod
+    def _check_cortical_for_spatial_average(connectivity, form, is_surface_simulation):
+        connectivity_index = ABCAdapter.load_entity_by_gid(connectivity)
+        connectivity = h5.load_from_index(connectivity_index)
```
liadomide, line 703: This load is inefficient. 
Can't we prepare a pre-computed flag on ConnectivityIndex and evaluate it when we import a Connectivity (in `ConnectivityIndex.fill_from_has_traits`? 
The situation won't ever change in TVB after import, so no need to evaluate at every sim execution time

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -749,6 +757,10 @@ def set_monitors(self, **data):
         form = get_form_for_monitor(type(monitor))(indexes, '', common.get_current_project().id)
         form.fill_from_trait(monitor)
 
+        if isinstance(monitor, SpatialAverage):
```
liadomide, line 760: this `if isinstance ...` is ugly. Pls find another way.
Maybe `get_form_for_monitor ` could receive an extra param (e.g. the `session_stored_simulator ` to retrieve the ConnIndex from it in each Monitor Form) ?

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -813,6 +825,10 @@ def set_monitor_params(self, current_monitor, **data):
             indexes = self._get_variables_of_interest_indexes(all_variables, chosen_variables)
             form = get_form_for_monitor(type(monitor))(indexes)
             form.fill_from_post(data)
+
+            if isinstance(form, SpatialAverageMonitorForm) and form.default_mask.data is None:
+                form.default_mask.data = 'hemispheres'
```
liadomide, line 830: Why this is necessary? Isn't enough to have a default in the form and this field to be requored?

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -838,6 +854,22 @@ def set_monitor_params(self, current_monitor, **data):
                 SimulatorWizzardURLs.SET_MONITOR_PARAMS_URL, current_monitor)
             return rendering_rules.to_dict()
 
+        if isinstance(monitor, Projection) and cherrypy.request.method == 'POST':
```
liadomide, line 857: @popaula937 Maybe you know what happened here at rebase. Is this still necessary?

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -838,6 +854,22 @@ def set_monitor_params(self, current_monitor, **data):
...
+            # load region mapping
+            region_mapping_index = ABCAdapter.load_entity_by_gid(data['region_mapping'])
+            region_mapping = h5.load_from_index(region_mapping_index)
+            monitor.region_mapping = region_mapping.gid
```
liadomide, line 861: This is really strange anyway .... why load the full HasTraits object in memory to only get the GID from it?
What am I missing ?

scientific_library/tvb/simulator/monitors.py
```diff
@@ -248,14 +248,14 @@ def config_for_sim(self, simulator):
                 self.spatial_mask = simulator.surface.region_mapping
             else:
                 conn = simulator.connectivity
-                if self.default_mask[0] == 'cortical':
```
liadomide, line 251: From reading the `SpatialAverage.config_for_sim` method, it seem to me that attribute `default_mask` should have 3 options: `region_mapping` (only one accepted in case of Surface simulations), 'cortical' (only for region sim, when the connectivity has this field properly defined), and 'hemispheres' (only for region sim, and again only when this field is properly defined).
Could you try a small refactor in this direction, pls ?

framework_tvb/tvb/adapters/simulator/simulator_fragments.py
VinczeRobert: You are right, the connectivity filter should have been added only for the region mapping.

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
VinczeRobert: Okay, I implemented this change, but that adds a new field to the database for ConnectivityIndex. Is it okay?

liadomide: yes, it is ok to have extra extra fields in *Index entities, to accommodate our use-cases. why wouldn't be ?

VinczeRobert: Solved.

VinczeRobert:
I didn't implement all required changes yet, but I moved the variables of interests and spatial average monitor to forms. I hope it's okay to have ABCAdapter in a form, I needed it for the conn. index.

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
popaula937: This should not be necessary since at this point we already have the correct GIDs as strings in `data`. And the `fill_from_post` is now handling the conversion of GIDs from string to UUID for the `TraitDataTypeSelectField`.

popaula937: @VinczeRobert can you detail a bit on what is not working?

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
popaula937: You are right, these are not necessary since we already have the correct GIDs as strings in `data`

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
VinczeRobert: In fact it's not working when you don't choose a RM, so I discussed with Paula to make that field required and in case a surface is chosen, I will add a filter to only allow the region mapping which was selected at that step.

popaula937: Discussed with Robert and the error he was: 
`Please specify a region mapping on the EEG/MEG/iEEG monitor when performing a region simulation.
`
This happens because the RegionMapping was optional on the Projection Monitor, but it is actually required to select one for region-level simulations. Thus, if he did not select a RegionMapping before launching the simulation, he got this error.
It works now because he declared the RegionMapping as required on the Projection monitor, thus, the select field is filled in by default (not because of putting back this code).
Not sure whether we should keep it required, but if we would, we could just ensure that for Surface simulations, the same RM is chosen as for the Cortex.
@liadomide what do you think?

scientific_library/tvb/simulator/monitors.py
VinczeRobert: I tried something. Please look at my latest changes if they are okay.

framework_tvb/tvb/adapters/datatypes/db/connectivity.py
```diff
@@ -48,6 +48,9 @@ class ConnectivityIndex(DataType):
     tract_lengths_max = Column(Float)
     tract_lengths_mean = Column(Float)
 
+    cortical_mask = Column(Boolean)
+    hemispheres_mask = Column(Boolean)
```
liadomide, line 52: Maybe rename these into `has_cortical_mask` and `has_hemisphere_mask` to signal that is a boolean?

scientific_library/tvb/simulator/monitors.py
```diff
@@ -228,9 +230,10 @@ class SpatialAverage(Monitor):
 
     default_mask = Attr(
         str,
-        choices=("cortical", "hemispheres"),
-        default="hemispheres",
+        choices=(CORTICAL, HEMISPHERES),
+        default=None,
```
liadomide, line 234: Maybe leave default HEMISPHERES, for the library API consistency?

scientific_library/tvb/simulator/monitors.py
```diff
@@ -371,8 +368,7 @@ class Projection(Monitor):
 
     region_mapping = Attr(
         RegionMapping,
-        required=False,
-        label="region mapping",  #order=3,
+        label="Region Mapping",  #order=3,
```
liadomide, line 371: Comments like `#order=...` can be removed (they are deprecated)

scientific_library/tvb/simulator/monitors.py
```diff
@@ -371,8 +368,7 @@ class Projection(Monitor):
 
     region_mapping = Attr(
         RegionMapping,
-        required=False,
```
liadomide, line 374: The way currently `Projection.config_for_sim` is written, `Projection. region_mapping ` attribute can be missing in case of a surface based simulation, and then the RM is read from there.

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
liadomide: Exactly. After reading the code in the library I think we should put this RM required only in the web GUI, as otherwise the library api will change in a not so efficient manner (the user will need to duplicately put RM both on the simulator.cortex and on simulator.monitors[Projection].rm

scientific_library/tvb/simulator/monitors.py
liadomide: It looks much better now. Very small things are left to solve (see comments).

scientific_library/tvb/simulator/monitors.py
VinczeRobert: Done.

scientific_library/tvb/simulator/monitors.py
VinczeRobert: Done.

framework_tvb/tvb/adapters/datatypes/db/connectivity.py
VinczeRobert: Done.

scientific_library/tvb/simulator/monitors.py
```diff
@@ -216,21 +216,24 @@ class SpatialAverage(Monitor):
...
     _ui_name = "Spatial average with temporal sub-sample"
+    CORTICAL = "cortical"
+    HEMISPHERES = "hemispheres"
 
     spatial_mask = NArray(  #TODO: Check it's a vector of length Nodes (like region mapping for surface)
         dtype=int,
-        label="An index mask of nodes into areas",
-        required=False,
+        label="Spatial Mask",
+        required=HEMISPHERES,
```
liadomide, line 225: `required` should be a boolean. I think you meant to change the default on the next field

VinczeRobert: Yes, sorry.

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
liadomide: I saw the required flag on RM reverted in the library. Is it put TRUE in the ViewModel  or another place from the framework?

scientific_library/tvb/simulator/monitors.py
liadomide: How about the surface simulations? 
Could you put the "region_mapping" - uneditable value for that use-case?

VinczeRobert: Wouldn't it be better to add a filter there to only allow the RM chosen before?

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
VinczeRobert: It is FALSE now in ProjectionViewModel. I understood that I have to keep it False because it is not required in case of Surface Simulations. 

My past idea was to keep it required and if we have a surface simulation then there would be a filter to only allow the RM already chosen, but I think it wasn't accepted.

popaula937: It was accepted with the observation that we should not change the flag directly in tvb-library, but only in ProjectionViewModel 
> put this RM required only in the web GUI (from Lia)

And yes, show only the already chosen RM, and maybe disable the field entirely.

VinczeRobert:
I added the filter for RM. I tried to make the select field disabled with surface simulations and even though form.region_mapping.disabled=True, in UI it didn't appear as disabled so I didn't waste more time with that.

popaula937:
> I added the filter for RM. I tried to make the select field disabled with surface simulations and even though form.region_mapping.disabled=True, in UI it didn't appear as disabled so I didn't waste more time with that.

You have to add the `disabled` attr to `TraitDatatypeSelectField` and then pass it to `super()`.
But making it disabled would mean some more work since, at submit, the disabled fields are not sent to the server. Then our `fill_trait()` method would fail because RM is None on the form while it is required on the ProjectionMonitorViewModel.
Not sure it is worth to do this since you are only displaying one RM in that select field anyway.

popaula937:
Don't forget to fix the failed tests!

VinczeRobert:
I tried to simplify the handling of variables of interest too much but the failing tests showed me that it wasn't quite correct, so I had to change that. Now it's okay and the tests are passing as well.

scientific_library/tvb/simulator/monitors.py
liadomide: I am not referring to the Projection Monitor `region_mapping ` attribute here, but to the Monitor SpatialAverage.
See my first comment on this thread: if you have a surface level simulation, with `SpacialAverageMonitor`, then `default_mask` can not have values `cortical` or `hemisphere`, but instead should be a 3rd value, available only for SurfaceSimulations: `region_mapping_mask`. At least this is how I understand it from the `SpatialAverage.config_for_sim` method implementation logic. Would you have a different understanding ?

liadomide:
@maedoc @maedoc-amu Here Robert did a small fix in `SpatialAverage` Monitor. 
The previous code was no longer working and it also had a somehow confusing written code in the `SpatialAverage.config_for_sim` method.
It would be great if you could just acknowledge the new code and tell us if you spot something wrong. unit-tests in tvb-library all pass.

framework_tvb/tvb/tests/framework/core/services/hpc_scheduler_client_test.py
```diff
@@ -179,11 +180,12 @@ def test_prepare_inputs_with_eeg_monitor(self, operation_factory, simulator_fact
         dao.store_entity(prj_db_db)
 
         eeg_monitor = EEGViewModel(projection=proj.gid, sensors=sensors.gid)
+        eeg_monitor.region_mapping = region_mapping.gid
 
         sim_folder, sim_gid = simulator_factory(op=op, monitor=eeg_monitor)
         hpc_client = HPCSchedulerClient()
         input_files = hpc_client._prepare_input(op, sim_gid)
-        assert len(input_files) == 10
+        assert len(input_files) == 13
```
popaula937, line 188: I don't agree with this change. The test should be adjusted in such a manner that it still prepares 10 input files.
I agree with adding a RegionMapping since it is now required on the ProjMonitorViewModel, but it should be linked with the Connectivity and Surface we put on the simulator. 
By using RegionMappingIndex factory, without linking it to the connectivity and surface used on the simulator, we generate extra Connectivity and Surface H5 files in the input_files list.

framework_tvb/tvb/adapters/simulator/simulator_fragments.py
```diff
@@ -71,18 +71,36 @@ def fill_trait(self, datatype):
...
         self.lc = TraitDataTypeSelectField(CortexViewModel.local_connectivity, self, name='local_connectivity',
-                                           conditions=conditions)
+                                           conditions=lc_conditions)
         self.coupling_strength = ArrayField(CortexViewModel.coupling_strength, self)
 
+    def fill_from_trait(self, trait):
+        # type: (Simulator) -> None
+        self.coupling_strength.data = trait.coupling_strength
+        if hasattr(trait, 'region_mapping_data'):
+            self.rm.data = trait.region_mapping_data.hex
```
maedoc, line 96: what is the `.hex` here?

framework_tvb/tvb/adapters/datatypes/db/connectivity.py
```diff
@@ -61,11 +64,21 @@ class ConnectivityIndex(DataType):
     def fill_from_has_traits(self, datatype):
         # type: (Connectivity)  -> None
         super(ConnectivityIndex, self).fill_from_has_traits(datatype)
+
+        self.has_cortical_mask = True
+        if datatype.cortical is None or len(set(datatype.cortical)) != 2:
```
maedoc, line 69: why `len(set(datatype.cortical)) != 2`? If all elements of `.cortical` are `True`, it is still a cortical mask.  I am surprised that `.cortical is None` is insufficient.  It would be a lot cleaner to write
```python
self.has_cortical_mask = datatype.cortical is not None
```

framework_tvb/tvb/adapters/simulator/simulator_fragments.py
popaula937: `trait.region_mapping_data` here holds a GID of type UUID and by using `.hex` we take a string version for that GID to put on our forms.
But this overriding of `fill_from_trait` it's not necessary since these cases are handled by each `Field` object.

framework_tvb/tvb/adapters/datatypes/db/connectivity.py
liadomide: You are right @maedoc this was wrongly covering another problem documented by Paula today: when `Conn.cortical` was all True (so only cortical regions) in the SpatialMonitor it was converted in a `spatial_mask` of only value 1, which was incompatible with the current monitor code (which requires `spatial_mask` to have values starting from 0.
It should be fixed now.

popaula937:
Will merge this solution, if there are other observations, we will address then in a separate task.

PR #153, opened 2020-05-19
TVB-2707: Made SimulatorAdapter always use store references

VinczeRobert:


framework_tvb/tvb/adapters/simulator/simulator_adapter.py
```diff
@@ -387,9 +387,8 @@ def launch(self, view_model):
 
             if self.algorithm.surface:
                 ts_index.fk_surface_gid = self.algorithm.surface.region_mapping_data.surface.gid.hex
-                ts_h5.surface.store(self.algorithm.surface.gid)
-            else:
-                ts_h5.store_references(ts)
+
+            ts_h5.store_references(ts)
```
liadomide, line 391: This looks better.
Couldn't we drop the `if` entirely, and just make sure that `ts_index .fill_from_has_traits(ts)` above fills also the `fk_surface_gid` when necessary ?

liadomide:
I think the task referred should be TVB-2707

VinczeRobert:
> I think the task referred should be TVB-2707

Yeah, you are right. I don't know why I named both the branch TVB-2702....

framework_tvb/tvb/adapters/simulator/simulator_adapter.py
VinczeRobert: Well, I don't think so, I can't take the gid from datatype inside of fill_from_has_traits, I think it only exists on self.algorithm.surface...

popaula937: Actually, if you look in TimeSeriesSurfaceIndex.fill_from_trait() this is already done.
You can do it because TimeSeriesSurface receives the self.algorithm.surface.region_mapping_data.surface and keeps it on its surface attribute when we call ts = monitor.create_time_series inside the SimulatorAdapter.launch().
So, I think this `if` can be dropped, but still, please test the `fk_surface_gid` is filled correctly aftrewards.

VinczeRobert: Yeah, you are right, it works without that line as well.

PR #155, opened 2020-05-20
TVB-2703. Improve Simulation Configure - Persist - Launch flow

popaula937:
Encountered another special use-case if we want to use methods like load_with_references for ProjectionMonitor: it has obsnoise, a reference to Noise, which cannot go through the current reference loading procedure, because it looks for the datatype in DB, and Noise won't be kept there.
I am wondering whether would be helpful to have a different Reference object for configuration objects like: monitor, coupling, model, noise. And within load_with_references treat in different manners the references towards datatypes and references to configurations.


framework_tvb/tvb/adapters/simulator/simulator_adapter.py
```diff
@@ -230,9 +230,15 @@ def _prepare_simulator_from_view_model(self, view_model):
         simulator.model = view_model.model
         simulator.integrator = view_model.integrator
         simulator.initial_conditions = view_model.initial_conditions
-        simulator.monitors = view_model.monitors
         simulator.simulation_length = view_model.simulation_length
 
+        monitors = list()
+        for monitor in view_model.monitors:
+            monitor,  _ = h5.load_with_references_from_dir(self.storage_path, monitor.gid, type(monitor))
```
popaula937, line 237: This line does not work for ProjectionMonitor as described above.

liadomide:
After a short brainstorm with Paula: it will be worth to try that `load_with_references _from_dir` first always looks in the current dir, and only as a fallback when the GID is not found there, it looks into the DB. This has chances to work both on HPC and on normal web deploy.
We could try to make it generic enough to function as a load mechanism for the entire simulatorH5.
As for the performance part we could cache as map[guid, file_name] at the first pass over DirLoader all the encountered H5 files.

framework_tvb/tvb/core/neocom/_registry.py
```diff
@@ -29,10 +29,21 @@
...
+from tvb.adapters.datatypes.h5.projections_h5 import ProjectionMatrixH5
+from tvb.adapters.datatypes.h5.sensors_h5 import SensorsH5
+from tvb.adapters.datatypes.h5.surface_h5 import SurfaceH5
 from tvb.basic.neotraits.api import HasTraits
 from tvb.core.entities.model.model_datatype import DataType
 from tvb.core.neotraits.h5 import H5File
 from tvb.core.neotraits.db import HasTraitsIndex
+from tvb.datatypes.projections import make_proj_matrix, ProjectionMatrix
+from tvb.datatypes.sensors import make_sensors
+from tvb.datatypes.surfaces import make_surface
```
liadomide, line 46: These specific imports here are not nice. 
They are breaking the module dependencies we want to achieve.
Could these go into tvb.config.init.data_registry.py ?

framework_tvb/tvb/core/neocom/_registry.py
```diff
@@ -57,10 +68,24 @@ def get_h5file_for_datatype(self, datatype_class):
                 return self._h5file_for_datatype[base]
         return H5File
 
-    def get_datatype_for_h5file(self, h5file_class):
+    def get_base_datatype_for_h5file(self, h5file_class):
         # type: (typing.Type[H5File]) -> typing.Type[HasTraits]
         return self._datatype_for_h5file[h5file_class]
 
+    def _determine_proper_datatype_from_h5(self, h5file, subtype):
+        dct = {SurfaceH5: partial(make_surface, subtype),
```
liadomide, line 76: Could it work to add 4th optional parameter in `Registry.register_datatype()`
something like `REGISTRY.register_datatype(Surface, SurfaceH5, SurfaceIndex, make_surface)` ?

popaula937: Yes, I changed the code to work as suggested, thanks!

popaula937:
Done from my PoV, tests are also passing. We can merge if there are no further comments.

PR #156, opened 2020-05-21
HBP-81 Use the validate flag also for keycloak users.

bvalean:
By default all keycloak users are validated. If an invalidated user attempts to login we will show an error message: "Your account is not validated. Please contact us at support@thevirtualbrain.org for more details". Keycloak users who will have the ADMINISTRATOR role assigned in the keycloak system will be able to validate/invalidate users.

### DIFF
--- a/framework_tvb/tvb/core/services/user_service.py
+++ b/framework_tvb/tvb/core/services/user_service.py
@@ -197,10 +197,10 @@ class UserService:
                 return False
             user.validated = True
             user = dao.store_entity(user)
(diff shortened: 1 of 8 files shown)


liadomide:
I would leave this not merged for another day, hopefully we get a conformation from Petra by then.

liadomide:
Petra also approved this direction by email. 
Let me know when you are done, and we can merge

bvalean:
I am ready. We can merge this. In the latest commit I fixed also a small problem related to the requested scopes when the user is aleardy logged in the keycloak system.

PR #157, opened 2020-05-22
TVB-2671: fix display titles, reuse and improve old implementation

adrian-dordea:


framework_tvb/tvb/interfaces/web/templates/jinja2/burst/main_burst.html
```diff
@@ -53,7 +53,7 @@ <h4><mark>Parameter Space Exploration</mark> <span id="title-visualizers"></span
 	
 	<!--! Configure + View portlets for selected Burst -->
 	<section id="section-portlets" class="view-portlets view-column col-3">
-		<h4><mark>Visualize</mark> <span id="title-visualizers">New simulation</span></h4>
+		<h4><mark>Visualize</mark> <span id="section-title-visualizers">New simulation</span></h4>
```
liadomide, line 56: I think we should keep the previous name.
It was ok the "duplication" because it was either one or another section to display, and both should change name on selected burst switch

framework_tvb/tvb/interfaces/web/static/js/bursts.js
```diff
@@ -579,7 +583,7 @@ function markBurstChanged() {
 function fill_burst_name(burstName, isReadOnly, addPrefix) {
     const inputBurstName = $("#input-burst-name-id");
     const titleSimulation = $("#title-simulation");
-    const titlePortlets = $("#title-visualizers");
+    const titlePortlets = $("#section-title-visualizers");
```
liadomide, line 586: I suggest keep the previous ID

liadomide:
This also needed a rebase with master for jenkins tests to be ok.
I would appreciate some more manual testing on web GUI. If still ok, we can merge.

framework_tvb/tvb/interfaces/web/templates/jinja2/burst/main_burst.html
adrian-dordea: If we keep it like this the title in for the Visualise section will never update.

liadomide: Why not? I also changed the JS?
Have you tested after a clear browser cache ?

PR #158, opened 2020-05-22
Tvb 2672 - Fix burst display in history

adrian-dordea:


framework_tvb/tvb/core/services/burst_service.py
```diff
@@ -102,6 +103,11 @@ def persist_operation_state(self, operation, operation_status, message=None):
         dao.store_entity(operation)
         operation = dao.get_operation_by_id(operation.id)
         self.file_helper.write_operation_metadata(operation)
+        #update burst also
+        burst_id = json.loads(operation.meta_data)[DataTypeMetaData.KEY_BURST]
+        if burst_id:
+            burst_config = self.load_burst_configuration(burst_id)
+            self.mark_burst_finished(burst_config, burst_config.BURST_ERROR, message)
```
liadomide, line 110: This can not be right for the following reasons:
- it should be `burst_config = self.get_burst_for_operation_id(operation.id)` (instead of parsing the `operation.meta_data`)
- not always an operation has a linked burst, so a check for `if burst_config is None:` afterwards will be necessary
- in this method we receive `operation_status`, (could be OK, ERROR, CANCELED) while in the new code it is always put `burst_config.BURST_ERROR`
- I see places, like the `backend_client.py` were `burst_service.persist_operation_state` and `burst_service.mark_burst_finished` are called consecutively. With the new code the BurstConfig will be unnecessarily updated twice

liadomide:
This needed also a rebase with master.
Unit-tests seem to be fine now.
I would appreciate some manual testing in web GUI after recent review changes. 
If still ok, we can merge

framework_tvb/tvb/interfaces/web/static/js/bursts.js
```diff
@@ -101,6 +101,9 @@ function copyBurst(burstID, first_wizzard_form_url) {
                 success: function (response) {
                     let simParamElem = $("#div-simulator-parameters");
                     simParamElem.html(response);
+                    const selectedBurst = $("#burst_id_" + burstID)[0];
+                    changeBurstHistory(burstID, false);
+                    selectedBurst.className = selectedBurst.className + ' ' + ACTIVE_BURST_CLASS;
```
liadomide, line 106: In such cases it is preferable to use JQuery `addClass`

liadomide:
I did some testing I think it is still not functions properly for all cases (headers change and active style on burst li):
- copy
- load-read-only
- click "new simulation" button
- launch a new Sim/PSE and refresh after the op has finished
I have initiated a small review on burst.js for the above cases

liadomide:
> I did some testing I think it is still not functions properly for all cases (headers change and active style on burst li):
> 
> * copy
> * load-read-only
> * click "new simulation" button
> * launch a new Sim/PSE and refresh after the op has finished
>   I have initiated a small review on burst.js for the above cases

I think I fixed the strange behavior on all of the above cases, except on Copy. That is linked with a review in Controller and will be part of another task

PR #159, opened 2020-05-26
TVB-2659: Make "Next" buttons accessible by keyboard

VinczeRobert:
So far I have made possible to run a simulation with default parameters by setting initial focus on the next buttons or on the launch/launch_pse in their respective fragments. When the focus is not on any active element (in that case the focus is by default on the body element) at the press of any arrow key, the focus will be restored to the next or launch buttons. Also, using left and right arrows it is possible to navigate through the buttons. 

My first problem was that all the next buttons in every fragment have the same id, so I renamed them to "next/burst/{current_fragment_name}".

I don't like that I have a little hardcoded things there, so I'll try to find a better solution to get the initial fragment or to get only the fragment url out of the one having "localhost....". Besides these, should the select fields also be accessible by keyboards? Maybe we could use left-right only on the buttons and up-down on the data fields?

### DIFF
--- a/framework_tvb/tvb/interfaces/web/static/js/bursts.js
+++ b/framework_tvb/tvb/interfaces/web/static/js/bursts.js
@@ -52,7 +52,6 @@ function clone(object_) {
  * When clicking on the New Burst Button reset to defaults for simulator interface and portlets.
  */
 function resetToNewBurst() {
-    document.getElementById('next/burst/set_connectivity').focus();
     doAjaxCall({
         type: "POST",
         url: '/burst/reset_simulator_configuration/',
@@ -62,6 +61,7 @@ function resetToNewBurst() {
             displayBurstTree(undefined);
             displayMessage("Completely new configuration loaded!");
             changeBurstHistory(null, false, false, '');
+            document.getElementsByClassName('btn btn-primary next')[0].focus();
         },
         error: function () {
             displayMessage("We encountered an error while generating the new simulation. Please try reload and then check the logs!", "errorMessage");
@@ -99,13 +99,14 @@ function copyBurst(burstID, first_wizzard_form_url) {
 }
 
 function _renderAllSimulatorForms(url, stop_at_url = '') {
+    const simulator_params = document.getElementById('div-simulator-parameters');
     if (stop_at_url !== url) {
         doAjaxCall({
             type: "GET",
             url: url,
             success: function (response) {
                 var t = document.createRange().createContextualFragment(response);
-                document.getElementById('div-simulator-parameters').appendChild(t);
+                simulator_params.appendChild(t);
                 MathJax.Hub.Queue(["Typeset", MathJax.Hub, "div-simulator-parameters"]);
 
                 next_url = $(response).attr("action");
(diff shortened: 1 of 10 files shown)


liadomide:
Regarding navigation conventions. I think TAB / SHIFT+TAB is the common used paradigm for navigating between input fields / buttons / select fields /etc. The arrows are common to navigate inside a text inout field (e.g. to edit a prefilled text) or for changing a select choice.  But for sure we can find these conventions properly documented somewhere. Pls try make a research what is most common.

framework_tvb/tvb/interfaces/web/static/js/bursts.js
```diff
@@ -435,6 +437,28 @@ function setPseRangeParameters(){
...
 
+function setArrowKeyEventListener(){
+    document.addEventListener('keyup', (e) => {
+        // if focus is not on any of the current fragment's elements, set it on the next or launch button
+        const focused_element = document.activeElement;
+        if(focused_element.tagName === 'BODY'){
+            const simulator_form = document.getElementById('div-simulator-parameters');
+            setInitialFocusOnButton(simulator_form);
+        }else {
+            if (e.code === 'ArrowRight'){
```
liadomide, line 448: I think that if we leave the default (access next by TAB) behaviour, then we only need to make sure that attribute `tabindex ` is properly set in the HTML and that the initial focus is set correctly.
This needs to be checked. Something on this direction already happens in the 1.5.* versions https://www.thevirtualbrain.org/burst/

VinczeRobert: I updated my code. It all works fine.

Only one observation: I have checked the 1.5 version and it's true that tabindex="1" is set on each element but pressing tab takes you through all elements, including the simulations from burst history, the main menu from the lower part of the page and so on, so it's not just the elements from the form. And here I set tabindex only on the buttons cause I don't know how to put it on the inputs, select fields etc. cause they are dinamically created. But even without doing that, it still functions the same way as in the 1.5 version.

liadomide:
A rebase with master will be required, to fix the conflict on bursts.js

framework_tvb/tvb/interfaces/web/static/js/bursts.js
```diff
@@ -661,9 +665,21 @@ function launchNewBurst(currentForm, launchMode) {
...
 }
 
+function setInitialFocusOnButton(simulator_params){
+    const current_url = simulator_params.lastElementChild.action;
+    if(current_url !== undefined && current_url.includes('setup_pse')){
+        document.getElementById('launch_simulation').focus();
+    }else if(current_url !== undefined && current_url.includes('launch_pse')){
+        document.getElementById('launch_pse').focus();
+    }else {
+        const next_buttons = document.getElementsByClassName('btn btn-primary next');
```
liadomide, line 675: I suggest here we make a specific class, easy to search at a possible future rename.
e.g. 'next-button'

liadomide: It is preferably to use JQuery `$('#launch_simulation').focus()` instead of `document.getElementById('launch_simulation').focus()` as usually is faster and more realiable

PR #160, opened 2020-05-27
TVB-2644: Fix AnimatedViewer on TS obtained with Global Average Monitor

VinczeRobert:
In case of page_chunk_data has only one shape (x), it should have the shape (x, 1) so the next for doesn't fail. 

I made some tests and the only visualizer that it has (Animated Time Series) works. Only trying to launch a PSE on this won't work...

### DIFF
--- a/framework_tvb/tvb/adapters/visualizers/eeg_monitor.py
+++ b/framework_tvb/tvb/adapters/visualizers/eeg_monitor.py
@@ -349,9 +349,6 @@ class EegMonitor(ABCSpaceDisplayer):
 
             page_chunk_data = timeseries.read_data_page(self.current_page * self.page_size,
                                                         (self.current_page + 1) * self.page_size)
-            if len(page_chunk_data.shape) == 1:
-                page_chunk_data = page_chunk_data.reshape(page_chunk_data.shape[0], 1)
-
             channels_per_set.append(int(resulting_shape[1]))
 
             for idx in range(resulting_shape[1]):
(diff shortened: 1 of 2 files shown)


framework_tvb/tvb/adapters/visualizers/eeg_monitor.py
```diff
@@ -349,6 +349,9 @@ def compute_required_info(self, list_of_timeseries):
 
             page_chunk_data = timeseries.read_data_page(self.current_page * self.page_size,
                                                         (self.current_page + 1) * self.page_size)
+            if len(page_chunk_data.shape) == 1:
+                page_chunk_data = page_chunk_data.reshape(page_chunk_data.shape[0], 1)
+
```
liadomide, line 354: If I understand correctly your analysis result, then  I do not think this is the proper place to put the fix.
Looking inside `TimeSeriesH5.read_data_page` we already have there an attempt to not return 1D results. See `time_series_h5.py` lines 161-165. Pls do a debug there and try to fix directly inside `TimeSeriesH5.read_data_page`

liadomide: I out my bet on line 161 ;-)

VinczeRobert: Well, the place you suggested was right, of course. In time_series_h5.py there data has a 4d shape, for example (1276, 1, 76, 1) and we should test the element at the second index, cause after squeezing this array for example will have the shape (1276, 76). So if the second index has a 1 element, then we should not let it 1D.

liadomide: Your recent change wound indeed fix the current problem, but we might still have other 4D data which is still failing.
Maybe do the `data.squeeze()` before (for all cases) and check after in a condition similar to what you initially proposed `if len(page_chunk_data.shape) == 1:`. Off course, all these in `time_series_h5.py`.

VinczeRobert: I actually thought about the same thing, cause squeeze will either result in a 2D shape or in a 1D shape and if it's 1D then I do the same thing as originally. :)

PR #161, opened 2020-05-27
TVB-2722: Made buttons invisible after launching

VinczeRobert:
I made the buttons invisible after launching, but the problem that I couldn't solve so far is that if you refresh the simulation before it finishes, the buttons will reappear. I would be happy hearing some suggestions for that :)

### DIFF
--- a/framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
+++ b/framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
@@ -165,8 +165,7 @@ class SimulatorFragmentRenderingRules(object):
 
     @property
     def hide_previous_button(self):
-        if self.load_readonly and not (self.is_simulation_copy and self.is_launch_fragment and
-                                       self.last_form_url == SimulatorWizzardURLs.SETUP_PSE_URL):
+        if self.load_readonly:
             return True
         return False
 
--- a/framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
+++ b/framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
@@ -260,7 +260,12 @@ class SimulatorController(BurstBaseController):
             burst_config = BurstConfiguration(project.id)
             common.add2session(common.KEY_BURST_CONFIG, burst_config)
 
-        is_simulator_load = common.get_from_session(common.KEY_IS_SIMULATOR_LOAD) or False
+        if burst_config.start_time is not None:
+            is_simulator_load = True
+            common.add2session(common.KEY_IS_SIMULATOR_LOAD, is_simulator_load)
+        else:
+            is_simulator_load = common.get_from_session(common.KEY_IS_SIMULATOR_LOAD) or False
+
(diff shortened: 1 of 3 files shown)


liadomide:
Pls rebase after recent merge of TVB-2672
For the page refresh we have the JS function `initBurstConfiguration` in bursts.js. I suggest you try there to see what can/should be done with the buttons.

VinczeRobert:
I (think) have solved this task by using start_time as a way to know if a simulation has been already launched or not at refreshing.

The reason why fk_simulation_id is None after launching and refreshing is because the burst_config in session is not updated after it is stored in the database.

I have also found another problem which could have something to do with my previous paragraph and this is about copying/branching a simulation (I think we may have discussed this already, I don't remember exactly). For example, let's say I launched a simulation and burst.fk_simulation_id is 8. Then I launch a copy or branch from that and that would have burst.fk_simulation_id 9. If I refresh after launching the copy or branch, in index the burst_config will have fk.simulation_id 8 so basically the simulation FROM WHICH we branched or copied will appear, not the one that we just launched, and in this case the buttons will appear...

liadomide:
Have you tested also with "set model params"/"noise..." and comming back to the Sim page, to check that everything is still correctly set ?

VinczeRobert:
> Have you tested also with "set model params"/"noise..." and comming back to the Sim page, to check that everything is still correctly set ?

Yes, I did now. Everything is still correct.

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -258,7 +258,12 @@ def index(self):
...
             common.add2session(common.KEY_BURST_CONFIG, burst_config)
 
-        is_simulator_load = common.get_from_session(common.KEY_IS_SIMULATOR_LOAD) or False
+        if burst_config.start_time is not None:
+            is_simulator_load = True
+            common.add2session(common.KEY_IS_SIMULATOR_LOAD, is_simulator_load)
+        else:
+            is_simulator_load = common.get_from_session(common.KEY_IS_SIMULATOR_LOAD) or False
+
         is_simulator_copy = common.get_from_session(common.KEY_IS_SIMULATOR_COPY) or False
```
popaula937, line 267: Shouldn't this be inside `else` too? What happens when you copy a simulation, launch it and then refresh the page?

VinczeRobert: What I said in another comment:
"I have also found another problem which could have something to do with my previous paragraph and this is about copying/branching a simulation (I think we may have discussed this already, I don't remember exactly). For example, let's say I launched a simulation and burst.fk_simulation_id is 8. Then I launch a copy or branch from that and that would have burst.fk_simulation_id 9. If I refresh after launching the copy or branch, in index the burst_config will have fk.simulation_id 8 so basically the simulation FROM WHICH we branched or copied will appear, not the one that we just launched, and in this case the buttons will appear...".

Now there is a task for this which I'm currently working on (TVB-2723).

In your question did you mean to put the line with is_simulator_copy in else?

PR #162, opened 2020-05-28
WIP TVB-2687: Fix display of details for finished Operations in web GUI

adrian-dordea:
Web GUI - Project - Operations - one finished operations with Success - Button Details - should show all changed input params different from the default.
- fix operation.parameters at initiate_prelaunch

### DIFF
--- a/framework_tvb/tvb/core/adapters/abcadapter.py
+++ b/framework_tvb/tvb/core/adapters/abcadapter.py
@@ -524,24 +524,23 @@ class ABCAdapter(object):
             LOGGER.exception(msg)
             raise IntrospectionException(msg)
 
-    def review_operation_inputs(self, adapter, operation):
+    def review_operation_inputs(self, operation):
         """
         :returns: a list with the inputs from the parameters list that are instances of DataType,\
             and a dictionary with all parameters which are different than the declared defauts
         """
         changed_attr = {}
         inputs_datatypes = []
-        view_model = self.load_view_model(adapter, operation)
-        form_model = adapter.get_view_model_class()()
-        form_fields = adapter.get_form_class()().fields
+        view_model = self.load_view_model(self, operation)
+        form_model = self.get_view_model_class()()
+        form_fields = self.get_form_class()().fields
 
         for field in form_fields:
             if not isinstance(field, TraitDataTypeSelectField):
                 attr_form = getattr(form_model, field.name)
                 attr_vm = getattr(view_model, field.name)
                 if attr_vm != attr_form:
-                    field_name = field.name.title().replace("_", " ")
-                    changed_attr[field_name] = attr_vm
+                    changed_attr[field.label] = attr_vm
             else:
                 attr_vm = getattr(view_model, field.name)
                 data_type = ABCAdapter.load_entity_by_gid(attr_vm)
--- a/framework_tvb/tvb/core/adapters/abcadapter.py
+++ b/framework_tvb/tvb/core/adapters/abcadapter.py
@@ -53,6 +53,7 @@ from tvb.core.adapters import constants
 from tvb.core.entities.generic_attributes import GenericAttributes
 from tvb.core.entities.load import load_entity_by_gid
 from tvb.core.neocom import h5
+from tvb.core.neotraits._h5core import ViewModelH5
 from tvb.core.neotraits.h5 import H5File
 from tvb.core.utils import date2string, LESS_COMPLEX_TIME_FORMAT
 from tvb.core.entities.storage import dao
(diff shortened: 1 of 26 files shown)


framework_tvb/tvb/core/services/project_service.py
```diff
@@ -790,7 +790,7 @@ def _review_operation_inputs(self, operation_gid):
         parameters = json.loads(operation.parameters)
         try:
             adapter = ABCAdapter.build_adapter(operation.algorithm)
-            return adapter.review_operation_inputs(parameters)
+            return adapter.review_operation_inputs(adapter, operation)
```
liadomide, line 793: Why the need to pass `adapter ` as parameter? Shouldn't it be `self` ?

framework_tvb/tvb/core/adapters/abcadapter.py
```diff
@@ -523,13 +524,39 @@ def build_adapter(stored_adapter):
...
+                if attr_vm != attr_form:
+                    changed_attr[field.label] = attr_vm
+            else:
+                attr_vm = getattr(view_model, field.name)
+                data_type = ABCAdapter.load_entity_by_gid(attr_vm)
+                inputs_datatypes.append(data_type)
+
+        return inputs_datatypes, changed_attr
+
+    def load_view_model(self, adapter_instance, operation):
```
popaula937, line 551: This method is duplicated. We also have it in OperationService. 
Have you tested this for simulations? This would probably not work for simulations, since we have to use the SimulatorSerializer there, as in the other method.
Please review this and keep it in a single common place.

framework_tvb/tvb/core/adapters/abcadapter.py
```diff
@@ -523,13 +524,39 @@ def build_adapter(stored_adapter):
...
-        return {}, None
+        changed_attr = {}
+        inputs_datatypes = []
+        view_model = self.load_view_model(self, operation)
+        form_model = self.get_view_model_class()()
+        form_fields = self.get_form_class()().fields
+
+        for field in form_fields:
+            if not isinstance(field, TraitDataTypeSelectField):
+                attr_form = getattr(form_model, field.name)
```
popaula937, line 540: I would call this attr_default, to be more clear what we are comparing here.
This is ok for handling analyzers/creators, but I don't think it is enough to support also simulations.

framework_tvb/tvb/core/services/project_service.py
```diff
@@ -790,7 +790,7 @@ def _review_operation_inputs(self, operation_gid):
         parameters = json.loads(operation.parameters)
         try:
             adapter = ABCAdapter.build_adapter(operation.algorithm)
-            return adapter.review_operation_inputs(parameters)
+            return adapter.review_operation_inputs(operation)
```
popaula937, line 793: I think we should also change the exception handling here. Remove TODO once this is done.

framework_tvb/tvb/core/services/operation_service.py
```diff
@@ -303,6 +304,53 @@ def load_view_model(self, adapter_instance, operation):
...
+                 And a dictionary will all operation parameters different then the default ones.
+        """
+        operation = dao.get_operation_by_gid(operation_gid)
+        parameters = json.loads(operation.parameters)
+        try:
+            adapter = ABCAdapter.build_adapter(operation.algorithm)
+            return self.review_operation_inputs(adapter, operation)
+
+        except Exception:
+            self.logger.exception("Could not load details for operation %s" % operation_gid)
```
liadomide, line 319: When exactly is this exception happening?
I suspect is when importing old operations (e.g. DefaultProject in previous forms). Is this correct?
If so, then pls adjust logged messages accordingly.

framework_tvb/tvb/core/services/project_service.py
```diff
@@ -760,7 +760,7 @@ def get_datatype_and_datatypegroup_inputs_for_operation(self, operation_gid, sel
         If any dataType is part of a dataType group then the dataType group will
         be returned instead of that dataType.
         """
-        all_datatypes = self._review_operation_inputs(operation_gid)[0]
+        all_datatypes = OperationService()._review_operation_inputs(operation_gid)[0]
```
liadomide, line 763: You should not call a `private` method from outside, unless some special reason.
Here, you could simply rename the OperationService method `_review_operation_inputs ` into something without leading `_`

liadomide:
> Reading through the lines, the changes look good to me, but I haven't tested the GUI.

The main thing I expected from a reviewer is to read the code. 
Whoever did the changes, or who is owner of the task should do the intensive manual testing + Jenkins the automated ones.
The reviewer could suggest tricky testing scenarios, though.

liadomide:
@popaula937 pls have another look at the Stimulus related forms changes. 
I removed the extra param on init related to equation choices and instead compute that directly in the form.
Otherwise, the forms changed params introspection has no chances to work for these operations, I think.

popaula937:
> @popaula937 pls have another look at the Stimulus related forms changes.
> I removed the extra param on init related to equation choices and instead compute that directly in the form.
> Otherwise, the forms changed params introspection has no chances to work for these operations, I think.

Yes, you are right it would not work in the current form. I think these changes are ok and it is even better to keep the choices inside the forms, now that they are used only for that.

framework_tvb/tvb/core/adapters/inputs_processor.py
```diff
@@ -0,0 +1,94 @@
...
+                                        changed_attr[field.label] = sub_attr.title
+                    else:
+                        attr_vm = getattr(view_model, field.name)
+                        data_type = ABCAdapter.load_entity_by_gid(attr_vm)
+                        if attr_vm:
+                            changed_attr[field.label] = data_type.display_name
+                        inputs_datatypes.append(data_type)
+    else:
+        inputs_datatypes, changed_attr = _review_operation_inputs_for_adapter_model(form_fields, form_model, view_model)
+
```
popaula937, line 93: Maybe it would be easier to read this method if the` if-else` statements were inversed (maybe no `else` needed if the return is done inside the `if`)

liadomide:
I merge this, as it is blocking other refactors due to multiple files being touched.
The task will still remain open until unit-tests are re-written and this is validated thoroughly.

PR #164, opened 2020-05-29
TVB-2724: implement cancel and remove of burst

adrian-dordea:
- Implement the Cancel and Remove of BurstConfiguration after recent changes in model
- refactor services to avoid cyclic dependency

### DIFF
--- a/framework_tvb/tvb/core/services/project_service.py
+++ b/framework_tvb/tvb/core/services/project_service.py
@@ -54,7 +54,7 @@ from tvb.core.entities.file.exceptions import FileStructureException
 from tvb.core.neocom import h5
 from tvb.core.neotraits.h5 import H5File
 from tvb.core.removers_factory import get_remover
-from tvb.core.services.flow_service import FlowService
+from tvb.core.services.algorithm_service import AlgorithmService
 from tvb.core.services.exceptions import StructureException, ProjectServiceException
 from tvb.core.services.exceptions import RemoveDataTypeException
 from tvb.core.services.user_service import UserService, MEMBERS_PAGE_SIZE
@@ -211,7 +211,7 @@ class ProjectService:
                         result["datatype_group_gid"] = datatype_group.gid
                         result["gid"] = operation_group.gid
                         ## Filter only viewers for current DataTypeGroup entity:
-                        result["view_groups"] = FlowService().get_visualizers_for_group(datatype_group.gid)
+                        result["view_groups"] = AlgorithmService().get_visualizers_for_group(datatype_group.gid)
                     except Exception:
                         self.logger.exception("We will ignore group on entity:" + str(one_op))
                         result["datatype_group_gid"] = None
--- a/framework_tvb/tvb/core/services/project_service.py
+++ b/framework_tvb/tvb/core/services/project_service.py
@@ -45,6 +45,12 @@ from tvb.core.adapters.inputs_processor import review_operation_inputs_from_adap
 from tvb.core.entities.model.model_datatype import Links, DataType, DataTypeGroup
 from tvb.core.entities.model.model_operation import Operation, OperationGroup
 from tvb.core.entities.model.model_project import Project
+from tvb.core.neocom import h5
+from tvb.core.neotraits.h5 import H5File
+from tvb.core.services.backend_client import BACKEND_CLIENT
+from tvb.core.services.flow_service import FlowService
+from tvb.core.utils import string2date, date2string, format_timedelta, format_bytes_human
+from tvb.core.removers_factory import get_remover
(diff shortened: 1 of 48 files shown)


framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -239,6 +240,10 @@ def _update_last_loaded_fragment_url(self, current_url):
         self.last_loaded_form_url = current_url
         common.add2session(common.KEY_LAST_LOADED_FORM_URL, self.last_loaded_form_url)
 
+    @cherrypy.expose
+    def cancel_or_remove_burst(self, burst_id):
+        ProjectService().cancel_or_remove_burst(burst_id)
```
popaula937, line 245: On the client-side, I see it expects an answer from this method.
Please check how this method handled results in the old tvb version:
https://github.com/the-virtual-brain/tvb-framework/blob/0db0f63c3f5c5485063cda014aa8698e23f709df/tvb/interfaces/web/controllers/burst/burst_controller.py#L423

framework_tvb/tvb/core/services/project_service.py
```diff
@@ -851,6 +852,33 @@ def get_operations_for_datatype(datatype_gid, visibility_filter, only_in_groups=
...
+            raise RemoveDataTypeException("Could not remove Burst entity!")
+
+        for datatype in datatypes:
+            ProjectService().remove_datatype(burst_entity.fk_project, datatype.gid, False)
+
+    def stop_burst(self, operation_id):
+        """
+        Stop the operation given by the operation id.
+        """
+        return BACKEND_CLIENT.stop_operation(int(operation_id))
```
popaula937, line 880: Call stop_operation here.
Also, shouldn't this handle stop also for PSE bursts?

adrian-dordea: The implemented code works for stop/cancel and also remove for both normal and PSE bursts.

framework_tvb/tvb/core/services/project_service.py
```diff
@@ -851,6 +852,36 @@ def get_operations_for_datatype(datatype_gid, visibility_filter, only_in_groups=
             return dao.get_operations_for_datatype(datatype_gid, only_relevant=False, only_in_groups=only_in_groups)
         return dao.get_operations_for_datatype(datatype_gid, only_in_groups=only_in_groups)
 
+    def cancel_or_remove_burst(self, burst_id):
```
liadomide, line 855: Doesn't it make more sense for these 2 methods to remain in `BurstService` (as they are referring to a BurstConfig instance)?

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
```diff
@@ -696,7 +697,7 @@ def get_adapter_template(self, project_id, algorithm_id, is_upload=False, back_p
     def reloadoperation(self, operation_id, **_):
         """Redirect to Operation Input selection page, 
         with input data already selected."""
-        operation = self.flow_service.load_operation(operation_id)
+        operation = self.operation_services.load_operation(operation_id)
```
bvalean, line 700: Maybe call the static load_operation directly from the OperationService class? In this way we will keep the consistency in the flow controller.

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -234,13 +235,37 @@ def __init__(self):
...
+            :returns 'canceled': When current burst was still running and was just stopped.
+            :returns 'done': When no action is required on the client.
+        """
+        burst_id = int(burst_id)
+        session_burst = common.get_from_session(common.KEY_BURST_CONFIG)
+
+        if session_burst.status == session_burst.BURST_RUNNING:
+            OperationService().stop_operation(session_burst.fk_simulation)
+            # Burst was stopped since it was running
+            return 'canceled'
```
bvalean, line 261: Is it a good idea to keep these "statuses" in an enum?

liadomide: In principle it is a very good idea to use Enums. 
In exactly this case, the only other place where these strings are used is in bursts.js function cancelOrRemoveBurst, thus it would require a manual synchronization between the two parts (py vs JS) anyway. 
I think the entire controllers + JS part would benefit greatly from a rewrite with a heavier (than plain) JS framework, thus I would leave such a refactor for those times. But good that you point this.

liadomide: In the end I dropped the return string entirely, as we don't really need it. The client side can be done smarter.

PR #165, opened 2020-06-03
TVB-2694  Integer input validation + part TVB-2728 ICA Fix

VinczeRobert:
For the integer input validation I made the exact same thing as with floats. I saw there a TODO saying that we should raise an exception but I don't know what kind of exception so I just left it like that.

For ICA, I don't really know what those parameters mean, but based on their shapes, I deduced how they should be rearranged so the analyzer doesn't fail at launching. It works now, and the visualizer works as well.

### DIFF
--- a/framework_tvb/tvb/core/neotraits/forms.py
+++ b/framework_tvb/tvb/core/neotraits/forms.py
@@ -488,7 +488,6 @@ class IntField(TraitField):
 
     def _from_post(self):
         super(IntField, self)._from_post()
-        # TODO: Throws exception if attr is optional and has no value
         if self.unvalidated_data and len(self.unvalidated_data) == 0:
             self.unvalidated_data = None
         if self.unvalidated_data:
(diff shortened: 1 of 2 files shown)


popaula937:
The TODO refers to the case when a field from a Form object is declared optional, we do not give a value for it from GUI, but still, setting it to None inside `_from_post` throws an error. Not sure whether this is still the case since it seems to work correctly for the ICA example. Maybe it would be worth checking for such cases on other types of fields.

On another note, I just tried to run the ICA from GUI, and it finishes with an error if I choose a value for **Nr of components** field. I will put the stack trace on Jira. Please take a look at that.
eg: I choose 3 and received: _**could not broadcast input array from shape (76,3) into shape (3,76)**_

VinczeRobert:
If the tests are fine, I think we can merge this. ICA currently works if that optional field is left empty, and I will handle the rest in TVB-2728.

liadomide:
Tests are ok. 
Agreed, I will merge now, but you might still need to rethink the 2 lines changed in this PR on ica.py when validating ICA algorithm against TVB version  1.5

PR #167, opened 2020-06-05
TVB-2728: Fixed ICA Analyzer

VinczeRobert:
I just went along the lines and fixed the shaped of the matrixes where they didn't match or they couldn't be multiplied. 

Now it works when you leave the components field empty and it works if you choose a value for it as well. Tests also works.

I left the WIP because I still need to compare it to 1.5.6.

### DIFF
--- a/framework_tvb/tvb/adapters/visualizers/ica.py
+++ b/framework_tvb/tvb/adapters/visualizers/ica.py
@@ -111,6 +111,6 @@ class ICA(MappedArraySVGVisualizerMixin):
 
         unmixing_matrix = unmixing_matrix[..., view_model.i_svar, view_model.i_mode]
         prewhitening_matrix = prewhitening_matrix[..., view_model.i_svar, view_model.i_mode]
-        Cinv = unmixing_matrix.dot(prewhitening_matrix)
+        Cinv = unmixing_matrix.dot(prewhitening_matrix.T)
         pars = self.compute_params(Cinv, 'ICA region contribution', '(Ellipsis, %d, 0)' % (view_model.i_svar))
         return self.build_display_result("matrix/svg_view", pars)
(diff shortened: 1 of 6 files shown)


liadomide:
@maedoc-amu We would appreciate your review on the changes done here in ICA.
a "fix" was started from this commit, which unfortunately is not part of this PR https://github.com/the-virtual-brain/tvb-root/pull/165/commits/ff17e8737da8353262673cbd4f099f637bca1a18

The current ICA has subject of a bigger refactor since tvb version 1.5.6
First https://req.thevirtualbrain.org/browse/TVB-2507
and now https://req.thevirtualbrain.org/browse/TVB-2728

Robert will manually test the new algorithm vs the previous version of TVB, but your eye on the code logic would be necessary.

liadomide:
I reverted the change done on another PR https://github.com/the-virtual-brain/tvb-root/pull/165/commits/ff17e8737da8353262673cbd4f099f637bca1a18
Pls rebase before continuing investigations

maedoc:
The transpose that this PR proposes isn't correct, since `fastica` expects data of shape `(n_samples, n_features)` which should be here `(n_time, n_node)`, which matches `data[:, var, :, mode]` without the transpose.

I'm not sure how useful the analysis is in practice, since modes and state variables would be highly correlated, but this is a different question.

VinczeRobert:
I ran the test_fastica simple from sklearn in parallel with our test for ica, and it actually does transpose the source, but it does it inside the function, it doesn't call it transposed as I did.

If you go here: https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/decomposition/_fastica.py at 431 you can see that it calls self._validate_data and at the end there is the .T which transposes it. We don't have that validation and that transposing so that was the difference between our code and theirs that I couldn't find so far. 

In their test they are using a source with shape (1000, 2) so I tried to test our code with the same shape and adding this transposing change and everything would be the same this way.

liadomide:
> If you go here: https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/decomposition/_fastica.py at 431 you can see that it calls self._validate_data and at the end there is the .T which transposes it.

Could you add this validation and transposition step in our code?
Then the top level documentation and signature will be consistent with the ICA original algorithm, and should also work.

PR #168, opened 2020-06-09
TVB-2723: Review Copy Simulation Current Implementation

VinczeRobert:
To me it seems like it works as it should now, but I would appreciate some review.


VinczeRobert:
Actually no, it doesn't work completely as it fails if you try to branch it. I will fix it.

PR #172, opened 2020-06-13
WIP Add resting-state HRF mediated Bold response using input from text file

AmoghJohri:
This supplements the existing workflow for BOLD simulation. Instead of computing an identical HRF for all the regions, we use region-wise HRF (obtained by deconvolving the emperical region-wise BOLD Time-Series) for simulations (https://www.nitrc.org/projects/rshrf, https://github.com/BIDS-Apps/rsHRF). The rsHRF is provided as input from a file.

**Additions:**
1. _rsHRF_file_ attribute which takes the path of the input file.
2. __from_file_ variable (binary: {0, 1}) which signifies whether HRF is computed or taken as input.
3. _get_hrf_ function hich obtains the input from the text-file.
4. changes in _compute_hrf_  to upsample the input HRF for convolution.
5. changes in _sample_ to perform region-wise convolution.

This does not make any alterations to the existing workflow for BOLD simulation. In case of no input-file being provided, the default workflow is adopted.


maedoc:
I would prefer not to have the monitors loading stuff from files.  It would be better if you can implement a fixed equation type which loads from file, and then provide the ROI -wise HRFs to the BOLD monitor which can apply them as with the existing Volterra kernel.

AmoghJohri:
I am not sure if I completely understood that, is this what was required?

maedoc:
Yes, this is better, thanks.  Do you have a demo of this, such as a Jupyter notebook which shows the differences in simulation?  It would be good to document how/why of this feature.

maedoc:
Closing for lack fo response, feel free to reopen if you wish to follow up.

PR #174, opened 2020-06-15
TVB-2690 RegionSelector does not preserve hemisphere ordering

liadomide:
In recent TVB web versions, setting scaling on Stimulus Region is not preserving order after submit+save+reload.


liadomide:
This closes also TVB-2683

PR #175, opened 2020-06-16
TVB-2709 Fix launching simulations using EpileptorCodim3 models on Windows

bvalean:
The problem was that in the guvectorize decorator's  signature the int_ dtype was evaluated to int32 on Windows, and the type cast check was failing.


maedoc:
Change looks fine to me, but it is still suspicious.  Could be a bug in Numba? `numba.int_` should match `numpy.r_[5].dtype` (which is true on macOS at least)

PR #177, opened 2020-06-16
DSL: svboundaries and flexible path

DeLaVlag:
@maedoc Fixed 1 bug with svboundaries argument and added the capability for the user to add a path to his own XML model directory. The latter is because the pypi installation defaults the xml model location to installation and not working directories.


scientific_library/tvb/dsl/LEMS2python.py
```diff
@@ -106,4 +112,9 @@ def regTVB_templating(model_filename):
                 f.seek(0)
                 f.writelines(lines)
     except:
-        print('unable to add new model to __init__.py')
\ No newline at end of file
+        print('unable to add new model to __init__.py')
```
liadomide, line 115: Maybe you could use logging instead of print calls in the entire module ?

scientific_library/tvb/dsl/LEMS2python.py
```diff
@@ -66,19 +66,25 @@ def render_model(model_name, template=None):
...
+    example model files:
+        epileptort.xml
+        generic2doscillatort.xml
+        kuramotot.xml
+        montbriot.xml
+        reducedwongwangt.xml
     """
 
     # file locations
     modelfile = "{}{}{}{}".format(os.path.dirname(tvb.__file__),'/simulator/models/',model_filename.lower(),'.py')
```
maedoc, line 84: Please use 
```python
import os.path
modelfile = os.path.join(os.path.dirname(tvb.__file__), 'simulator', 'models', …..
```
since it handles the path construction correctly across platforms

scientific_library/tvb/dsl/LEMS2python.py
maedoc: In almost any other TVB module you should see calls to a logger which is defined at the top of the module; please replicate that pattern here.  

Also, no bare except, should be
```python
except IOError as e:
    logger.error('ioerror: %s', e)
```
since the error could be filesystem out of space, user presses ctrl-c, etc.

DeLaVlag:
Hi Guys, I've commited the changes as requested. The requested changes just now had nothing to do with the reason for this PR, so I assume my original changes are fine.

scientific_library/tvb/dsl/LEMS2python.py
DeLaVlag: has been updated as requested

scientific_library/tvb/dsl/LEMS2python.py
DeLaVlag: has been updated as requested

DeLaVlag:
For some reason the PR does not get updated for my latest commit on the __init__.py file. Don't merge these changes.

maedoc:
Do you mean you've pushed a newer commit than the one with the message "reset init for docker tests" to your branch?

DeLaVlag:
No I did not commit anything after "reset init for docker tests". The PR does not seem to get updated on this last commit. It still displays 3 files to be changed in the 'Files changed' section, should be 2.

DeLaVlag:
Can we merge and close this PR now? I think the init should be fine.

scientific_library/tvb/simulator/models/__init__.py
```diff
@@ -92,6 +92,10 @@ def _get_imported_model(model):
 
 
 _module_models = {
+    'reducedwongwangt': [ModelsEnum.REDUCEDWONGWANGT],
+    'reducedwongwang': [ModelsEnum.REDUCEDWONGWANG],
+    'montbrio': [ModelsEnum.MONTBRIO],
+    'generic2doscillator': [ModelsEnum.GENERIC2DOSCILLATOR],
```
liadomide, line 98: These new models won't work for us. 
Neither the modules nor the enum entries exist.

DeLaVlag: The enum will be generated as well when you run RateML (new DSL name). This makes sure that TVB recognizes the newly generate model. This is the file not necessary needs to be committed and I resetted it, but lines needs to be added during runtime. The _find_lems_models is not my work. For me it worked fine without this function.

liadomide: OK. Then I will try to remove the function as part of another review PR.
Thanks for resetting the dictionary.

maedoc:
thx @DeLaVlag !

PR #178, opened 2020-06-18
TVB-2549: Finish tests for Simulator Controller

VinczeRobert:
The last one, which I marked with skip is not working yet. It says the file is not a valid ZIP file even though it's the one I export and it has the zip extension... I have to look why it says that.

86% of the lines are covered. Also, except the one for ZIP I want to write one for launching with branch and maybe a few more where more consecutive lines of code where not covered. Also, I would extract some common code in functions to make it nicer.

### DIFF
--- a/framework_tvb/tvb/core/entities/file/simulator/burst_configuration_h5.py
+++ b/framework_tvb/tvb/core/entities/file/simulator/burst_configuration_h5.py
@@ -62,7 +62,9 @@ class BurstConfigurationH5(H5File):
         burst_config.status = self.status.load()
         burst_config.error_message = self.error_message.load()
         burst_config.start_time = string2date(self.start_time.load())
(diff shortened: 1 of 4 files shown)


liadomide:
This looks good to me now. 
Let us know in case you still want to improve it, or we can merge?

VinczeRobert:
> This looks good to me now.
> Let us know in case you still want to improve it, or we can merge?

Well, I added two more tests. Now the only method which is not covered at all is cancel_or_remove_burst because as far as I know, that method doesn't work yet. So if that's not a problem, this can be merged from my side.

liadomide:
>  Now the only method which is not covered at all is cancel_or_remove_burst because as far as I know, that method doesn't work yet.

That method should work now. There is another task and PR#180 to finish unit-tests for it (at least at the service level).

PR #181, opened 2020-06-26
TVB-2737: Inspect updateDivContent

VinczeRobert:


scientific_library/tvb/analyzers/ica_algorithm.py
```diff
@@ -298,6 +298,7 @@ def g(x, fun_args):
                   " should be one of 'logcosh', 'exp', 'cube' or callable"
                   % fun)
 
+    X = X.T
```
liadomide, line 301: This PR will need a rebase with master

framework_tvb/tvb/interfaces/web/templates/jinja2/flow/reduce_dimension_select.html
```diff
@@ -20,8 +20,7 @@
                 {% for idx in ((data | length) | xrange) %}
                     <td>
                         <select id="dimId_{{ select_name }}_{{ parameters_prefix }}_{{ idx }}"
-                                name='{{ select_name }}_{{ parameters_prefix }}_{{ idx }}' multiple='multiple'
-                                onchange="updateShapeLabel('{{ select_name }}', '{{ parameters_prefix }}', {{ idx }})">
```
liadomide, line 24: This removal does not seem right. 
If needs cut, then we should remove the entire `reduce_dimension_select.html` and method `FlowController.gettemplatefordimensionselect`
But I suspect it is still needed for `MatrixViewer` (together with few JS functions from `stepsUpdate.js`)

liadomide: You are right to remove stepsUpdate.js entirely.
Then I removed also `FlowController.gettemplatefordimensionselect` and `reduce_dimension_select.html` template, as they are dead code.

VinczeRobert: Okay, thank you for removing those, they escaped my notice.

PR #183, opened 2020-07-03
Allow function type for attributes

neich:
Hi,

This is a proposal to allow functions as values for attributes. Right now, if the value is a function it gets evaluated and the result is assigned to the attribute, thus preventing the use of functions as values. This is a very small change that allows to use function types without breaking the previous behavior.

Right now I'm using this type of value as factories


liadomide:
It will be great if you could also add a unit-test in one of the modules under `tvb.tests.library.basic.neotraits`

neich:
I've added a VERY simple test for function types, and it passes and it does not break the remaining tests in `tvb.tests.library.basic.neotraits` module. Let me know if it is enough

PR #188, opened 2020-07-08
TVB-2731: REST API - provide extra information for various TS analysis results

adrian-dordea:
- FCD, CrossCorrelation, CorrelationCoefficients
- populate `labels_ordering` in analyzer adapters correctly

### DIFF
--- a/framework_tvb/tvb/adapters/datatypes/db/temporal_correlations.py
+++ b/framework_tvb/tvb/adapters/datatypes/db/temporal_correlations.py
@@ -44,6 +44,12 @@ class CrossCorrelationIndex(DataTypeMatrix):
 
     labels_ordering = Column(String, nullable=False)
 
+    def get_extra_info(self):
+        labels_dict = {}
+        labels_dict["labels_ordering"] = self.labels_ordering
+        labels_dict["labels_dimensions"] = self.labels_dimensions
+        return labels_dict
+
     def fill_from_has_traits(self, datatype):
         # type: (CrossCorrelation)  -> None
         super(CrossCorrelationIndex, self).fill_from_has_traits(datatype)
--- a/framework_tvb/tvb/adapters/datatypes/db/temporal_correlations.py
+++ b/framework_tvb/tvb/adapters/datatypes/db/temporal_correlations.py
@@ -46,9 +46,12 @@ class CrossCorrelationIndex(DataTypeMatrix):
 
     def get_extra_info(self):
         labels_dict = {}
-        labels_dict["labels_ordering"] = self.labels_ordering
-        labels_dict["labels_dimensions"] = self.labels_dimensions
+        labels_dict["labels_ordering"] = self.source.labels_ordering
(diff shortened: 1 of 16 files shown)


framework_tvb/tvb/adapters/datatypes/db/temporal_correlations.py
```diff
@@ -43,9 +43,19 @@ class CrossCorrelationIndex(DataTypeMatrix):
     source = relationship(TimeSeriesIndex, foreign_keys=fk_source_gid, primaryjoin=TimeSeriesIndex.gid == fk_source_gid)
 
     labels_ordering = Column(String, nullable=False)
+    labels_dimensions = Column(String, nullable=False)
+    subtype = Column(String)
```
liadomide, line 47: Column subtype already exists in DTMatrix, right ?

adrian-dordea:
In the last commit we are using a lazy load instead of making DataTypes depend on DAO. Any other suggestions are wolcome.

framework_tvb/tvb/interfaces/rest/commons/strings.py
```diff
@@ -86,6 +86,7 @@ class RestLink(Enum):
     # DATATYPES
     GET_DATATYPE = "/{" + LinkPlaceholder.DATATYPE_GID.value + "}"
     DATATYPE_OPERATIONS = "/{" + LinkPlaceholder.DATATYPE_GID.value + "}/operations"
+    OPERATION_RESULT = "/{" + LinkPlaceholder.DATATYPE_GID.value + "}/result"
```
liadomide, line 89: Wouldn't this be better named `DATATYPE_EXTRA_INFO` or something to contain `extra_information` in the string ?

framework_tvb/tvb/interfaces/rest/server/resources/datatype/datatype_resource.py
```diff
@@ -47,6 +47,18 @@ def get(self, datatype_gid):
         file_name = os.path.basename(h5_file_path)
         return flask.send_file(h5_file_path, as_attachment=True, attachment_filename=file_name)
 
+class GetOperationResultForDatatypeResource(RestResource):
```
liadomide, line 50: Same rename also here maybe ?

liadomide (approved):
It looks good to me now. 
I only suggest some renaming, for consistency and clarity

framework_tvb/tvb/core/entities/storage/datatype_dao.py
```diff
@@ -393,6 +397,8 @@ def get_datatype_by_gid(self, gid, load_lazy=True):
                 result_dt.parent_operation.algorithm.algorithm_category
                 result_dt.parent_operation.operation_group
                 result_dt._parent_burst
+                if load_lazy_extra_info:
```
bvalean, line 400: Since we are using another flag for the extra info I think it would be better to move the "if load_lazy_extra_info" outside of the "if load_lazy" statement and when we call the get_extra_info method set just the load_lazy_extra_info flag on True.

PR #189, opened 2020-07-08
TVB-2428: Handle UI Filters

VinczeRobert:
I managed to return only the options from the Python side and replace them in the JS side. However, there is a problem that I still need to solve, and that is the fact that default None or All Options are deleted this way, so I guess I can't just delete all options and put the new ones, I have to keep those default options. Also probably some parts of the code could look better, but I just wanted to make the PR now so I can get a review if the way this went is a good one or not.

### DIFF
--- a/framework_tvb/tvb/core/entities/storage/datatype_dao.py
+++ b/framework_tvb/tvb/core/entities/storage/datatype_dao.py
@@ -512,15 +512,14 @@ class DatatypeDAO(RootDAO):
                                        func.max(text('"OPERATION_GROUPS_1".name')),
                                        func.max(DataType.user_tag_1)
                         ).join((Operation, datatype_class.fk_from_operation == Operation.id)
+                        ).outerjoin((BurstConfiguration, datatype_class.fk_from_operation ==
+                                     BurstConfiguration.fk_simulation)
                         ).outerjoin(Links
                         ).outerjoin((OperationGroup, Operation.fk_operation_group ==
                                      OperationGroup.id), aliased=True
                         ).filter(DataType.invalid == False
                         ).filter(or_(Operation.fk_launched_in == project_id,
                                      Links.fk_to_project == project_id))
-
-            if hasattr(datatype_class, 'time_series_type'):
-                query = query.filter(Operation.id == BurstConfiguration.fk_simulation)
 
             if filters:
                 filter_str = filters.get_sql_filter_equivalent(datatype_to_check='datatype_class')
--- a/framework_tvb/tvb/core/entities/storage/datatype_dao.py
+++ b/framework_tvb/tvb/core/entities/storage/datatype_dao.py
@@ -518,6 +518,10 @@ class DatatypeDAO(RootDAO):
                         ).filter(DataType.invalid == False
                         ).filter(or_(Operation.fk_launched_in == project_id,
                                      Links.fk_to_project == project_id))
+
+            if hasattr(datatype_class, 'time_series_type'):
+                query = query.filter(Operation.id == BurstConfiguration.fk_simulation)
+
             if filters:
                 filter_str = filters.get_sql_filter_equivalent(datatype_to_check='datatype_class')
                 if filter_str is not None:
--- a/framework_tvb/tvb/core/entities/storage/datatype_dao.py
+++ b/framework_tvb/tvb/core/entities/storage/datatype_dao.py
@@ -512,8 +512,8 @@ class DatatypeDAO(RootDAO):
                                        func.max(text('"OPERATION_GROUPS_1".name')),
                                        func.max(DataType.user_tag_1)
                         ).join((Operation, datatype_class.fk_from_operation == Operation.id)
-                        ).outerjoin((BurstConfiguration, datatype_class.fk_from_operation ==
-                                     BurstConfiguration.fk_simulation)
+                        ).outerjoin((BurstConfiguration, datatype_class.fk_parent_burst ==
+                                     BurstConfiguration.id)
                         ).outerjoin(Links
                         ).outerjoin((OperationGroup, Operation.fk_operation_group ==
                                      OperationGroup.id), aliased=True
--- a/framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
+++ b/framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
@@ -41,12 +41,14 @@ import cherrypy
 import formencode
 import numpy
 import six
+import sys
 from tvb.core.adapters import constants
 from tvb.core.adapters.abcadapter import ABCAdapter
 from tvb.core.adapters.abcdisplayer import ABCDisplayer
 from tvb.core.adapters.exceptions import LaunchException
 from tvb.core.entities.file.files_helper import FilesHelper
 from tvb.core.entities.filters.chain import FilterChain
+from tvb.core.entities.load import get_filtered_datatypes
 from tvb.core.entities.model.model_burst import BurstConfiguration
 from tvb.core.neocom import h5
 from tvb.core.services.burst_service import BurstService
@@ -58,8 +60,8 @@ from tvb.interfaces.web.controllers import common
 from tvb.interfaces.web.controllers.autologging import traced
 from tvb.interfaces.web.controllers.base_controller import BaseController
 from tvb.interfaces.web.controllers.common import InvalidFormValues
-from tvb.interfaces.web.controllers.decorators import expose_fragment, handle_error, check_user, expose_json
-from tvb.interfaces.web.controllers.decorators import expose_page, settings, context_selected, expose_numpy_array
+from tvb.interfaces.web.controllers.decorators import expose_page, settings, context_selected, expose_numpy_array, \
+    expose_fragment, handle_error, check_user, expose_json
 from tvb.interfaces.web.entities.context_selected_adapter import SelectedAdapterContext
 
 KEY_CONTENT = ABCDisplayer.KEY_CONTENT
@@ -249,27 +251,28 @@ class FlowController(BaseController):
         return template_specification
 
     def get_filtered_datatypes(self, name, parent_div, tree_session_key, filters):
-        # TODO: fix this use-case
         """
         Given the name from the input tree, the dataType required and a number of
         filters, return the available dataType that satisfy the conditions imposed.
         """
-        algorithm_id = common.get_from_session(common.KEY_ADAPTER)
-        algorithm = self.algorithm_service.get_algorithm_by_identifier(algorithm_id)
-        adapter_instance = ABCAdapter.build_adapter(algorithm)
+        name_start_index = datatype_index_path.rfind('.')
+        datatype_index_name = datatype_index_path[name_start_index+1:]
+        index_class = getattr(sys.modules[datatype_index_path[: name_start_index]], datatype_index_name)
+        filters_dict = json.loads(filters)
 
-        project_id = common.get_current_project().id
-        form = adapter_instance.get_form()(project_id=project_id)
-        current_node = getattr(form, name)
-        if current_node is None:
-            raise Exception("Could not find node :" + name)
+        fields = []
+        operations = []
+        values = []
 
-        new_filter = json.loads(filters)
-        chain_filter = FilterChain(fields=new_filter['fields'], operations=new_filter['operations'],
-                                   values=new_filter['values'])
-        current_node.conditions = chain_filter
+        for idx in range(len(filters_dict['fields'])):
+            fields.append(filters_dict['fields'][idx])
+            operations.append(filters_dict['operations'][idx])
+            values.append(filters_dict['values'][idx])
 
-        return {'field': current_node}
+        filter = FilterChain(fields=fields, operations=operations, values=values)
+        project = common.get_current_project()
+        filtered_datatypes = get_filtered_datatypes(project.id, index_class, filter)
+        return filtered_datatypes
 
     def _get_node(self, input_tree, name):
         """
--- a/framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
+++ b/framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
@@ -252,6 +252,9 @@ class FlowController(BaseController):
         self.fill_default_attributes(template_specification, algorithm.displayname)
         return template_specification
 
+    @expose_fragment('form_fields/options_field')
+    @settings
+    @context_selected
     def get_filtered_datatypes(self, datatype_index_path, filters):
         """
         Given the name from the input tree, the dataType required and a number of
@@ -278,11 +281,7 @@ class FlowController(BaseController):
         data_type_gid_attr = DataTypeGidAttr(linked_datatype=REGISTRY.get_datatype_for_index(index_class()))
         select_field = TraitDataTypeSelectField(data_type_gid_attr, form, conditions=filter)
 
-        options_response = ''
-        for option in select_field.options():
-            options_response = options_response + '<option value=' + '"' + option.value + '"' + ' class="form-control">' + option.label + '</option>'
-
-        return options_response
+        return {'options': select_field.options()}
 
     def _get_node(self, input_tree, name):
         """
--- a/framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
+++ b/framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
@@ -252,7 +252,7 @@ class FlowController(BaseController):
         self.fill_default_attributes(template_specification, algorithm.displayname)
         return template_specification
 
-    def get_filtered_datatypes(self, name, parent_div, tree_session_key, filters):
+    def get_filtered_datatypes(self, datatype_index_path, filters):
         """
         Given the name from the input tree, the dataType required and a number of
         filters, return the available dataType that satisfy the conditions imposed.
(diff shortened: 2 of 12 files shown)


framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -305,6 +306,8 @@ def index(self):
                                                           last_request_type=cherrypy.request.method,
                                                           is_first_fragment=True)
 
+        common.add2session(KEY_ADAPTER, '13')
```
liadomide, line 309: This should not be ...

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
```diff
@@ -377,64 +383,40 @@ def accepted__aggregation_functions():
...
+        filter = FilterChain(fields=fields, operations=operations, values=values)
+        project = common.get_current_project()
+
+        form = Form(project_id=project.id, draw_ranges=True)
+        data_type_gid_attr = DataTypeGidAttr(linked_datatype=REGISTRY.get_datatype_for_index(index_class()))
+        select_field = TraitDataTypeSelectField(data_type_gid_attr, form, conditions=filter)
+
+        options_response = ''
+        for option in select_field.options():
+            options_response = options_response + '<option value=' + '"' + option.value + '"' + ' class="form-control">' + option.label + '</option>'
```
liadomide, line 417: This HTML string concatenation is not good to have.
Can't you return a json which is parsed in the JS and converted in HTML code there on the client ?
Or even better use a Jinja template to render the response as HTML

framework_tvb/tvb/interfaces/web/static/js/filters.js
```diff
@@ -124,44 +123,15 @@ function refreshData(parentDivId, divSufix, name, sessionStoredTreeKey, gathered
...
-            } else {
-                displayMessage("Filter could not be applied!" + name, "infoMessage");
-                return;
-            }
-            displayMessage("Filters applied...");
+        type: 'POST',
+        url: "/flow/get_filtered_datatypes/" + name + '/' + datatypeIndex + '/' + $.toJSON(gatheredData),
+        success: function (response) {
+            var select_field = document.getElementById(name);
+            select_field.innerHTML = response;
```
liadomide, line 132: Basically here compose the options (from JSON response), if they can not be returned by a Jinja template

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
```diff
@@ -241,6 +245,8 @@ def default(self, step_key, adapter_key, cancel=False, back_page=None, not_reset
             template_specification[common.KEY_DISPLAY_MENU] = True
             template_specification[common.KEY_BACK_PAGE] = back_page_link
 
+        common.add2session(common.KEY_ADAPTER, adapter_key)
```
liadomide, line 248: Do we still need this in session ?

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
```diff
@@ -377,64 +383,40 @@ def accepted__aggregation_functions():
...
+            fields.append(filters_dict['fields'][idx])
+            operations.append(filters_dict['operations'][idx])
+            values.append(filters_dict['values'][idx])
+
+        filter = FilterChain(fields=fields, operations=operations, values=values)
+        project = common.get_current_project()
+
+        form = Form(project_id=project.id, draw_ranges=True)
+        data_type_gid_attr = DataTypeGidAttr(linked_datatype=REGISTRY.get_datatype_for_index(index_class()))
+        select_field = TraitDataTypeSelectField(data_type_gid_attr, form, conditions=filter)
```
liadomide, line 413: The new logic looks much cleaner that the prev code! Good job!

Not sure if it is really justified here, but could we potentially push some of this logic in a service and leave here in the controller only the part about inputs parsing and link with a jinja/response json part?

liadomide:
> I managed to return only the options from the Python side and replace them in the JS side. However, there is a problem that I still need to solve, and that is the fact that default None or All Options are deleted this way, so I guess I can't just delete all options and put the new ones, I have to keep those default options. Also probably some parts of the code could look better, but I just wanted to make the PR now so I can get a review if the way this went is a good one or not.

None option would be simple.
ALL should also be replaced after filtering, as it will contain a new set (the GUID after filtering) instead of the ones before filtering.

framework_tvb/tvb/interfaces/web/templates/jinja2/form_fields/datatype_select_field.html
```diff
@@ -1,14 +1,28 @@
-<p class="field-data">
+<div class="field-data">
```
liadomide, line 1: why this change from `p` into `div` is necessary ?
it will break consistency with the other fields...

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
VinczeRobert: Yeah, I intented to eliminate that, thanks for the reminder.

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
VinczeRobert: JSON would work while using some template, so should I create a new Jinja template for this?

framework_tvb/tvb/interfaces/web/static/js/filters.js
VinczeRobert: I think I can't have a JSON response without using some template, when I tried to return for example {'options': string_of_options} only the string options appeared in the response.

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
VinczeRobert: No

framework_tvb/tvb/interfaces/web/static/js/filters.js
liadomide: in this case (when the return value is a json) see to have the correct annotation(s) on the controller method also...
there are multiple examples in the code already and working

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
VinczeRobert: I think we could, why not... But in which one? OperationService? Or a new service just for filters?

liadomide: flowservice ?

framework_tvb/tvb/interfaces/web/templates/jinja2/form_fields/datatype_select_field.html
VinczeRobert: Well, when I did that the div added by me didn't want to appear where I wanted and I read on the internet that block elements such as div  should be inside of inline elements such as paragraph (not that it can't be done, it's just not recommented) and changing it to div helped my problem, but now I tried again with <p> and for some reason now it works, so I guess we can leave it as a paragraph.

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
liadomide: yes, either a new jinja template, and then you use @using_template or return a dict and then use annotation @jsonify on the controller method

framework_tvb/tvb/interfaces/web/static/js/filters.js
liadomide: if you choose annotation `@ using_template` in the controller then you expect a HTML string here (as you do right now) to put directly as innerHTML. If you choose  `@ jsonify` in the controller, then in the JS would be something like:
` r = $.parseJSON(r);
    if (r['status'] === 'ok') ...
`

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
VinczeRobert: I don't think we have a Flow Service, so should I create it?

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
VinczeRobert: This part is solved.

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
VinczeRobert: Wasn't it recently removed?

VinczeRobert:
I believe the general flow is okay now. Tomorrow I really hope I can solve the particular issues as well, like for example filters for simulation_name on TimeSeries don't work as they should.

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
liadomide: `FlowService` was recently renamed into `AlgorithmService` ..... by me ;-)

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
liadomide: Then, I suggest you remove this line

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
```diff
@@ -241,69 +245,43 @@ def default(self, step_key, adapter_key, cancel=False, back_page=None, not_reset
...
         self.fill_default_attributes(template_specification, algorithm.displayname)
         return template_specification
 
-    # @expose_fragment("flow/type2component/datatype2select_simple")
-    def getfiltereddatatypes(self, name, parent_div, tree_session_key, filters):
-        # TODO: fix this use-case
+    @expose_fragment('form_fields/options_field')
+    @settings
+    @context_selected
+    def get_filtered_datatypes(self, datatype_index_path, filters):
```
liadomide, line 258: Wouldn't be simpler the code if you would pass the `dt_module` and `dt_class` strings here?

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
liadomide: If I read this again, most of the logic is parsing of strings into filters, so it would be right to stay in Controller.

framework_tvb/tvb/interfaces/web/templates/jinja2/form_fields/datatype_select_field.html
```diff
@@ -1,13 +1,30 @@
-<p class="field-data">
-<select name="{{ field.name }}" id="{{ field.name }}" class="dataset-selector" tabindex="1">
+<script type="text/javascript" src="/static/js/filters.js?{{ currentVersionJS }}"></script>
```
liadomide, line 1: Can't you put the import of `filters.js` on a higher level, e.g. in other jinja template where `datatype_select_field.html` gets included also?
If you put it here, it will be included multiple times, as we might have the fragment `datatype_select_field.html` repeating on the same page.

framework_tvb/tvb/interfaces/web/templates/jinja2/form_fields/datatype_select_field.html
```diff
@@ -1,13 +1,30 @@
...
 {% for option in field.options() %}
     <option value="{{ option.value }}" {{ 'selected' if option.checked }} class="form-control">{{ option.label }}</option>
 {% endfor %}
 </select>
 {% if field.draw_dynamic_conditions_buttons %}
-    <p class="field-adapters" id="{{ field.owner.name }}{{ field.name }}data_select">
-        <input type="button" tabindex='1' value="Add Filter" onclick="addFilter('{{ field.owner.name }}' + '{{ field.name }}' + 'data_select', {{ field.get_dynamic_filters }})"/>
-        <input type="button" tabindex='1' value="Apply Filters" onclick="refreshData('{{ field.owner.name }}', '{{ field.name }}' + 'data_select', '{{ field.name }}', 'defaultTree')"/>
-    </p>
+    <div class="field-adapters" id="{{ field.owner.name }}{{ field.name }}data_select">
```
liadomide, line 11: Here again, I would prefer if you do not change the tag from `<p>` into `<div>` unless there is something really not working, otherwise the current CSS schema might break

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
VinczeRobert: Done.

framework_tvb/tvb/interfaces/web/templates/jinja2/form_fields/datatype_select_field.html
VinczeRobert: Done.

framework_tvb/tvb/core/entities/storage/datatype_dao.py
```diff
@@ -518,6 +518,10 @@ def get_values_of_datatype(self, project_id, datatype_class, filters=None, page_
                         ).filter(DataType.invalid == False
                         ).filter(or_(Operation.fk_launched_in == project_id,
                                      Links.fk_to_project == project_id))
+
+            if hasattr(datatype_class, 'time_series_type'):
```
liadomide, line 522: This if is not sane here.
We should covert the cartesian product with BurstConfig in a left/right outer join on this condition

framework_tvb/tvb/interfaces/web/templates/jinja2/form_fields/datatype_select_field.html
liadomide: You say `done` but still you changed from `<p>` into `<div>` and did not give a justification why this change is necessary.

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
liadomide: You resolved this, but the undesired change seem to be still here...

framework_tvb/tvb/interfaces/web/templates/jinja2/base_template.html
```diff
@@ -47,6 +47,7 @@
         <script type="text/javascript" src="/static/jquery/jquery-ui.js?{{ currentVersionJS }}"></script> <!--JQuery custom imports for the Range, Slider controller -->
         <script type="text/javascript" src="/static/jquery/jquery.json.js"></script>
         <script type="text/javascript" src="/static/js/genericTVB.js?{{ currentVersionJS }}"></script>
+        <script type="text/javascript" src="/static/js/filters.js?{{ currentVersionJS }}"></script>
```
liadomide, line 50: This import place might be too high.
Have you tried to import this script in `genericAdapterFormFields.html` and in `main_burst.html` ?

framework_tvb/tvb/interfaces/web/templates/jinja2/form_fields/datatype_select_field.html
VinczeRobert: Oh yes, I changed two <p>'s actually. The first one (the one from above in the code) can stay <p> but the second one not really... Because it won't include the <div>'s for the hidden form filters inside the second p, but outside them. I checked this on stackoverflow and in the html documentation it says that these kind of things can happen because inline elements are not supposed to contain block elements.

If we really wan't to keep it <p> then I need to find a different way to include the hidden filters new to the user defined ones.
image

framework_tvb/tvb/interfaces/web/templates/jinja2/base_template.html
VinczeRobert: If I put the inclusion in these two + spatial/includedResources.html (for the two stimulus types) then it's allright.

framework_tvb/tvb/core/entities/storage/datatype_dao.py
VinczeRobert: Oh yes, because I changed it in two places actually. The one which is more above in th html file can stay <p> but the second one should be a div, otherwise I the hidden form filters will have to be handled in a different way than the user defined ones (I attached a picture to show what happens when I leave it as a p, basically the way the code is now, when you apply the filters, the effect of the form filters will be ignored). The HTML documentation says that such things (i.e. elements not appearing where you'd expect them) can happen because inline elements are not supposed to contain block elements. 

In conclusion, I either make it a div, or I have to make some changes to the way form filters are handled.

image

framework_tvb/tvb/core/entities/storage/datatype_dao.py
```diff
@@ -512,12 +512,15 @@ def get_values_of_datatype(self, project_id, datatype_class, filters=None, page_
                                        func.max(text('"OPERATION_GROUPS_1".name')),
                                        func.max(DataType.user_tag_1)
                         ).join((Operation, datatype_class.fk_from_operation == Operation.id)
+                        ).outerjoin((BurstConfiguration, datatype_class.fk_from_operation ==
+                                     BurstConfiguration.fk_simulation)
```
liadomide, line 516: How about: 
`.outerjoin(BurstConfiguration, DataType.fk_parent_burst == BurstConfiguration.id)`
?

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -53,6 +53,7 @@
 from tvb.config.init.introspector_registry import IntrospectionRegistry
 from tvb.interfaces.web.controllers.autologging import traced
 from tvb.interfaces.web.controllers.burst.base_controller import BurstBaseController
+from tvb.interfaces.web.controllers.common import KEY_ADAPTER
```
liadomide, line 56: This import must be orphan, thus it can be removed right ?

PR #190, opened 2020-07-09
TVB NEST parallel

lionelkusch:
I don't how to fix the error of my previous Pull Request ( https://github.com/the-virtual-brain/tvb-root/pull/99 ). I create a new one for clarification of the modification.
Sorry for this. 

I refactor the code and the tests for best integration of the monitors in TVB.

### DIFF
--- a/scientific_library/tvb/simulator/interface_co_simulation.py
+++ b/scientific_library/tvb/simulator/interface_co_simulation.py
@@ -1,0 +1,336 @@
+# -*- coding: utf-8 -*-
+#
+#
+#  TheVirtualBrain-Scientific Package. This package holds all simulators, and
+# analysers necessary to run brain-simulations. You can use it stand alone or
+# in conjunction with TheVirtualBrain-Framework Package. See content of the
+# documentation-folder for more details. See also http://www.thevirtualbrain.org
+#
+# (c) 2012-2020, Baycrest Centre for Geriatric Care ("Baycrest") and others
+#
+# This program is free software: you can redistribute it and/or modify it under the
+# terms of the GNU General Public License as published by the Free Software Foundation,
+# either version 3 of the License, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+# PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+# You should have received a copy of the GNU General Public License along with this
+# program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#
+#   CITATION:
+# When using The Virtual Brain for scientific publications, please cite it as follows:
+#
+#   Paula Sanz Leon, Stuart A. Knock, M. Marmaduke Woodman, Lia Domide,
+#   Jochen Mersmann, Anthony R. McIntosh, Viktor Jirsa (2013)
+#       The Virtual Brain: a simulator of primate brain network dynamics.
+#   Frontiers in Neuroinformatics (7:10. doi: 10.3389/fninf.2013.00010)
+#
+#
+
+"""
+Defines a set Interface input and output of TVB.
+
+.. moduleauthor:: Lionel Kusch <lkusch@thevirtualbrain.org>
+
+"""
+from tvb.basic.neotraits.ex import TraitAttributeError
+from tvb.simulator.monitors import Raw, NArray, Float
+from tvb.simulator.history import BaseHistory, Dim, NDArray
+from types import MethodType
+import numpy
+
+
+class ModelWithProxy:
+    """
+    Minimum variables and functions for creating a model with proxy
+    """
+    _id_proxy = []  # identifier of the proxy node
+    _update = False  # if the proxy node have define by the integrator
+    _proxy_value = None  # the value of the proxy node for current computation
+
+    def copy_inst(self, model_without_proxy):
+        """
+        Copy the value of an instance without proxy
+        :param model_without_proxy: model without proxy
+        """
+        for key, value in vars(model_without_proxy).items():
+            try:
+                setattr(self, key, value)
+            except TraitAttributeError:
+                # variable final don't need to copy
+                pass
+        self.configure()
+
+    def set_id_proxy(self, id_proxy):
+        """
+        define the list of the different proxy
+        :param id_proxy: list with the identifier of the proxy node
+        """
+        self._id_proxy = id_proxy
+
+    def update_proxy(self, data):
+        """
+        the new data for the next computation
+        :param data: valeur for the proxy node
+        """
+        self._update = True
+        self._proxy_value = data
+
+    def coupling_variable(self):
+        """
+        function in order to have access to the variable of the  coupling variable
+        :return: firing rate of the model
+        """
+        return self._coupling_variable
+
+
+class HistoryProxy(BaseHistory):
+    """History implementation for saving proxy data."""
+
+    nb_proxy = Dim()  # number of proxy
+    dim_1 = Dim()  # dimension one
+    dt = NDArray((dim_1, dim_1), 'f')  # integration time of the simulator
+    current_step = NDArray((dim_1, dim_1), 'f', read_only=False)  # the current time step of the simulation
+    id_proxy = NDArray((nb_proxy,), 'i')  # the identification of proxy node
+    buffer = NDArray(('n_time', 'n_cvar', 'n_node', 'n_mode'), 'f8',
+                     read_only=False)  # the buffer of value of the proxy
+
+    def __init__(self, time_synchronize, dt, id_proxy, cvars, n_mode):
+        """
+        initialisation of the history for saving proxy value
+        :param time_synchronize: time between two receiving value
+        :param dt: time of integration
+        :param id_proxy: list of the proxy node
+        :param cvars: the number of coupling variable
+        :param n_mode: the number of mode
+        """
+        # TODO send warning for bad size delay
+        dim = max(1, (len(id_proxy)))
+        size_history = numpy.zeros((dim, dim), dtype=numpy.int)  # size of number of node
+        size_history[0, 0] = int(time_synchronize / dt) + 1  # size of the number of saving step
+        super(HistoryProxy, self).__init__(None, size_history, cvars, n_mode)
+        self.dim_1 = 1
+        self.dt = numpy.array([dt])
+        self.nb_proxy = id_proxy.shape[0]
+        self.id_proxy = id_proxy
+
+    def update(self, step, data):
+        """
+        update the history with new value
+        :param step: the step of the update value
+        :param data: the data for the proxy node
+        """
+        if self.id_proxy.size != 0:
+            step_n = data[0] / self.dt[0] - step  # the index of the buffer
+            if any(step_n > self.n_time):  # check if there are not too much data
+                raise Exception('ERROR too early')
+            if any(numpy.rint(step_n).astype(int) < 0.0):  # check if it's not missing value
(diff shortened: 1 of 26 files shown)


lionelkusch:
The name of the new monitors are not the best.
Can you propose me some name?

maedoc:
@lionelkusch I started to read through, but missing the communication part.  Where's the communication with the translators?

lionelkusch:
In this context, there are not translator. There are only proxy node.

If you want to use a translator, the output of the translator need to be a coupling variable for the interface_co_simulation_parallel. and for the interface_co_simulation monitor, the output can is defined by modified model.

maedoc:
Ok, right, so either the full co-simulation-tvb-nest codes should get in here, so we should get some unit tests here just for the monitor components so we can check they are ok even if not doing full integration tests.  What do you think?

lionelkusch:
There are already unitest in the pull request incorporated with the test of TVB. However, the time for these tests are unnecessarily long. I can reduce them if it needs it.

The full co-simulation-tvb-nest code doesn't need to get it for the testing. I can add it, if you want, but I don't see the actual need of it. You can find the new public release version of the proof of concept in this repository : https://github.com/multiscale-cosim/TVB-NEST . The folder nest-elephant-tvb/tvb contains all the file link to TVB.

The only important element missing in this pull request is the MPI wrapper of TVB for the co-simulation. I didn't add it for the moment because TVB doesn't have dependence with MPI and I don't know where to add it (perhaps in tvb-contrib/tvb_contrib/simulator/file or external/MPI or tvb-root/scientific_library/tvb/simulator/script/ ?)

I will propose a notebook with an example to couple Jansen-Rit and WongWang. I will do it in September. r. 

I have questions : 
What is it reasonable time for a test?
Do you want all the project of TVB-Nest in the pool request?
If yes, can you provide me an indication to merge the two projects?
If you want to have the MPI wrapper, can you indicate me in which folder I should add it?

Can you propose me better name for the monitor?
Want you to integrate some part of the monitor in the simulator?
If yes, which monitors do you want to integrate? and which part of it?

dionperd:
Hi, 

I am making some comments on the pull request, in preparation also of tomorrow's meeting.

1. I think that changes to TVB to allow co-simulation should be as minimal as possible, and agnostic to the actual interfaces a user might implement from/to another simulator.

2. Therefore, all we need is to make available the TVB state for communicating it to the interface to another simulator and/or for being modified by the interface from another simulator.

3. Additionally, it is a generally good idea, as indeed in this pull request, to allow for less total communications by communicating every min_delay, instead at every step, and to communicate less data, by knowing the region nodes and state variables that should be communicated/modified.
However, we could leave this as an option to the user, given that TVB doesn't send/receive a big amount of data, and therefore, I don't think we miss a lot if the full state is communicated or modified at EACH TVB time step.

4. Anyway, summing up the above requirements, all we need is 
(a) access to TVB state
(b)(and optionally to its history up to min_delay, 
(c) whereas the target nodes' indices and
(d) state variables' indices 
might also be needed in case we don't communicate the whole state. 
These are the minimal requirements as I see.

5. So, I find a quite nice idea to have a specialized implementation of a Monitor and a History class, although I am not 100% sure that they are necessary.

6. However, the current implementation is logically circular, because it allows a Monitor (which is contained in the Simulator class) to modify everything else, the Simulator itself included, as well as the Model, the History etc. 
Instead, I think that we should start from a CoSimulator class that will (optionally) hold by default a specific Monitor, and History class. So, I think that some refactoring is necessary at this point.
In case of parallel co-simulation, I guess that the CoSimulator should:
a. make the TVB state (and optionally history for delayed communication) available to some MPI process via the dedicated Monitor,
b. allow the modification of the TVB state by the MPI process, at each time step (look point 9 below).
In case of sequential implementation, it could even be simpler, with the external interface directly having access to the TVB state as in the current implementation in [tvb-multiscale-cosim](https://github.com/the-virtual-brain/tvb-root/tree/tvb-multiscale-cosim) branch.

7. Moreover, I am not persuaded that we need a co-simulation specific Model class. The proxy ids and state variables ids can be information that is hold by either the CoSimulation Monitor class, or even directly by the interface class the user will implement. 
And in no case should we keep information on the dynamical state in the Model class (the "_proxy_value" in the current pull request). The Model is not a dynamic entity in TVB. Only the state, the history, the coupling, and the stimulus, + monitors are dynamic in TVB. I think we should stick to these principles.

8. Similarly, I don't think we need any changes to the Integrator schemes.
 Instead, we are already dealing with the problem of dynamical variables that are NOT state variables, but intermediate ones, and about how to update them after and/or before integration of each time step. See the [dynamic-non-state-vars  ](https://github.com/the-virtual-brain/tvb-root/tree/dynamic-non-state-vars) branch for more details on this.
The state variables that get updated by an external interface pose a similar problem, since it is an update that has to happen before and/or after the integration of a time step.

9. I haven't understood fully when/how the TVB state is updated by the interface from another simulator.

In my opinion, we should insist on what I call the "integrity" of the full-brain TVB state, i.e., we shouldn't allow neither that the TVB brain (as represented in its connectivity) has some holes, i.e., nodes that have disappeared or their recorded state is false, because they are modeled by another simulator, nor that there is another part of the brain, not represented in the TVB connectome, i.e., additional nodes, or even a second brain (!), which "couples" to the TVB brain as represented in the TVB connectome.

Therefore, I don't see how we can avoid updating the TVB state from the other simulator's interface at EACH TVB time step.
If not, then all other monitors will get the wrong information, since they are also updated at EACH TVB time step.

This point is rather crucial:

I propose a different status between 
TVB -> another simulator 
and 
another simulator -> TVB
communications.

The TVB->... should be a coupling from some TVB regions, towards some e.g. spiking populations that **have** to reside in some of the EXISTING brain regions of the TVB connectome. Since it is a coupling, indeed it can happen every min_delay time steps, in which case TVB will handle the time delays, and we need to make available the TVB history to the interface.

Instead, the ...-> TVB communication is like an update of the TVB state for specific region nodes (that are modeled in the other simulator), and for specific state variables (that by the way may be different than the coupling variable(s) of the TVB model...), which have to be overwritten by the data coming from the other simulator, 
i.e., if region i is modeled in another simulator, then at each time step t, and before the state gets to the monitors or the history, the target state variable(s) of the region i have to be overwritten in the TVB state vector by the information coming from the other simulator.

The reason for this differentiation between the two directions is that TVB is the only WHOLE Brain simulator I am aware until today, and it computes neuroimaging signals. Moreover, it is a "TVB-centered view" of the problem of co-simulation...

Even in the case that one wants to model only the cortex with TVB and consider other e.g. subcortical regions, as coupling to TVB from the outside, the problem can be easily solved if the TVB connectome is augmented by one or more "dummy" brain regions. Anyway, this should be necessary in most cases for the large-scale connectivity to be considered in a uniform manner.

I haven't fully understood how all this works in the current implementation of this pull request.


In summary, I think that by introducing a CoSimulator class, and with a little bit of refactoring and cleaning up, we could have a nice co-simulation infrastructure with very minimal changes to TVB, and easy to couple with a large variety of different interfaces users might come up with.

maedoc:
@lionelkusch very nice job with all the testing, the PR is majority tests now.

maedoc:
@dionperd these are a lot of comments to drop here the day before a meeting to exchange on this topic, but here are some responses:

> I think that changes to TVB to allow co-simulation should be as minimal as possible, and agnostic to the actual interfaces a user might implement from/to another simulator.

Sure, it could be nice, but in practice these implementations target science deliverables in HBP on the EBRAINS platforms, and if some extra changes in the short term are required to get the proofs of concepts running, so be it.  We are limited on all teams in terms of developer time.

> all we need is to make available the TVB state for communicating it to the interface to another simulator and/or for being modified by the interface from another simulator.

This is too generic a comment, given the complexity of NEST, MPI and TVB interacting, to warrant a response.

> Additionally, it is a generally good idea, as indeed in this pull request, to allow for less total communications by communicating every min_delay, instead at every step, and to communicate less data, by knowing the region nodes and state variables that should be communicated/modified.
However, we could leave this as an option to the user, given that TVB doesn't send/receive a big amount of data, and therefore, I don't think we miss a lot if the full state is communicated or modified at EACH TVB time step.

The HPC clusters involved usually have 100 GBps Infiniband, traffic from TVB will likely be a rounding error compared to NEST. 

> Anyway, summing up the above requirements, all we need is
(a) access to TVB state
(b)(and optionally to its history up to min_delay,
(c) whereas the target nodes' indices and
(d) state variables' indices
might also be needed in case we don't communicate the whole state.
These are the minimal requirements as I see.

This could be a start on a specification that perhaps we could agree on, a few months ago.  This could be a point in convergence of a solution.

> So, I find a quite nice idea to have a specialized implementation of a Monitor and a History class, although I am not 100% sure that they are necessary.

Did you read the implementation? Did you see Lionel's presentations on the design?  Please make your comment on the details instead of imagined necessity.

> In case of sequential implementation, it could even be simpler, with the external interface directly having access to the TVB state as in the current implementation in tvb-multiscale-cosim branch.

Sequential implementation doesn't satisfy requirements of various deliverables, and it doesn't even run robustly as deployed at https://tvb-nest.apps.hbp.eu

> The Model is not a dynamic entity in TVB. Only the state, the history, the coupling, and the stimulus, + monitors are dynamic in TVB. I think we should stick to these principles.

This might actually be a interesting comment if you had made in the appropriate place as a line comment in the code in the PR interface.  Perhaps you can provide a working alternative which corresponds to your proposal?

>  I don't think we need any changes to the Integrator schemes.

I agree with this, but it's again a question of planning the work over time.  It's ok to have both (a) a working tested hack and (b) a plan to fix the hack longer term.

> I haven't understood fully when/how the TVB state is updated by the interface from another simulator.

I don't think anyone really understands the best way to make this work, but the more people we have working on the same implementation, the easier it will be to find bugs and fix them.  Lionel's PR is more tests than code, which is a great start on this.

> we should insist on what I call the "integrity" of the full-brain TVB state

This really just an opinion on how to build a model.  We don't have to agree on that opinion, but we should agree on the implementation details here, since they have implications for our work going forward.

> Moreover, it is a "TVB-centered view" of the problem of co-simulation..

But not everyone is TVB centered, some are NEST centered and see TVB as just a another thing that speaks MPI.

We'll soon speak on the VC, but since you brought your comments here to the PR, it would be appropriate to make your comments inline on the PR changes, per conventional code review process.

lionelkusch:
For helping to understand the hack of the monitor, I attach a document who explain what I did. 
[PR_190_v2.pdf](https://github.com/the-virtual-brain/tvb-root/files/5175516/PR_190_v2.pdf)

lionelkusch:
> Additionally, it is a generally good idea, as indeed in this pull request, to allow for less total communications by communicating every min_delay, instead at every step, and to communicate less data, by knowing the region nodes and state variables that should be communicated/modified.
The 2 solutions that I propose can be step-by-step but I didn't test it. 
However, doing step-by-step integration kills the performance of simulators on clusters or the ones which use the parallelization because it obligates the simulator to synchronize every time step which very often not desirable behaviour.
In particular for Nest, the step-by-step is not recommended. ( See this paper for more information : https://link.springer.com/content/pdf/10.1007%2F978-3-540-74466-5_71.pdf )  
However, doing step by step integration kills the performance of simulator on cluster or the ones which use the parallelization because it obligates the simulator to synchronize every at time step which very often not the desirable behaviour.
In particular for Nest, the step-by-step is not recommended. ( see this paper for more information : https://link.springer.com/content/pdf/10.1007%2F978-3-540-74466-5_71.pdf )

maedoc:
@lionelkusch should this be closed?

lionelkusch:
yes. You can close it.

maedoc:
thanks @lionelkusch

PR #194, opened 2020-07-14
TVB-2661: Happy flow Import/Export of Project

adrian-dordea:
Export and Import Project for tvb-framework version 2.*
Working:
- BurstConfigurations
- FK between DataType and BurstConfiguration: `DataType.fk_parent_burst` preserved after export&import
- Project Figures (without op_gid reference restored)
- Import with Operation.xml or with ViewModelH5

To be done in a separate PR, the restore of:
- Links between projects
- DataTypeGroups
- creation of a VMH5 from a deprecated Operation.xml


liadomide:
This looks decent. What is blocking you?

framework_tvb/tvb/core/services/import_service.py
```diff
@@ -314,36 +293,112 @@ def _store_imported_images(self, project):
...
+                                                                                 datatype_group)
+                self._store_imported_datatypes_in_db(project, operation_datatypes, dt_burst_mappings,
+                                                     burst_ids_mapping)
+                imported_operations.append(operation_entity)
+            else:
+                view_model = self._get_main_view_model(vm_paths)
+
+                if view_model:
+                    start_date = datetime.now()
+                    alg = SimulatorAdapter().stored_adapter
```
popaula937, line 332: This should be generalized to work on any adapter type. We do not have that info in the ViewModelH5, we could add it or we could have a property on ViewModel which couples it to the proper adapter.

framework_tvb/tvb/core/services/import_service.py
```diff
@@ -314,36 +293,112 @@ def _store_imported_images(self, project):
...
+    def _get_view_model_and_datatypes_paths(import_path):
+        vm_paths = []
+        dt_paths = []
+        for root, _, files in os.walk(import_path):
+            for file in files:
+                if file.endswith(".h5"):
+                    vm_file_oath = root + "\\" + file
+                    h5_class = H5File.h5_class_from_file(vm_file_oath)
+                    if h5_class is ViewModelH5:
+                        vm_paths.append(vm_file_oath)
```
popaula937, line 370: We could check here whether the current ViewModel is the main one and return only that path. We do not use the other paths anywhere else anyway.

framework_tvb/tvb/core/services/import_service.py
```diff
@@ -314,36 +293,112 @@ def _store_imported_images(self, project):
...
+                        dt_paths.append(vm_file_oath)
+        return vm_paths, dt_paths
+
+    @staticmethod
+    def _get_main_view_model(view_model_paths):
+        if view_model_paths:
+            for main_vm_file_path in view_model_paths:
+                view_model = h5.load_view_model_from_file(main_vm_file_path)
+                if hasattr(view_model, "is_main") and view_model.is_main == True:
+                    return view_model
```
popaula937, line 381: If you apply the suggestion above, this check would move inside `_get_view_model_and_datatypes_paths` and here we would load directly the main view model, not all of them.

framework_tvb/tvb/core/services/import_service.py
```diff
@@ -314,36 +293,112 @@ def _store_imported_images(self, project):
...
-                                                                             datatype_group)
-            self._store_imported_datatypes_in_db(project, operation_datatypes, dt_burst_mappings, burst_ids_mapping)
-            imported_operations.append(operation_entity)
+    @staticmethod
+    def _get_operation_paths(paths):
+        return [f.path for f in os.scandir(paths) if f.is_dir()]
 
-        return imported_operations
+    def get_new_operation_for_view_model(self, project, view_model, alg_id):
+        op_param = '{"gid": "' + str(view_model.gid) + '"}'
```
popaula937, line 356: We have similar constructions in many places in the code. Would be nice if you would put it in a commonplace to avoid future maintenance issues.

popaula937:
Discussed with Adi all these review comments. He will address them in future commits. I don't think there is a blocker, but he still needs to work a bit on generalization so that import works on any adapter.

framework_tvb/tvb/core/entities/file/simulator/view_model.py
```diff
@@ -318,6 +318,22 @@ class SimulatorAdapterModel(ViewModel, Simulator):
     def linked_has_traits(self):
         return Simulator
 
+    #TODO: add get_algorithm_module and get_algorithm_class_name to all modules (for project import)
+
+    @property
+    def get_algorithm_module(self):
```
bvalean, line 324: Since this function represents is actually an object property, I would rename it to algorithm_module. Same thing for 'get_algorithm_class_name'

framework_tvb/tvb/core/services/import_service.py
liadomide: I agree that this should be generalized, but the current 2 properties in each of the ViewModels seems like not the best solution
Could we build an in memory registry with Adapter - ViewModel, when we introspect at TVB startup all the adapters ? I think that is done in `tvb.config.init.initializer. _populate_algorithms method`

framework_tvb/tvb/config/init/initializer.py
```diff
@@ -188,6 +188,10 @@ def _populate_algorithms(self, algo_category_class, algo_category_id):
                     stored_adapter.id = inst_from_db.id
 
                 stored_adapter = dao.store_entity(stored_adapter, inst_from_db is not None)
+
+                model = adapter_form().get_view_model()()
```
popaula937, line 192: `get_view_model` is a static method. You don't need to create instances for `adapter_form`, nor the returned view_model. This would cause problems if the `adapter_form` required some arguments (eg: `LocalConnectivityCreatorForm`)

framework_tvb/tvb/core/services/import_service.py
```diff
@@ -38,19 +39,24 @@
...
 from cgi import FieldStorage
+from collections import OrderedDict
 from datetime import datetime
 from cherrypy._cpreqbody import Part
 from sqlalchemy.orm.attributes import manager_of_class
 from sqlalchemy.exc import IntegrityError, SQLAlchemyError
 from tvb.basic.profile import TvbProfile
 from tvb.basic.logger.builder import get_logger
 from tvb.config.algorithm_categories import UploadAlgorithmCategoryConfig
+from tvb.config.init.introspector_registry import IntrospectionRegistry
```
popaula937, line 50: A circular dependency is introduced here, thus tvb_importer won't be part of introspection and the import of Default_Project at startup won't work.

liadomide: Correct. My suggestion to put the new dictionary on the `IntrospectionRegistry` does not work due to dependency chain.
But also the recent fix to put it on FilesHelper does not make much sense, from a logical perspective does not belong there, in my opinion.
How about putting it under tvb.config (maybe directly in init.py, as an empty dict first, then to be filled at introspection)?

framework_tvb/tvb/core/services/import_service.py
```diff
@@ -318,30 +268,110 @@ def import_project_operations(self, project, import_path, dt_burst_mappings=None
...
-                                                                             datatype_group)
-            self._store_imported_datatypes_in_db(project, operation_datatypes, dt_burst_mappings, burst_ids_mapping)
-            imported_operations.append(operation_entity)
+    @staticmethod
+    def _get_operation_paths(paths):
+        path_list = [f.path for f in os.scandir(paths) if f.is_dir()]
+        sorted_dir = {}
+        for path in path_list:
+            op_number = os.path.basename(path)
+            sorted_dir[int(op_number)] = path
```
liadomide, line 328: This convert to `int` produces problems when importing links. We should order ops differently. Maybe by `ViewModel.create_date` field (similar with the previous process

framework_tvb/tvb/core/services/import_service.py
liadomide: I agree. Probably in the constructor of the Operation class itself, instead of params string to pass the `view_model` or `view_model.gid`?

framework_tvb/tvb/core/services/import_service.py
liadomide: I removed this part, but the code now has a big TODO in line 229, as the ordering of the operations is missing.
Instead I added field `create_data` in every ViewModelH5 file. We should read that and put it on the Operation.create_data before we save it. Then we could order Operations by the create_date (either read from `Operation.xml` or from `ViewModelH5`) This should give a correct order, I think.

framework_tvb/tvb/core/services/import_service.py
```diff
@@ -219,80 +219,102 @@ def _store_imported_images(self, project, temp_project_path, project_name):
...
-                self.logger.debug("Importing operation " + str(operation))
-                operation_entity, datatype_group = self.__import_operation(operation)
-                new_op_folder = self.files_helper.get_project_folder(project, str(operation_entity.id))
-                # TODO ViewModel H5 should be created, as we want to preserve only the new structure in the future
-                operation_datatypes = self._load_datatypes_from_operation_folder(root, operation_entity,
-                                                                                 datatype_group, new_op_folder)
-                self._store_imported_datatypes_in_db(project, operation_datatypes)
-                imported_operations.append(operation_entity)
-
+                directory_list[operation.create_date] = root
```
liadomide, line 228: How about when 2 operations are created at the same time ?
Theoretically, are small chances, but it could happen in case of PSE, for example .

liadomide:
I think this PR has become already big, and is fixing and refactoring already many things compared with master.
Thus, I suggest to merge it soon, unless something critical is identified as broken, and afterwards we can continue with the remaining TODOs in another PR (from the same branch).

PR #196, opened 2020-07-15
Mistake in parameter labelling W-C Model

Spokhim:
The variable name is not the same with the label (which I assume is what appears in the TVB GUI).
Change variable name to suit label and also align with the differential equations.
Checked with pg10 of Wilson Cowan 1972.


liadomide (approved):

maedoc (approved):
Good catch. I checked the paper and the equations. This fix is correct.

maedoc:
The fix will invalidate results already produced with this model unfortunately:  I would suggest making a note in the release notes and/or docstring so the someone trying to reproduce the results will have a hint.

liadomide:
> The fix will invalidate results already produced with this model unfortunately: I would suggest making a note in the release notes and/or docstring so the someone trying to reproduce the results will have a hint.

I have just created a bug in Jira with this. We generate the release notes from there, so we will have it included.

PR #198, opened 2020-07-15
HBP-87 Move sensors type to library. Create an enum

bvalean:
- From tvb-contrib move the Sensors enum higher into tvb-library, as it will be useful for other cases also.
- for sensors keep only "Internal" type, and remove the duplicate SEEG. We might need to migrate previous data files (if any exist). We should be consistent

### DIFF
--- a/framework_tvb/tvb/adapters/simulator/monitor_forms.py
+++ b/framework_tvb/tvb/adapters/simulator/monitor_forms.py
@@ -157,7 +157,7 @@ class EEGMonitorForm(ProjectionMonitorForm):
                                     values=[SensorTypes.TYPE_EEG.value])
 
         projection_filter = FilterChain(fields=[FilterChain.datatype + '.projection_type'], operations=["=="],
-                                        values=[ProjectionsType.EEG_POLYMORPHIC_IDENTITY.value])
+                                        values=[ProjectionsType.EEG.value])
 
         self.projection = TraitDataTypeSelectField(EEGViewModel.projection, self, name='projection',
                                                    conditions=projection_filter)
@@ -172,10 +172,10 @@ class MEGMonitorForm(ProjectionMonitorForm):
         super(MEGMonitorForm, self).__init__(variables_of_interest_indexes, prefix, project_id)
 
         sensor_filter = FilterChain(fields=[FilterChain.datatype + '.sensors_type'], operations=["=="],
-                                    values=[SensorTypes.TYPE_MEG.value)
+                                    values=[SensorTypes.TYPE_MEG.value])
 
         projection_filter = FilterChain(fields=[FilterChain.datatype + '.projection_type'], operations=["=="],
-                                        values=[ProjectionsType.MEG_POLYMORPHIC_IDENTITY.value])
+                                        values=[ProjectionsType.MEG.value])
 
         self.projection = TraitDataTypeSelectField(MEGViewModel.projection, self, name='projection',
                                                    conditions=projection_filter)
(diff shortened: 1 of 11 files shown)


framework_tvb/tvb/adapters/simulator/monitor_forms.py
```diff
@@ -192,10 +188,10 @@ def __init__(self, variables_of_interest_indexes={}, prefix='', project_id=None)
         super(iEEGMonitorForm, self).__init__(variables_of_interest_indexes, prefix, project_id)
 
         sensor_filter = FilterChain(fields=[FilterChain.datatype + '.sensors_type'], operations=["=="],
-                                    values=[SEEG_S])
+                                    values=[SensorTypes.TYPE_SEEG.value])
```
popaula937, line 191: Shouldn't this be `TYPE_INTERNAL` instead `TYPE_SEEG`?

scientific_library/tvb/datatypes/sensors.py
```diff
@@ -38,13 +38,18 @@
...
 
-EEG_POLYMORPHIC_IDENTITY = "EEG"
-MEG_POLYMORPHIC_IDENTITY = "MEG"
-INTERNAL_POLYMORPHIC_IDENTITY = "Internal"
+
+class SensorTypes(Enum):
+    TYPE_EEG = "EEG"
+    TYPE_MEG = "MEG"
+    TYPE_INTERNAL = "Internal"
+    TYPE_SEEG = "SEEG"
```
popaula937, line 52: 'Internal' and 'SEEG' terms are referring to the same type of sensors. I suppose you have put both here because Denis uses the 'SEEG' term?

scientific_library/tvb/datatypes/projections.py
```diff
@@ -32,14 +32,17 @@
...
 from tvb.basic.neotraits.api import HasTraits, Attr, NArray
 
-EEG_POLYMORPHIC_IDENTITY = "projEEG"
-MEG_POLYMORPHIC_IDENTITY = "projMEG"
-SEEG_POLYMORPHIC_IDENTITY = "projSEEG"
+
+class ProjectionsType(Enum):
+    EEG_POLYMORPHIC_IDENTITY = "projEEG"
+    MEG_POLYMORPHIC_IDENTITY = "projMEG"
+    SEEG_POLYMORPHIC_IDENTITY = "projSEEG"
```
liadomide, line 45: Maybe remove the suffix `_POLYMORPHIC_IDENTITY` from inside the ENUM item names ?

scientific_library/tvb/datatypes/sensors.py
liadomide, line 52: `SEEG` and `Internal` are synonyms. We should only keep one of these. I propose `TYPE_INTERNAL`

liadomide:
@maedoc @maedoc-amu @dionperd These changes are mostly for cosmetics (of the code) right now, but by introducing ENUM instead of string constants we should improve code stability for the future (in case of future changes, renames, additions, etc).
Your acknowledge and feedback would be nice to have.

maedoc:
> by introducing ENUM instead of string constants

Yes absolutely, it would be good to do this everywhere.

scientific_library/tvb/datatypes/projections.py
```diff
@@ -104,7 +107,7 @@ class ProjectionSurfaceEEG(ProjectionMatrix):
     Specific projection, from a CorticalSurface to EEG sensors.
     """
 
-    projection_type = Attr(field_type=str, default=EEG_POLYMORPHIC_IDENTITY)
+    projection_type = Attr(field_type=str, default=ProjectionsType.EEG.value)
```
maedoc, line 110: why not `field_type=ProjectionsType, default=ProjectionsType.EEG`? The actual value of an enum is a implementation detail.

or with neotraits support, perhaps more concisely,
```python
projection_type = EnumAttr(default=ProjectionsType.EEG)
```
with internally initializing `field_type=default.__class__`?

maedoc: Pushing this further the enum class could also allow generating instances from a enum member directly, since manually coding that is error-prone.

liadomide:
> The changes look good to me. I would prefer to avoid having `.value` everywhere

I understand and agree. 
Unless something else is drastic, I propose we merge this as it is and try to add the `EnumAttr` as a separate followup task & PR.

PR #199, opened 2020-07-15
TVB-2744 Refactor all Matrix viewers to be consistent and faster

liadomide:
- avoid unnecessary H5 loads
- consistent load and inputs parsing


liadomide:
> Looks good from my side. (we also have to fix the top-right form so it will have the current Datatype selected by default)

True. I have prepared a task on this for the next sprint(s): TVB-2752

PR #200, opened 2020-07-17
HBP-86 Use keycloak group instead of the list of GIDs in order to set the admin role for a user.

bvalean:


liadomide:
Based on this first commit, do I understand correctly that you checked and we do not have yet the ROLE TVB-ADMIN in our clients ?
I see you only checked a certain GROUP, while the comment from Marc mentions also a ROLE assigned to members of this GROUP.

bvalean:
On the user info response I get only the group of the logged user, not the roles of that group. 
This is the response:
{
  'sub': 'my_id',
  'email_verified': True, 
  'roles': 
    {
      'group': ['feature:keycloak-bug-workaround', 'group-TVB-admins']
    }, 
  'name': 'Valean Bogdan', 
  'mitreid-sub': 'mitre_id', 
  'preferred_username': 'bvalean', 
  'given_name': 'Valean', 
  'family_name': 'Bogdan', 
  'email': 'bogdan.valean@codemart.ro'
}

I think that if users are assigned to a group, you will get only the group name on the response, not the actual group's roles. For example if you have multiple roles assiged to a certain group you don't have to check for each role, you check only for the group name. I will do some more investigations.

bvalean:
I managed to configure our internal keycloak to receive a similar response, so now we can test locally: tvb-user is a normal user and tvb-admin user is in the TVB-admins group. I will merge this PR and if we will receive more details or if they will change something I will do another PR.

PR #203, opened 2020-07-22
TVB-2750. Avoid incompatibility between Postgres and numpy.bool type.

popaula937:


liadomide:
Great for catching this problem. 
But I think that if we apply `.item()` inside `DataSetMetaData.from_array` we will have fewer changes, and potentially fewer future mistakes.

PR #204, opened 2020-07-28
HBP-106. Run uploaders in a separate process to avoid access to in-memory data

popaula937:


popaula937:
To be merged after framework tests are fixed.

framework_tvb/tvb/tests/framework/core/factory.py
```diff
@@ -255,7 +255,7 @@ def import_region_mapping(user, project, import_file_path, surface_gid, connecti
         view_model.mapping_file = import_file_path
         view_model.surface = surface_gid
         view_model.connectivity = connectivity_gid
-        TestFactory.launch_importer(RegionMappingImporter, view_model, user, project.id)
+        TestFactory.launch_importer(RegionMappingImporter, view_model, user, project)
```
liadomide, line 258: I think in methods like this one is ok to do a `launch_synchronously` because these are used for data preparation.
But in `launch_importer ` (which seems used when we are testing the importers real flow, I think we should strive to preserve the real flow.

liadomide: Seeing the code changes again, I have second thoughts about what I said (that we might leave the code not refactored, and continue at a later point). I fear we might gather too much unclean code which will be forgotten.
@popaula937 What do you think? Could we split this in 2 or better do it clean now ?

popaula937: I have the same feeling, especially because of the tests which do not test the actual flow, that's why I did not merge this PR yet.

liadomide:
Should we close this un-merged, considering #214 ?

popaula937:
> Should we close this un-merged, considering #214 ?

Yes, I would close it since I have moved the relevant changes there already.

PR #205, opened 2020-07-29
WIP HBP-95: Enable Upload of Encrypted data files in TVB Web GUI

VinczeRobert:
It's not quite complete, but I thought I'll push what I have so far because as we discussed with Lia today, it got a little more complicated than we expected.

The first thing to note is that the script I wrote treats each file in the same way, so for example we won't unzip a zip and encrypt each file separately,but the whole zip itself.

The decryption is not done yet, because I don't really know what would be the best way to do it in terms of how much we modify the existing code in uploaders. As a result of the previous paragraph, for example, in ZipConnectivityImporter currently the path to the chosen zip is sent to be unzipped, but now we would need to first decrypt the zip file, save it somewhere temporarily, and send that to the unzip function. And we should do this only if we take the file as encrypted i.e. if the user specified the encrypted AES key.

As for why we need both asymmetric and AES encryption (which is symmetric), I think my comment on the task and especially the links added should be pretty good explanations: https://req.thevirtualbrain.org/browse/HBP-95.

Besides any other ideas for improvement that there might exist, I would have three questions:
1) Related to the 3rd paragraph here. The way the code is now, we must
a)  Decrypt the zip file.
b) Save the decrypted zip file.
c) Send it to FilesHelper.

Couldn't we do something so that we skip step B or is it not a problem if we save it?

2) I understood that there will be one public key and one private key which will be constant (currently a new pair generates at each launch, the public key can be downloaded by the user and the private key is kept in session). I also understood that the public key will be available on the website, but where do we keep the private key?

3) In the crypting.py script I wrote, on line 26, there is a parameter modes.CTR(iv). That iv is supposed to be generated randomly and it is supposed to be different at each encryption, I left it a constant now so I can make this PR. I will probably research it a little bit more to see what it does exactly, but it is probably a part of the AES encryption/decryption algorithms. My idea would be to somehow include that parameter in the same .pem file as the encrypted key, but that would complicate the encryption for the user even more. Or we could just leave it a constant. The reason why it shouldn't be a constant is that it is not secure enough, but we don't use AES alone, we use it in combination for RSA, so maybe we could live with that. Any ideas here?


framework_tvb/tvb/core/services/crypto_service.py
```diff
@@ -0,0 +1,65 @@
+import os
```
liadomide, line 1: Missing tvb copyrights header and author

framework_tvb/tvb/adapters/uploaders/crypting/crypting.py
```diff
@@ -0,0 +1,60 @@
+import os
```
liadomide, line 1: This file is missing the tvb copyright header and author.
I think this is the demo on how the user should encrypt files locally before upload into TVB.
I think it needs a bit more text commenting the content (once it is done).
Should we maybe move it under `tvb.interfaces.command.demos.importers` ? Or maybe a notebook under tvb_documentation/demos ?
The public key file should sit very close to this example

framework_tvb/tvb/core/services/crypto_service.py
```diff
@@ -0,0 +1,65 @@
...
+        upload_path = getattr(view_model, trait_upload_field_name)
+        with open(upload_path, 'rb') as f:
+            encrypted_content = f.read()
+
+        # Get the encrypted AES symmetric-key
+        with open(view_model.encrypted_aes_key, 'rb') as f:
+            encrypted_aes_key = f.read()
+
+        # Read the private key
+        with open("../../adapters/uploaders/keys/private_key.pem", "rb") as key_file:
```
liadomide, line 30: I suggest to put the full path towards the Private Key inside .tvb.configuration (and then read it similarly to the way we read `KEYCLOAK_CONFIGURATION`)

framework_tvb/tvb/adapters/uploaders/crypting/crypting.py
VinczeRobert: I added a notebook and placed it there.

framework_tvb/tvb/core/services/crypto_service.py
VinczeRobert: I am moving the code from crypto_service in a new method in ABC Uploader, because there is not much code.

framework_tvb/tvb/core/adapters/abcuploader.py
```diff
@@ -77,8 +94,68 @@ def _prelaunch(self, operation, uid=None, available_disk_space=0, view_model=Non
...
+            return
+
+        if TvbProfile.current.UPLOAD_KEY_PATH is None or not os.path.exists(TvbProfile.current.UPLOAD_KEY_PATH):
+            raise LaunchException("We can not process Encrypted files at this moment, "
+                                  "due to missing PK for decryption! Please contact the administrator!")
+
+        # Open encrypted file
+        upload_path = getattr(view_model, trait_upload_field_name)
+        with open(upload_path, 'rb') as f:
+            encrypted_content = f.read()
```
liadomide, line 114: Will this full read of the uploaded file be safe in case of very large files (e.g. surfaces) ?
Could we buffer the encryption somehow ?

framework_tvb/tvb/core/services/project_service.py
```diff
@@ -156,6 +156,12 @@ def find_project(self, project_id):
             self.logger.exception("Given Project ID was not found in DB!")
             raise ProjectServiceException(str(excep))
 
+    def find_project_path(self, project_name):
```
liadomide, line 159: I think this new method is never used. I will remove it.

framework_tvb/tvb/interfaces/web/static/js/genericTVB.js
```diff
@@ -877,7 +877,7 @@ function waitForStorageUpdateToEnd() {
  *
  * @param projectId the selected project
  */
-function showDataUploadOverlay(projectId) {
+function showDataUploadOverlay(projectId, projectName) {
```
liadomide, line 880: This new argument doesn't seem to be used. I will remove it.

framework_tvb/tvb/core/services/project_service.py
VinczeRobert: I forgot it there...

framework_tvb/tvb/interfaces/web/static/js/genericTVB.js
VinczeRobert: I forgot it there...

VinczeRobert:
Encrypting and decrypting with buffer is working using pyAesCrypt if we asynchronously encrypt the password (the test that I wrote is passing succesfully). I just have to edit the notebook and it's a little problematic because I can't initialize EncryptionHandler there, but with a fresh mine tomorrow I think it will go fast.

Also, like this it will be much much easier to encrypt multiple files with the same key.

VinczeRobert:
From my part this task is finished now. 

EncryptionHandler can't be used directly because:
1. Both ABCUploader and the Jupyter Notebook gives some directory initialization error.
2. Encrypted files would had been stored with .aes extension, so they would have not been seen when you choose a file to upload (e.g. at ZipConnectivity uploading you can only choose .zip files).

As a result, I use pyAesCrypt directly. It's easy to encrypt more files using the same key from the notebook, the user will just have to insert all the paths in a list. Also, the encrypted files will be stored in the same place as the original ones, having the '_encrypted' extension. 

As for the test part, I used pytest.parametrize which in this case means there are 4 new tests now which do the same thing, but with different file types. I wanted to demonstrate that it works on any file type and with bigger as well (cortex_80k.zip). 

And lastly, I think the buffer size is okay, the files are encrypted and decrypted very fast. I noticed something rather curious. I uploaded encrypted cortex_80k.zip and it took 10 seconds. Then I uploaded the original cortex_80.zip directly and it took 11 seconds. I don't understand why the non-encrypted file would take more time in some cases. The algorithm has to be correct since the tests are passing and in the UI the two files look the same when I visualize them. If someone has an answer to this question, I would be curious to hear it.

framework_tvb/tvb/core/services/encryption_handler.py
```diff
@@ -40,6 +40,7 @@
 import pyAesCrypt
 from tvb.basic.logger.builder import get_logger
 from tvb.basic.profile import TvbProfile
+from tvb.interfaces.command.demos.importers.encrypt_data import generate_random_password
```
liadomide, line 43: This dependency is not correct (`interfaces` package can depend on `core`, nor reverse). 
Maybe put function `generate_random_password` here and use it in the interfaces ?

framework_tvb/tvb/core/adapters/abcuploader.py
```diff
@@ -77,8 +95,55 @@ def _prelaunch(self, operation, uid=None, available_disk_space=0, view_model=Non
...
 
+        trait_upload_field_names = list(self.get_form_class().get_upload_information().keys())
+        for upload_field_name in trait_upload_field_names:
+            self._decrypt_content(view_model, upload_field_name)
+
         return ABCSynchronous._prelaunch(self, operation, uid, available_disk_space, view_model, **kwargs)
 
+    @staticmethod
+    def _decrypt_content(view_model, trait_upload_field_name):
+        if view_model.encrypted_aes_key is None:
```
liadomide, line 106: Maybe move this check before the loop over `trait_upload_field_names ` as it applies to all inputs ?

framework_tvb/tvb/core/adapters/abcuploader.py
```diff
@@ -29,25 +29,40 @@
...
+from tvb.core.adapters.exceptions import LaunchException
 from tvb.core.entities.transient.structure_entities import DataTypeMetaData
-from tvb.core.neotraits.forms import StrField
+from tvb.core.neotraits.forms import StrField, TraitUploadField
 from tvb.core.neotraits.uploader_view_model import UploaderViewModel
+from cryptography.hazmat.primitives import serialization
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives.asymmetric import padding
+from cryptography.hazmat.backends import default_backend
+from tvb.interfaces.command.demos.importers.encrypt_data import ENCRYPTED_DATA_SUFFIX, DECRYPTED_DATA_SUFFIX
```
liadomide, line 52: This dependency from `tvb.core` towards `tvb.interfaces` is not correct (it should be inverse).

framework_tvb/tvb/core/adapters/abcuploader.py
```diff
@@ -77,8 +95,55 @@ def _prelaunch(self, operation, uid=None, available_disk_space=0, view_model=Non
...
+        )
+
+        decrypted_password = decrypted_password.decode()
+
+        # Get path to decrypted file
+        decrypted_download_path = upload_path.replace(ENCRYPTED_DATA_SUFFIX, DECRYPTED_DATA_SUFFIX)
+
+        # Use the decrypted password to decrypt the message
+        pyAesCrypt.decryptFile(upload_path, decrypted_download_path, decrypted_password,
+                               TvbProfile.current.hpc.CRYPT_BUFFER_SIZE)
```
liadomide, line 144: Could it be worthy to create a dedicated `CryptographySettings` class and the put all the constants like `CRYPT_BUFFER_SIZE, CRYPT_PASS_SIZE, ENCRYPTED_DATA_SUFFIX, DECRYPTED_DATA_SUFFIX` there ?
Or under `tvb.core.adapters.constants` ? (I have a tendency for the second alternative, as it is closer to the usages, but I could be ok with these living under the profile, also.

framework_tvb/tvb/interfaces/command/demos/importers/encrypt_data.py
```diff
@@ -0,0 +1,83 @@
...
+            label=None
+        )
+    )
+
+    return encrypted_symmetric_key
+
+def save_encrypted_password(encrypted_password, path_to_encrypted_password):
+
+    with open(os.path.join(path_to_encrypted_password, ENCRYPTED_PASSWORD_NAME), 'wb') as f:
+        f.write(encrypted_password)
```
liadomide, line 83: Maybe add a  `__main__` to demonstrate usage ?

framework_tvb/tvb/tests/framework/adapters/uploaders/encrypt_decrypt_test.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+class TestEncryptionDecryption(TransactionalTestCase):
+
+    # noinspection PyTypeChecker
+    @pytest.mark.parametrize("dir_name, file_name", [('connectivity', 'connectivity_76.zip'),
+                                                     ('surfaceData', 'cortex_2x120k.zip'),
+                                                     ('projectionMatrix', 'projection_meg_276_surface_16k.npy'),
+                                                     ('h5', 'TimeSeriesRegion.h5')])
+    def test_encrypt_decrypt(self, dir_name, file_name):
```
liadomide, line 57: This runs surprisingly fast on Jenkins 
https://jenkins.codemart.ro/view/TVB/job/TVB%20-%20Tests%20-%20SqLite/869/testReport/adapters.uploaders.encrypt_decrypt_test/TestEncryptDecryptImporter/

Then your strange times for uploader execution (with/without encryption) could be due to other reason (e.g. machine load), as 0.x s for encrypt&decrypt in a ~10s job range would not be something to reasonably benchmark with one single manual test, but would require an automated setup to reliably measure multiple times.

framework_tvb/tvb/core/services/encryption_handler.py
VinczeRobert: Okay, but I had to make it a static method in EncryptionHandler and send the pass_size as parameter instead of using self.pass_size in order to avoid initializing an EncryptionHandler object, which would result in an error.

framework_tvb/tvb/core/adapters/abcuploader.py
VinczeRobert: Okay.

framework_tvb/tvb/core/adapters/abcuploader.py
VinczeRobert: Yeah, I didn't pay attention to the dependencies. I reversed it now.

framework_tvb/tvb/core/adapters/abcuploader.py
VinczeRobert: I can move the constants defined by me, but CRYPT_BUFFER_SIZE AND CRYPT_PASS_SIZE I think should stay under the profile because they use a manager object given as a parameter to the constructor.

framework_tvb/tvb/interfaces/command/demos/importers/encrypt_data.py
VinczeRobert: So basically I should write the same thing as in the notebook in a __main__ here?

framework_tvb/tvb/tests/framework/adapters/uploaders/encrypt_decrypt_test.py
VinczeRobert: Yeah, I know that it runs surprisingly fast...

VinczeRobert:
I did a wrong rebase, so I will close this PR and open another one.

PR #210, opened 2020-08-10
TVB-2661: Export/Import of projects from web GUI

liadomide:
This PR continues #194 by touching remaining TODOs in the Import part, as well as refactoring related code.

Done here:
- On import from the old form with `Operation.xml`, create a `ViewModelH5`
- Links Import

Left for another PR (same task, same branch):
- ResultFigures op_gid reference
- check coverage and potentially add unit-tests
- Review the way `ViewModel.gid` is passed to a new `Operation` constructor. Maybe replace `params` with `view_model_gid` column directly ?

Left for another task:
- DataTypeGroups (export and import)


framework_tvb/tvb/core/entities/file/files_helper.py
```diff
@@ -99,6 +99,13 @@ def get_project_folder(self, project, *sub_folders):
             self.check_created(complete_path)
         return complete_path
 
+    @staticmethod
+    def get_key_by_value(list, value):
```
liadomide, line 103: Why this static method in FilesHelper? it doe not seem to be related to files.
Maybe put it as a function in `tvb.core.utils` ?

PR #212, opened 2020-08-10
TVB-2769 Correct Metadata editing in a DT overlay and ensure DB&H5 persistence

liadomide:
In TVB Web GUI, on a DataType, we can open an overlay to show details.
Some of these details can be edited (e.g. "user tags").
This PR ensures that changing those fields can be done without throwing errors and that the changes are persisted both in DB and in H5 files

### DIFF
--- a/framework_tvb/tvb/core/neotraits/_h5core.py
+++ b/framework_tvb/tvb/core/neotraits/_h5core.py
@@ -84,6 +84,7 @@ class H5File(object):
         self.user_tag_3 = Scalar(Attr(str), self, name='user_tag_3')
         self.user_tag_4 = Scalar(Attr(str), self, name='user_tag_4')
         self.user_tag_5 = Scalar(Attr(str), self, name='user_tag_5')
+        self.operation_tag = Scalar(Attr(str), self, name='operation_tag')
         self.parent_burst = Uuid(Attr(uuid.UUID, required=False), self, name='parent_burst')
         self.visible = Scalar(Attr(bool), self, name='visible')
         self.metadata_cache = None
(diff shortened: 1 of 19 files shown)


framework_tvb/tvb/core/neotraits/_h5core.py
```diff
@@ -273,6 +276,7 @@ def __repr__(self):
 
 class ViewModelH5(H5File):
 
+    # TODO  it will be good to be able to just call with H5File.from_file(h5_path) as f for ViewModelH5 also
```
liadomide, line 279: @popaula937 I would like your opinion if this can be achieved or not. I think it can, then we should create a dedicated task for this TODO

framework_tvb/tvb/core/services/project_service.py
```diff
@@ -720,33 +716,43 @@ def _edit_data(self, datatype, new_data, from_group=False):
             operation = dao.get_operation_by_id(datatype.fk_from_operation)
             operation.user_group = new_group_name
             dao.store_entity(operation)
+            op_folder = self.structure_helper.get_project_folder(operation.project, str(operation.id))
+            vm_gid = json.loads(operation.parameters)['gid']
+            view_model_file = h5.determine_filepath(vm_gid, op_folder)
+            if view_model_file:
+                view_model_class = H5File.determine_type(view_model_file)
+                view_model = view_model_class()
+                with ViewModelH5(view_model_file, view_model) as f:
```
liadomide, line 725: @popaula937 this is one example of usage where fixing the previously mentioned TODO would help

liadomide:
Work is done here, but it should be merged after #210 as some unit-tests are now failing, as they expect the project&ops in the final 2.* form (with a View Model and an Operation.parameters with vm.GID)

PR #213, opened 2020-08-11
TVB-2771: Clear SessionStoredSim on project delete (and not only)

VinczeRobert:
I made some tests and deleting only the session stored simulator won't work, the burst config and the last_loaded_url should be removed as well + is_simulationloaded must be False in case you loaded a simulation and then you switched the project.

Is this okay in ProjectController or would it be better in SimulationController?


VinczeRobert:
I agree with the changes (I didn't know about the _mark_selected method until now).

PR #214, opened 2020-08-11
HBP-106 and TVB-2765. Run uploaders in a separate process and fix tests 

popaula937:
Follow-up on PR #204
- uploaders tests are not transactional and follow the real flow (sync but in diff process)
- uploads done by other tests are sync in the main process

TVB-2765:
- removed ABCAsync and ABCSync
- added AdapterLaunchModeEnum to describe the new 3 launching modes for adapters
- added a way to handle exception that are thrown in the subprocess and we want to display them in the GUI for uploaders

### DIFF
--- a/framework_tvb/tvb/tests/framework/core/factory.py
+++ b/framework_tvb/tvb/tests/framework/core/factory.py
@@ -41,7 +41,6 @@ Project, User, Operation, basic imports (e.g. CFF).
 import os
 import random
 import uuid
-from time import sleep
 
 import tvb_data
 from tvb.adapters.datatypes.db.connectivity import ConnectivityIndex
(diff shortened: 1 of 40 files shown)


framework_tvb/tvb/tests/framework/core/factory.py
```diff
@@ -240,34 +242,37 @@ def import_default_project(admin_user=None):
...
-    def launch_importer(importer_class, view_model, user, project_id):
-        # type: (type, ViewModel, User, int) -> None
+    def launch_importer(importer_class, view_model, user, project, sync=True):
+        # type: (type, ViewModel, User, Project, bool) -> None
         importer = ABCAdapter.build_adapter_from_class(importer_class)
-        OperationService().fire_operation(importer, user, project_id, view_model=view_model)
+        if sync:
+            TestFactory.launch_synchronously(user, project, importer, view_model)
+        else:
+            OperationService().fire_operation(importer, user, project.id, view_model=view_model)
```
liadomide, line 251: Everything looks very good to me in these changes. 
This `if sync` was puzzling, as I didn't understand it at first.

Maybe add a comment here:
- `sync = False` will do the normal flow, with Uploaders running synchronously but in a different process. This branch won't be compatible with usage in subclasses of `TransactionalTestCase` because the upload results won't be available for the unit-test running.
-  `sync = True` for usage in subclasses of `TransactionalTestCase`, as data preparation, for example. Won't test the "real" upload flow, but it is very close to that.

popaula937: ABCAdapter looks clean now after you also reviewed it in #194

PR #215, opened 2020-08-12
TVB-2511: Display Equations in Phase Plane and ?(green doc bubble) for the Model

VinczeRobert:
I have done most of this task, but there are two problems:
1. I don't know how to align the text and the equations with the select field, now they are left-alligned (in the old TVB they were aligned with the select field).
2. I had to change the names of the models get_ui_name_to_model for my solution to work because normally HTML id's should not contain spaces and it was only working for the ones which didn't have a space in their name, so I decided to put underscores instead of spaces. The problem is that the same values are used both for the value of the option fields and for the texts, so the users would see them with underscores too and I guess we don't want that, so there should be a way to not have the same values for display and for the id's.
Bonus question: Is the place where I put matjax.py okay? I didn't know where to put them and one of those methods formally was in tvb.basic.traits.utils, but that path doesn't t exist anymore.


scientific_library/tvb/basic/neotraits/matjax.py
```diff
@@ -0,0 +1,113 @@
+# -*- coding: utf-8 -*-
```
liadomide, line 1: Does this new file really fit under `neotraits` ? 
From a logical perspective it does not seem fitting here to me. Is it ?
Not sure where to put it, but maybe closer to where it is used (in the web interface e.g. `framework_tvb/tvb/interfaces/web/controllers/burst/`) ?

liadomide:
> 2. I had to change the names of the models get_ui_name_to_model for my solution to work because normally HTML id's should not contain spaces and it was only working for the ones which didn't have a space in their name, so I decided to put underscores instead of spaces. The problem is that the same values are used both for the value of the option fields and for the texts, so the users would see them with underscores too and I guess we don't want that, so there should be a way to not have the same values for display and for the id's.

Indeed, we should fix this (ID to be without spaces, but display name to have the previous form).
Again I am not sure what is the best solution in the code, but at some point maybe we can just replace `space` with `_` or viceversa ?

scientific_library/tvb/basic/neotraits/matjax.py
VinczeRobert: Okay, I moved it there.

VinczeRobert:
Maybe the new JS code in burst_dynamic could be nicer (I will think about it), but at least now it works. The only problem that remains is that I can't center the doc text like it was before (actually it is not centered, it is aligned under the select field).

framework_tvb/tvb/adapters/simulator/model_forms.py
```diff
@@ -69,7 +69,7 @@ def get_ui_name_to_model():
         'Epileptor': ModelsEnum.EPILEPTOR.get_class(),
         'Epileptor2D': ModelsEnum.EPILEPTOR_2D.get_class(),
         'Epileptor codim 3': ModelsEnum.EPILEPTOR_CODIM_3.get_class(),
-        'Epileptor codim 3 ultra-slow modulations': ModelsEnum.EPILEPTOR_CODIM_3_SLOW.get_class(),
+        'Epileptor codim 3 ultra-slow_modulations': ModelsEnum.EPILEPTOR_CODIM_3_SLOW.get_class(),
```
liadomide, line 72: Is this a left-over `_` or intentional change ?

framework_tvb/tvb/interfaces/web/controllers/burst/matjax.py
```diff
@@ -0,0 +1,113 @@
...
+#   CITATION:
+# When using The Virtual Brain for scientific publications, please cite it as follows:
+#
+#   Paula Sanz Leon, Stuart A. Knock, M. Marmaduke Woodman, Lia Domide,
+#   Jochen Mersmann, Anthony R. McIntosh, Viktor Jirsa (2013)
+#       The Virtual Brain: a simulator of primate brain network dynamics.
+#   Frontiers in Neuroinformatics (7:10. doi: 10.3389/fninf.2013.00010)
+#
+#
+
```
liadomide, line 30: @VinczeRobert I encourage you to mention yourself as the module author, and to include a short sentence describing the current module at the top.

framework_tvb/tvb/interfaces/web/controllers/burst/matjax.py
```diff
@@ -0,0 +1,113 @@
...
+    except (AttributeError, IndexError):
+        doc = None
+
+    if doc is not None:
+        return format_doc('Documentation is missing. Copy-ed from parent\n' + doc)
+
+    return 'Documentation is missing. '
+
+
+def multiline_math_directives_to_matjax(doc):
```
liadomide, line 74: This function could be renamed `_ multiline_math_directives_to_matjax ` for making it explicit it is not part of the public API of the current module?
Or is it used somewhere?

framework_tvb/tvb/interfaces/web/static/js/bursts_dynamic.js
```diff
@@ -534,4 +534,23 @@ function setEventsOnFormFields(param, div_id) {
...
+    $("#data_model_" + model_id.replace(/ /g,"_")).css("display", "block");
+}
+
+function fillDoc(){
+    var model = $('#model');
+    model.on("change", function(event){
+        $("#model > option").each(function(){
+            model_id = $(this).val();
+            $("#data_model_" + model_id.replace(/ /g,"_")).css("display", "none");
+        })
```
liadomide, line 550: Could this `each` and string processing be replaced by jquery selector: `div[id^="data_model_"]` ?
It should return all `div` elements with `id` property starting with `data_model_`. Is this what you want?

liadomide (approved):
Looks good to me. Few cosmetics/questions I noted only

framework_tvb/tvb/adapters/simulator/model_forms.py
VinczeRobert: Oops, left-over, thanks for noticing it.

framework_tvb/tvb/interfaces/web/controllers/burst/matjax.py
VinczeRobert: No, it is not.

framework_tvb/tvb/interfaces/web/controllers/burst/matjax.py
VinczeRobert: I hesitated mentioning myself because I didn't write this code, I just found part of it in the old tvb (I think that code was written by Mihai Andrei some years ago) and I found the multiline_math_directives_to_matjax method here https://programtalk.com/vs2/?source=python/12878/tvb-library/tvb/basic/traits/util.py (I don't know who wrote this). All I did was put them together and extracting the code from _LeftFragmentAdapter's get_input_tree (old TVB)  in configure_matjax_doc. Maybe I could mention Andrei and myself?

framework_tvb/tvb/interfaces/web/static/js/bursts_dynamic.js
VinczeRobert: Yes, I didn't know about this, but that each can be replaced with a single line: $('div[id^="data_model_"]').css("display", "none");.

PR #216, opened 2020-08-12
TVB-2746: Fix Reload Simulation from web page

adrian-dordea:
In TVB Web GUI page Project -> Operations click or the button "play" in case of an operation.
You should be taken to the Simulator page, and the clicked operation should be loaded in memory as a copy (same as when you would click "copy" in the Simulator History left column).
It should work regardless what Simulation was previously selected on the Simulator page.

### DIFF
--- a/framework_tvb/tvb/core/neocom/h5.py
+++ b/framework_tvb/tvb/core/neocom/h5.py
@@ -232,7 +232,7 @@ def determine_filepath(gid, base_dir):
 
 
 def load_view_model(gid, base_dir):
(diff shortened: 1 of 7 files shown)


adrian-dordea:
> @adrian-dordea
> I have tested multiple scenarios and I think it works fine now, but it would be good if you can give it a try tomorrow with the latest form also.
> The refactor affected Burst/Operation Cancel/Delete flow also (to not have cyclic dependency), so pls test that also.

I retested with the latest version multiple reload scenarios and also burst/operations cancel and delete flows, all seems to work fine.

PR #219, opened 2020-08-19
TVB-2773: Convert Operation.parameters into view_model_gid. Fix TVB-2777 and touch TVB-2774

adrian-dordea:
Now we store an operation parameters serialized on disk (in objects `ViewModel`).
Before we were keeping a JSON in the DB column Operation.parameters.
This is a refactor to adapt the `Operation` table to the new simplified column, and to adapt usages.


framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
```diff
@@ -662,7 +661,7 @@ def store_exploration_section(self, val_range, step, dt_group_guid):
         datatype_group_ob = ProjectService().get_datatypegroup_by_gid(dt_group_guid)
         operation_grp = datatype_group_ob.parent_operation_group
         operation_obj = OperationService.load_operation(datatype_group_ob.fk_from_operation)
-        parameters = json.loads(operation_obj.parameters)
+        parameters = {}
```
liadomide, line 664: Why keep this empty dict ?

framework_tvb/tvb/tests/framework/core/factory.py
```diff
@@ -154,8 +154,8 @@ def create_operation(test_user=None, test_project=None, operation_status=STATUS_
         adapter = ABCAdapter.build_adapter(algorithm)
         view_model = adapter.get_view_model_class()()
         view_model.data_file = "."
-        operation = Operation(test_user.id, test_project.id, algorithm.id,
-                              '{"gid": "' + view_model.gid.hex + '"}', status=operation_status)
+        operation = Operation(None, test_user.id, test_project.id, algorithm.id,
```
liadomide, line 157: Why not link the `view_model.gid` here ?

framework_tvb/tvb/core/services/burst_service.py
```diff
@@ -315,7 +315,7 @@ def prepare_metrics_operation(operation):
         range_values = operation.range_values
         metric_algo = dao.get_algorithm_by_module(MEASURE_METRICS_MODULE, MEASURE_METRICS_CLASS)
 
-        metric_operation = Operation(operation.fk_launched_by, operation.fk_launched_in, metric_algo.id, '',
+        metric_operation = Operation(None, operation.fk_launched_by, operation.fk_launched_in, metric_algo.id,
```
liadomide, line 318: We need a `ViewModel` create here also, right ?

framework_tvb/tvb/core/services/backend_clients/cluster_scheduler_client.py
```diff
@@ -65,9 +65,8 @@ def _run_cluster_job(operation_identifier, user_name_label, adapter_instance):
         """
         # Load operation so we can estimate the execution time
         operation = dao.get_operation_by_id(operation_identifier)
-        kwargs = parse_json_parameters(operation.parameters)
         # kwargs = adapter_instance.prepare_ui_inputs(kwargs)
-        time_estimate = int(adapter_instance.get_execution_time_approximation(**kwargs))
+        time_estimate = int(adapter_instance.get_execution_time_approximation())
```
liadomide, line 69: `ABCAdapter.get_execution_time_approximation` expects a loaded `ViewModel` param for correct evaluation.
We should not pass empty here

framework_tvb/tvb/core/entities/storage/datatype_dao.py
```diff
@@ -245,7 +245,7 @@ def count_datatypes_generated_from(self, datatype_gid):
         count = 0
         try:
             count = self.session.query(DataType).join(Operation
-                                        ).filter(Operation.parameters.ilike('%' + datatype_gid + '%')).count()
+                                        ).filter(Operation.view_model_gid.ilike('%' + datatype_gid + '%')).count()
```
liadomide, line 248: After this change, this counter becomes useless. We were expecting to match a DataType child, now we compare with a ViewModel.gid (it will never return True). We should review all its usages, and try to replace the logic

framework_tvb/tvb/core/entities/storage/operation_dao.py
```diff
@@ -281,7 +281,7 @@ def get_operations_for_datatype(self, datatype_gid, only_relevant=True, only_in_
         """
         try:
             query = self.session.query(Operation).filter(
-                                Operation.parameters.like('%' + datatype_gid + '%')).join(
+                                Operation.view_model_gid.like('%' + datatype_gid + '%')).join(
```
liadomide, line 284: Same here, the query becomes useless with this change.
We should review its upper usages, and try to replace the logic.

liadomide (changes requested):
I see few more places were we should propagate changes/cleanup.
We should not merge unless we fix those also, otherwise those will be forgotten and become very hidden bugs.

framework_tvb/tvb/tests/framework/core/services/project_structure_test.py
```diff
@@ -496,23 +496,21 @@ def _create_operations_with_inputs(self, datatype_group, is_group_parent=False):
...
 
         # groups
         ops_group = dao.get_operations_in_group(datatype_group.fk_from_operation)
         assert 10 == len(ops_group)
-        ops_group[0].parameters = parameters
+        ops_group[0].view_model_gid = datatype_gid
         ops_group[0] = dao.store_entity(ops_group[0])
         ops_group[1].visible = False
-        ops_group[1].parameters = parameters
+        ops_group[1].view_model_gid = datatype_gid
```
liadomide, line 513: This change in unit-test is incorrect. 
It hides the problems I mentioned in my previous review, about the DAO recently touched by this PR changes.

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
adrian-dordea: it is populated below and uses range1 and range2

framework_tvb/tvb/core/services/import_service.py
```diff
@@ -363,16 +366,15 @@ def import_project_operations(self, project, import_path):
...
-    @staticmethod
-    def _get_new_form_view_model(operation):
+    def _get_new_form_view_model(operation, xml_parameters):
         # type (Operation) -> ViewModel
         ad = ABCAdapter.build_adapter(operation.algorithm)
         view_model = ad.get_view_model_class()()
-        if operation.parameters:
-            params = json.loads(operation.parameters)
+        operation.view_model_gid = view_model.gid.hex
+        dao.store_entity(operation)
```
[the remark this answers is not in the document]
adrian-dordea: The operation view_model_gid is updated here, but it is true that the method that contains this is in a for, but that is the ideea to update all operation view_model_gid's, because at initial creation they are all None/Null singe there is no view_model then

framework_tvb/tvb/tests/framework/core/factory.py
adrian-dordea: it was a miss on my side, the gid should have been linked

framework_tvb/tvb/core/services/import_service.py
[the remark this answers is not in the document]
liadomide: I still think the same operation is submitted for storage twice in DB, in an unnecessary manner.

framework_tvb/tvb/core/services/burst_service.py
```diff
@@ -315,7 +316,10 @@ def prepare_metrics_operation(operation):
         range_values = operation.range_values
         metric_algo = dao.get_algorithm_by_module(MEASURE_METRICS_MODULE, MEASURE_METRICS_CLASS)
 
-        metric_operation = Operation(operation.fk_launched_by, operation.fk_launched_in, metric_algo.id, '',
+        adapter = ABCAdapter.build_adapter(metric_algo)
+        view_model = adapter.get_view_model_class()()
```
liadomide, line 320: Is this VM stored on disk also ?

adrian-dordea: committed the save on the second review

framework_tvb/tvb/core/entities/storage/datatype_dao.py
liadomide: After analyzing where these are used, as well as discussing it with @popaula937 the proposal here is the following:
- datatypes overlay: simply remove the TAB "Derived DataTypes". Make sure entire removal: from the HTML Jinja, linked JS code, if any, code in python (down to these DAO calls). Also double check if these are not used elsewhere.
- document in Jira in an explicit comment that we do this removal, as well as describe a potential solution (with a `DataType.from_other_datatypes_gids = Column(String) # JSON with datatypes inputs in the operation.view_model used when producing the current DT`

framework_tvb/tvb/core/services/burst_service.py
```diff
@@ -315,7 +316,10 @@ def prepare_metrics_operation(operation):
         range_values = operation.range_values
         metric_algo = dao.get_algorithm_by_module(MEASURE_METRICS_MODULE, MEASURE_METRICS_CLASS)
 
-        metric_operation = Operation(operation.fk_launched_by, operation.fk_launched_in, metric_algo.id, '',
+        adapter = ABCAdapter.build_adapter(metric_algo)
+        view_model = adapter.load_view_model(operation)
```
liadomide, line 320: Is this `ViewModel` properly stored in an H5 ?

framework_tvb/tvb/core/services/project_service.py
```diff
@@ -570,9 +576,10 @@ def _remove_project_node_files(self, project_id, gid, skip_validation=False):
                     new_op = dao.store_entity(new_op)
                     to_project = self.find_project(links[0].fk_to_project).name
                     full_path = h5.path_for_stored_index(datatype)
-                    self.structure_helper.move_datatype(datatype, to_project, str(new_op.id), full_path)
+                    new_op_folfer = self.structure_helper.move_datatype(datatype, to_project, str(new_op.id), full_path)
                     datatype.fk_from_operation = new_op.id
                     datatype.parent_operation = new_op
+                    h5.store_view_model(view_model, new_op_folfer)
```
liadomide, line 582: This new storage of the VM will work, and mostly do the job.
But we could also just move the original VM H5 file, isn't it ?
I fear that by calling store_vie_model again we might change some attributes (e.g. `create_data`). Could this be true?

framework_tvb/tvb/core/services/burst_service.py
popaula937: I don't think this fix is correct, here you would load the view_model of the simulation, not the view_model of the metrics `TimeseriesMetricsAdapterModel`. I added a commit to store the metrics view_model on HPC and then stage it out and update the db accordingly.

adrian-dordea:
added 2 TODO's on operation_dao for 2 queryes that need fixing, they are not working correctly since the introduction of the view models. New task created for them: TVB-2777

framework_tvb/tvb/core/services/burst_service.py
liadomide: I just saw, that the VM gets populated in `HPCSchedulerClient._handle_metric_results`, so we are all good here.
Still I think the TODO in this method should be considered in a later refactor

liadomide (approved):

PR #220, opened 2020-08-24
TVB-2763: Made analyzers preserve the simulation name in a generic way.

VinczeRobert:


VinczeRobert:
There is one little problem related to this task: If I launch TimeSeries Metrics Analyzers, in the Simulation page there will be a web console error and the right side will keep loading but it will never display the datatypes. I think there could be a particular error with this analyzer, because the other ones don't have it.

liadomide (commented):
This could work, and it is somehow generic, but will only work one iteration (so on TS results, not if we have chains like: run Sim + run An 1  on results + An 2 on results
How about #221 ? Could this be even more generic (I haven't fully tested) ?

liadomide:
> There is one little problem related to this task: If I launch TimeSeries Metrics Analyzers, in the Simulation page there will be a web console error and the right side will keep loading but it will never display the datatypes. I think there could be a particular error with this analyzer, because the other ones don't have it.

This could be a separate bug. Does it happen also on master branch ?

VinczeRobert:
> > There is one little problem related to this task: If I launch TimeSeries Metrics Analyzers, in the Simulation page there will be a web console error and the right side will keep loading but it will never display the datatypes. I think there could be a particular error with this analyzer, because the other ones don't have it.
> 
> This could be a separate bug. Does it happen also on master branch ?

It doesn't happen on master, I guess because the linked analyzer result don't appear on the right side in Simulation Page on master.

liadomide:
> > > There is one little problem related to this task: If I launch TimeSeries Metrics Analyzers, in the Simulation page there will be a web console error and the right side will keep loading but it will never display the datatypes. I think there could be a particular error with this analyzer, because the other ones don't have it.
> > 
> > 
> > This could be a separate bug. Does it happen also on master branch ?
> 
> It doesn't happen on master, I guess because the linked analyzer result don't appear on the right side in Simulation Page on master.

I think it an also be reproduced on master, but you will need to try and display the tree from the Project -> Data Structure page. I suspect it is related to some of the specific attributes of a DataTypeMeasure entity (e.g. `title` or `display_name`).

liadomide:
> I suspect this bug is related to some of the specific attributes of a DataTypeMeasure entity (e.g. `title` or `display_name`).

I confirm it was a problem in the implementation of `DatatypeMeasureIndex.display_name`. 
I just fixed it at this commit #bbc3ae4299147b3bc5fa49215bb08c7e2f9cac6f

VinczeRobert:
The solution from PR #221 is better than this one, because it works on chains as well, not just on one iteration. So I will close this PR.

PR #224, opened 2020-08-26
HBP-87. Avoid need to install flask for tvb-multiscale when using h5 layer

popaula937:
Move get_voxel_region from RegionVolumeMappingH5 a level up on the visualizer that uses it because it needs access to more H5 files. This way we keep the h5 layer more isolated in order to easily use it in other modules, eg tvb-multiscale. 
Having get_voxel_region on RegionVolumeMappingH5 resulted in the need to install flask for tvb-multiscale.

### DIFF
--- a/framework_tvb/tvb/adapters/datatypes/h5/region_mapping_h5.py
+++ b/framework_tvb/tvb/adapters/datatypes/h5/region_mapping_h5.py
@@ -28,11 +28,8 @@
 #
 #
 
-from tvb.core.adapters.arguments_serialisation import preprocess_space_parameters
 from tvb.adapters.datatypes.h5.spectral_h5 import DataTypeMatrixH5
 from tvb.adapters.datatypes.h5.structural_h5 import VolumetricDataMixin
-from tvb.core.entities.load import load_entity_by_gid
-from tvb.core.neocom import h5
 from tvb.core.neotraits.h5 import H5File, DataSet, Reference
 from tvb.datatypes.region_mapping import RegionMapping, RegionVolumeMapping
 
@@ -57,7 +54,6 @@ class RegionMappingH5(H5File):
         return self.array_data.load()[int(start_idx): int(end_idx)].T
 
 
-
 class RegionVolumeMappingH5(VolumetricDataMixin, DataTypeMatrixH5):
 
     def __init__(self, path):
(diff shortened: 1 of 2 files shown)


liadomide (approved):
I fully agree with moving this method at a different level (in visualizer).

It is not clear to me how the flask dependency was being enforced in the prev code. 
Could you pls elaborate (as it might hide another wrong dependency)?

popaula937:
> I fully agree with moving this method at a different level (in visualizer).
> 
> It is not clear to me how the flask dependency was being enforced in the prev code.
> Could you pls elaborate (as it might hide another wrong dependency)?

It is not necessary to install the full flask, only werkzeug. But flask was added in the tvb-multiscale Dockerfile as it brings werkzeug, too.
RegionVolumeMappinhH5 had the following chain of dependencies:  `from tvb.core.entities.load import load_entity_by_gid` -> `from tvb.core.entities.file.files_update_manager import FilesUpdateManager` -> `from tvb.core.entities.file.files_helper import FilesHelper` -> `from werkzeug.utils import secure_filename`

liadomide:
> `from tvb.core.entities.file.files_helper import FilesHelper` -> `from werkzeug.utils import secure_filename`

Thanks for the clarification. Perhaps we should review the code in this `files_helper.py` as part of another task.
This module seems to have become a bit heavy lately, while we have it included deep into our core. 
Maybe we can move some of the function from here closer to their actual usages. 
For example the method `save_temporary_file` where `werkzeug` is used, is needed for `tvb.interfaces.rest.server` only. I do not see a reason to cary its 3rd party lib in the rest of the core, just for the potential that it might be needed in the future by other layers...

popaula937:
> > `from tvb.core.entities.file.files_helper import FilesHelper` -> `from werkzeug.utils import secure_filename`
> 
> Thanks for the clarification. Perhaps we should review the code in this `files_helper.py` as part of another task.
> This module seems to have become a bit heavy lately, while we have it included deep into our core.
> Maybe we can move some of the function from here closer to their actual usages.
> For example the method `save_temporary_file` where `werkzeug` is used, is needed for `tvb.interfaces.rest.server` only. I do not see a reason to cary its 3rd party lib in the rest of the core, just for the potential that it might be needed in the future by other layers...

Added [TVB-2775](https://req.thevirtualbrain.org/browse/TVB-2775) for this.

PR #226, opened 2020-08-28
TVB-2772: Corrected equations in Phase Plane

VinczeRobert:
A fixed the obvious syntax errors and I removed the commas from the end of the equations (I didn't get the meaning of them, somewhere they where put, somewhere they weren't, I don't think they look good but I can put them back if it's needed for some reason). A few things I would like to point out:
1. The Linear and the Zetterberg-Jansen function don't have a documentation in their dfun function, I thought to add them myself from the class doc, cause I noticed that usually the dfun doc is like a shorter version of the class doc. I didn't add them yet, but I could.
2. I think I should still look in the mathjax syntax of the model class variables cause there are some problems there too.
3. I think someone who has expertise in this area should look if the equations are correct. I think there might be errors, I even found one possible mistake myself.  Look at the third equation of the Reduced Wong-Wang With Excitatory.... model. It is can be found here https://www.researchgate.net/publication/245029745_Resting-State_Functional_Connectivity_Emerges_from_Structurally_and_Dynamically_Shaped_Slow_Linear_Fluctuations (equation 6). If you look there, there is an additional  term in the end which we don't have.


scientific_library/tvb/simulator/models/epileptor.py
```diff
@@ -288,8 +288,7 @@ class Epileptor(ModelNumbaDfun):
     cvar = numpy.array([0, 3], dtype=numpy.int32)  # should these not be constant Attr's?
     cvar.setflags(write=False)  # todo review this
 
-    def _numpy_dfun(self, state_variables, coupling, local_coupling=0.0,
-             array=numpy.array, where=numpy.where, concat=numpy.concatenate):
+    def dfun(self, x, c, local_coupling=0.0):
```
liadomide, line 291: In this model (and few others bellow) I see that you propose to remove the `_numpy_dfun ` method.
Indeed, that method is not directly used, but I suggest to keep it, as it is a nice fallback in case the numba dependency does not  exist. 
As for the equations that you are after for this bug, I suggest you move the documentation content from under `_numpy_dfun` on the `dfun` to be consistent to the way we read it, and all the other models.

framework_tvb/tvb/adapters/simulator/model_forms.py
```diff
@@ -64,13 +64,13 @@ def get_ui_name_to_model():
     ui_name_to_model = {
         'Generic 2d Oscillator': ModelsEnum.GENERIC_2D_OSCILLATOR.get_class(),
         'Kuramoto Oscillator': ModelsEnum.KURAMOTO.get_class(),
-        'supHopf': ModelsEnum.KURAMOTO.get_class(),
+        'supHopf': ModelsEnum.SUP_HOPF.get_class(),
```
liadomide, line 67: Good catch

liadomide:
> A fixed the obvious syntax errors and I removed the commas from the end of the equations (I didn't get the meaning of them, somewhere they where put, somewhere they weren't, I don't think they look good but I can put them back if it's needed for some reason). A few things I would like to point out:
> 
> 1. The Linear and the Zetterberg-Jansen function don't have a documentation in their dfun function, I thought to add them myself from the class doc, cause I noticed that usually the dfun doc is like a shorter version of the class doc. I didn't add them yet, but I could.
> 2. I think I should still look in the mathjax syntax of the model class variables cause there are some problems there too.
> 3. I think someone who has expertise in this area should look if the equations are correct. I think there might be errors, I even found one possible mistake myself.  Look at the third equation of the Reduced Wong-Wang With Excitatory.... model. It is can be found here https://www.researchgate.net/publication/245029745_Resting-State_Functional_Connectivity_Emerges_from_Structurally_and_Dynamically_Shaped_Slow_Linear_Fluctuations (equation 6). If you look there, there is an additional  term in the end which we don't have.

For 1 and 2 pls go ahead as you propose above. 
For 3, we should contact the model author. I think @dionperd  is author at the model you mention here

maedoc (approved):
All these changes look fine to me.  I however do not have time to check the content of the LaTeX right now, so it may be good to periodically review those equations.

VinczeRobert:
We can check and discuss about the mathematical corectness of the equation at some point in the future. I will merge these changes now as there are no mathjax errors or other obvious problems.

PR #228, opened 2020-08-31
TVB-2752: fill Top-Right call-out with current loaded ViewModel

adrian-dordea:


liadomide (commented):
This addition looks clean, and if it solves the problem, great!

How about the previous InputTree/Adapter which must still be referred in the code? 
Can you find and clean that code ?

framework_tvb/tvb/interfaces/web/entities/context_selected_adapter.py
```diff
@@ -44,14 +44,25 @@ class SelectedAdapterContext(object):
     """
     
     KEY_CURRENT_ADAPTER_INFO = "currentAdapterInfo"
-    _KEY_INPUT_TREE = "inputList"
     _KEY_CURRENT_STEP = "currentStepCategoryId"
     _KEY_CURRENT_SUBSTEP = "currentAlgoGroupId"
     _KEY_SELECTED_DATA = 'defaultData'
     KEY_PORTLET_CONFIGURATION = 'portletConfig'
-    KEY_TREE_DEFAULT = "defaultTree"
+    KEY_VIEW_MODEL = "viewModel"
```
popaula937, line 51: code related to portletConfig can also be cleaned

VinczeRobert (approved):

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
```diff
@@ -522,13 +509,11 @@ def get_adapter_template(self, project_id, algorithm_id, is_upload=False, back_p
     @handle_error(redirect=True)
     @context_selected
     def reloadoperation(self, operation_id, **_):
-        """Redirect to Operation Input selection page,
-        with input data already selected."""
+        """Redirect to Operation Input selection page, with input data already selected."""
```
liadomide, line 512: This method was no longer working properly in case of Analyzers.
I think I just found the fix. Pls test

framework_tvb/tvb/interfaces/web/entities/context_selected_adapter.py
liadomide: This context has become now very simple (only one key) but I still think it is better to keep it as a "context", and we should reuse the concept for grouping simulation keys also.

liadomide (approved):

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
adrian-dordea: tested and it seems to be working fine :)

PR #229, opened 2020-08-31
HBP-113 Encrypt data at rest

bvalean:
I started to work at a solution for storage encryption and a temporary decrypt (only when user works on a specific project).

TBD: How/where do we store the password used for encryption/decryption.
TODO: Check and make sure that we push folder to sync queue every-time we update/create a new file.


framework_tvb/__init__.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
```
liadomide, line 1: this file should not exist. our python packages start from `tvb`

framework_tvb/tvb/core/data_encryption_handler.py
```diff
@@ -0,0 +1,193 @@
+# -*- coding: utf-8 -*-
```
liadomide, line 1: could this module stay under `services` or in `core.entities.file`?

framework_tvb/tvb/adapters/simulator/simulator_adapter.py
```diff
@@ -338,4 +340,5 @@ def launch(self, view_model):
             result_h5[m_name].close()
         self.log.debug("%s: Adapter simulation finished!!" % str(self))
         results.extend(result_indexes.values())
+        DataEncryptionHandler.push_folder_to_sync(FilesHelper.get_project_folder_from_h5(ts_h5_path))
```
liadomide, line 343: can't we do this in the `abcadaper`, so uniformly for all operations?

framework_tvb/tvb/core/data_encryption_handler.py
```diff
@@ -0,0 +1,193 @@
...
+# Dict used to count the project frequency from the queue
+queue_elements_count = {}
+
+# Dict used to count how the project usage
+users_project_usage = {}
+
+# Dict used to keep projects which are marked for deletion
+marked_for_delete = set()
+
+lock = Lock()
```
liadomide, line 59: these seem like the properties of a singleton.
while staying on the module does effectively the same job, from an oop architecture pov woudn't be nicer to have an actual class ?

framework_tvb/tvb/core/entities/file/files_helper.py
```diff
@@ -43,6 +43,7 @@
 from tvb.core.entities.file.exceptions import FileStructureException
 from tvb.core.entities.file.xml_metadata_handlers import XMLReader, XMLWriter
 from tvb.core.entities.transient.structure_entities import GenericMetaData
+from tvb.core.data_encryption_handler import DataEncryptionHandler
```
liadomide, line 46: i understand the reasons for putting this here, but as part of another task we identified that this `helper` (already a not so nice pattern) has become much too heavy.
so, could you maybe move the renaming and removal of the enc folders in the upper layers (the services where for sire these methods are called)?

framework_tvb/tvb/core/neocom/_h5loader.py
```diff
@@ -161,6 +162,7 @@ def store(self, datatype, fname=None):
             subdt = getattr(datatype, traited_attr.field_name)
             if subdt is not None:  # Because a non required reference may be not populated
                 self.store(subdt)
+        DataEncryptionHandler.push_folder_to_sync(FilesHelper.get_project_folder_from_h5(path))
```
liadomide, line 165: is this call (and few others e.g SimAdapter, _h2core.py, h5.py) really necessary, if you add the triggers in hdf5_storage_manager)?

liadomide (commented):
I haven't yet fully processed everything. 
I have few questions/comments on the structure only, for now.

framework_tvb/tvb/core/data_encryption_handler.py
bvalean: I moved it under services

framework_tvb/tvb/core/entities/file/files_helper.py
```diff
@@ -108,8 +114,6 @@ def rename_project_structure(self, project_name, new_name):
 
             if os.path.exists(new_full_name):
                 raise IOError("Path exists %s " % new_full_name)
-
-            os.rename(path, new_full_name)
```
liadomide, line 112: Is this removal by intention ?
What if used in a deployment where encryption is disabled ?

framework_tvb/tvb/core/entities/file/hdf5_storage_manager.py
```diff
@@ -35,18 +35,21 @@
...
 import numpy as numpy
 import tvb.core.utils as utils
-from datetime import datetime
 from tvb.basic.logger.builder import get_logger
 from tvb.basic.profile import TvbProfile
 from tvb.core.entities.file.exceptions import FileStructureException, MissingDataSetException
 from tvb.core.entities.file.exceptions import IncompatibleFileManagerException, MissingDataFileException
+from tvb.core.entities.file.files_helper import FilesHelper
 from tvb.core.entities.transient.structure_entities import GenericMetaData
+from tvb.core.services.data_encryption_handler import encryption_handler
```
liadomide, line 52: This import might show that `encryption_handler.py` should stay here under `tvb.core.entities.file` (and not under `services` as I wrongly suggested before as a valid alternative).

framework_tvb/tvb/core/services/data_encryption_handler.py
```diff
@@ -0,0 +1,239 @@
...
+.. moduleauthor:: Bogdan Valean <bogdan.valean@codemart.ro>
+"""
+
+import os
+import shutil
+import threading
+from queue import Queue
+from threading import Lock
+
+from syncrypto import Crypto, Syncrypto
```
liadomide, line 40: I think you should install this library in dockerfile, so that unit-tests on Jenkins can run

liadomide (commented):
Thank you for considering the previous comments!
The solution looks good. I still haven't managed to fully analyze the logic under enc_handler.... 

I see as missing some tests (will probably not be easy to write, but extremely useful). 
Also fixing the current tests (I think only a dependency is missing) to validate that the prev logic does not break.
`ResultFigure` is also an entity writing on drive. Maybe a trigger to the synch queue will need to be added for it also?

framework_tvb/tvb/core/entities/file/files_helper.py
bvalean: You are right. This line was accidentally deleted.

framework_tvb/tvb/core/entities/file/data_encryption_handler.py
```diff
@@ -0,0 +1,248 @@
...
+.. moduleauthor:: Bogdan Valean <bogdan.valean@codemart.ro>
+"""
+
+import os
+import shutil
+import threading
+from queue import Queue
+from threading import Lock
+
+from syncrypto import Crypto, Syncrypto
```
liadomide, line 40: This could be a heavy dependency for tvb-framework, which we will use only very seldom.
Could we consider making this import optional, and raise exception, in case the setting to encrypt is set to True, but we can not find `syncrypto` in the current env ? 
Just a suggestion ....

framework_tvb/tvb/core/neotraits/_h5core.py
```diff
@@ -332,7 +333,7 @@ def gather_references_by_uuid(self):
         ret = []
         for accessor in self.iter_accessors():
             if isinstance(accessor, Uuid) and not isinstance(accessor, Reference):
-                if accessor.field_name is 'gid':
+                if accessor.field_name is 'gid' or accessor.field_name is 'parent_burst':
```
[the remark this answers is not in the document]
liadomide: After discussing this with Bogdan, the change is valid, but we should make it more clear in the code by renaming the method or at least adding documentation that only DataTypes are returned here (not all references as it currently seems suggested by the method name).

framework_tvb/tvb/core/services/import_service.py
```diff
@@ -203,6 +204,11 @@ def _import_projects_from_folder(self, temp_folder):
             self.import_project_operations(project, temp_project_path)
             # Import images and move them from temp into target
             self._store_imported_images(project, temp_project_path, project.name)
+            if DataEncryptionHandler.encryption_enabled():
+                project_folder = self.files_helper.get_project_folder(project)
```
popaula937, line 208: Isn't this folder the same as project_path above?

framework_tvb/tvb/core/services/project_service.py
```diff
@@ -308,7 +321,14 @@ def remove_project(self, project_id):
             for burst in project_bursts:
                 dao.remove_entity(burst.__class__, burst.id)
 
+            project_folder = self.structure_helper.get_project_folder(project2delete)
             self.structure_helper.remove_project_structure(project2delete.name)
+            encrypted_path = encryption_handler.compute_encrypted_folder_path(project_folder)
+            if os.path.exists(encrypted_path):
+                if os.path.isdir(encrypted_path):
```
popaula937, line 328: Isn't it always a dir?

popaula937 (commented):
Still have to go through a few files, but here are my first questions/comments.

framework_tvb/tvb/core/neocom/h5.py
```diff
@@ -34,6 +34,7 @@
 from datetime import datetime
 
 from tvb.basic.neotraits.api import HasTraits
+from tvb.core.entities.file.data_encryption_handler import encryption_handler
```
popaula937, line 37: This import is not used anymore

framework_tvb/tvb/core/neotraits/_h5core.py
```diff
@@ -37,9 +37,12 @@
 import numpy
 import scipy.sparse
 from tvb.basic.logger.builder import get_logger
-from tvb.basic.neotraits.api import Final, HasTraits, Attr, List, NArray, Range
+from tvb.basic.neotraits.api import Final
+from tvb.basic.neotraits.api import HasTraits, Attr, List, NArray, Range
 from tvb.basic.neotraits.ex import TraitFinalAttributeError
+from tvb.core.entities.file.data_encryption_handler import encryption_handler
 from tvb.core.entities.file.exceptions import MissingDataSetException
+from tvb.core.entities.file.files_helper import FilesHelper
```
popaula937, line 45: Same for these 2 new imports

PR #232, opened 2020-09-02
TVB-2762: Fixed Complex numbers problem in Postgres

VinczeRobert:
Postgres db has no datatype for complex numbers, it can only be achieved through defining a new datatype, which is a little more complicated, it requires some code in C as explained here https://www.postgresql.org/docs/9.4/xtypes.html.

So we thought with Paula that it would be easier just to have strings instead of complex numbers where there were errors (for Covariance and Complex Coherence Analyzers, the others didn't have this problem).

### DIFF
--- a/framework_tvb/tvb/adapters/analyzers/node_covariance_adapter.py
+++ b/framework_tvb/tvb/adapters/analyzers/node_covariance_adapter.py
@@ -167,12 +167,6 @@ class NodeCovarianceAdapter(ABCAdapter):
         covariance_index.fk_source_gid = self.input_time_series_index.gid
         covariance_index.subtype = type(covariance_index).__name__
         covariance_index.array_has_complex = array_metadata.has_complex
-
-        if not covariance_index.array_has_complex:
-            covariance_index.array_data_min = array_metadata.min
-            covariance_index.array_data_max = array_metadata.max
-            covariance_index.array_data_mean = array_metadata.mean
-
         covariance_index.array_is_finite = array_metadata.is_finite
         covariance_index.shape = json.dumps(covariance_h5.array_data.shape)
         covariance_index.ndim = len(covariance_h5.array_data.shape)
--- a/framework_tvb/tvb/adapters/analyzers/node_covariance_adapter.py
+++ b/framework_tvb/tvb/adapters/analyzers/node_covariance_adapter.py
@@ -167,6 +167,12 @@ class NodeCovarianceAdapter(ABCAdapter):
         covariance_index.fk_source_gid = self.input_time_series_index.gid
         covariance_index.subtype = type(covariance_index).__name__
         covariance_index.array_has_complex = array_metadata.has_complex
+
+        if not covariance_index.array_has_complex:
+            covariance_index.array_data_min = float(array_metadata.min)
+            covariance_index.array_data_max = float(array_metadata.max)
+            covariance_index.array_data_mean = float(array_metadata.mean)
+
         covariance_index.array_is_finite = array_metadata.is_finite
         covariance_index.shape = json.dumps(covariance_h5.array_data.shape)
         covariance_index.ndim = len(covariance_h5.array_data.shape)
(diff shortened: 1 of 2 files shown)


liadomide:
I agree that this is a simple solution .... but I fear it might bring problems in the future.

The current proposed code change transforms these columns into `Strings` for all `DataTypeMatrix`, and if we simply use the strings into mathematical expressions they will return something, but not always will be the desired result. 
So, we will need in every usage to make sure that a conversion is done correctly back to numbers ... or something.

Have you considered for the complex DTs to have extra columns (to store both the real and complex parts independently as Float)?
Or simple when flag `DataTypeMatrix.array_has_complex` is True, wipe the values from `array_data_min`, `array_data_max`, etc ?

popaula937:
> I agree that this is a simple solution .... but I fear it might bring problems in the future.
> 
> The current proposed code change transforms these columns into `Strings` for all `DataTypeMatrix`, and if we simply use the strings into mathematical expressions they will return something, but not always will be the desired result.
> So, we will need in every usage to make sure that a conversion is done correctly back to numbers ... or something.
> 
> Have you considered for the complex DTs to have extra columns (to store both the real and complex parts independently as Float)?
> Or simple when flag `DataTypeMatrix.array_has_complex` is True, wipe the values from `array_data_min`, `array_data_max`, etc ?

I am skeptical about changing the type for all of them either, only for the ones that we expect to have complex values. 
I have thought about separate float columns for the real and imaginary parts. It might be a better option since we would move anyway these columns on the subclasses.
Haven't considered wiping the values...maybe we could do a combination of the 2? So, we would still keep min/max values on DataTypeMatrix but wipe them when we have complex numbers, and add min_real/max_real/min_imaginary/max_imaginary only on the Covariance/Complex Coherence?

liadomide:
> I have thought about separate float columns for the real and imaginary parts. It might be a better option since we would move anyway these columns on the subclasses.
> Haven't considered wiping the values...maybe we could do a combination of the 2? So, we would still keep min/max values on DataTypeMatrix but wipe them when we have complex numbers, and add min_real/max_real/min_imaginary/max_imaginary only on the Covariance/Complex Coherence?

I think wiping the current Float columns will be a good short term solution.
We might need the combination as you propose it.  It all depends on the Viewers, and the way they use these values.
I think for now, all the viewers (with one exception that I am aware of) are incapable to display Complex numbers, and we will be better with simply hiding those viewers in case of complex numbers. For complex numbers to be handled in the viewers I think major changes in there will be needed, in some I am not even certain what sense will it make to display the complex part ...

liadomide:
One more thing: if the `dtype=complex` is properly defined in the other datatypes that I mentioned on Jira (`Fourier`, `Wavelet`, etc) why do you say they do not have the same problem? 

I tend to believe that the `HasTraits dtype` is correctly specified, and for certain inputs the array dtype resulting from the analyzer will be complex (even if not always). So we should make this solution generic enough to cover all `DataTypeMatrix` subclasses where  `array_has_complex ` is True.

VinczeRobert:
@liadomide  because I was able to run those analyzers without encountering the same issue.

VinczeRobert:
Okay so we all agreed on this solution, there is only one small issue left that I don't know: in node_covariance_adapter has_complex is False, even though it doesn't affect the latest pushed solution.

I will merge this tomorrow morning if there won't be objections.

liadomide:
> Okay so we all agreed on this solution, there is only one small issue left that I don't know: in node_covariance_adapter has_complex is False, even though it doesn't affect the latest pushed solution.
> 
> I will merge this tomorrow morning if there won't be objections.


I think that the flag `has_complex` was calculates as False because all values in the array were having `+0j`. 
I reverted your last commit, so that our solution considers flag `has_complex` and ensures the empty imaginary part is removed from the columns `min, max, mean`.
Pls re-test and if it works for you do the merge.

liadomide (approved):

PR #234, opened 2020-09-07
TVB-2749: fix burst status for pse

adrian-dordea:


VinczeRobert (approved):

framework_tvb/tvb/core/operation_async_launcher.py
```diff
@@ -75,6 +75,11 @@ def do_operation_launch(operation_id):
         # p = profiler.Profiler("/Users/lia.domide/TVB/profiler/")
         # p.run(OperationService().initiate_prelaunch, curent_operation, adapter_instance, {}, **PARAMS)
 
+        if curent_operation.fk_operation_group:
+            dt_group = dao.get_datatypegroup_by_op_group_id(curent_operation.fk_operation_group)
+            if dt_group.fk_parent_burst:
+                adapter_instance.generic_attributes.parent_burst = dt_group.fk_parent_burst
+
```
liadomide, line 82: This does not seem the ideal place to adjust the `adapter.generic_attributes`. 
Such attributes should already be stored on the `ViewModel.generic_attributes` as the adapter reads these from there in `_prelaunch`.

Have you tried to put something in `SimulatorService.async_launch_and_prepare_pse` like:

```
            ga = self.operation_service._prepare_metadata(simulator_algo.algorithm_category, {},
                                                          operation_group, burst_config.gid)
            session_stored_simulator.generic_attributes = ga

            for param1_value in range_param1.get_range_values():
                  ........
```
The current implementation of `SimulatorService.async_launch_and_prepare_pse`, seems to omit populating generic attributes (unlike `SimulatorService.async_launch_and_prepare_simulation` which does it).

popaula937:
I will merge this because it fixes the PSE run, but we still have one use-case where the status is not kept in sync. I described it on the task and it will be fixed in another PR.

PR #235, opened 2020-09-07
TVB-2758. Ensure values of zero work for float fields

popaula937:
On the base `Field` class, we have a `value` property that does not work correctly when float fields are zero. 
It ended up in returning a string '0' in the `view_stimulus` endpoint, thus the error.
I added an explicit check for this situation, as it is the quickest solution. But I believe this code needs a larger review. Maybe should be done together with the task [TVB-2669](https://req.thevirtualbrain.org/browse/TVB-2669), where we will also clean up the unnecessary `Field` classes.

### DIFF
--- a/framework_tvb/tvb/core/neotraits/forms.py
+++ b/framework_tvb/tvb/core/neotraits/forms.py
@@ -99,6 +99,7 @@ class Field(object):
 
     @property
     def value(self):
+        # TODO: to be reviewed in task TVB-2669 / avoid need for this check and the ones in FloatField.value
         if str(self.data) == self.unvalidated_data:
             return self.data
         return self.data or self.unvalidated_data
@@ -108,15 +109,6 @@ class Field(object):
 
     def __str__(self):
         return jinja_env.get_template(self.template).render(field=self)
-
-
-class SimpleBoolField(Field):
-    template = 'form_fields/bool_field.html'
-
-    def _from_post(self):
-        if self.unvalidated_data is None:
-            self.data = False
-        self.data = bool(self.unvalidated_data)
 
 
 class SimpleStrField(Field):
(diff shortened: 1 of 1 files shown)


VinczeRobert (approved):

framework_tvb/tvb/core/neotraits/forms.py
```diff
@@ -412,6 +394,11 @@ def _from_post(self):
         else:
             self.data = None
 
+    @property
+    def value(self):
+        if self.data == 0:
+            return self.data
+        return super(FloatField, self).value
```
liadomide, line 401: I suppose this is the fix (plus the `SimpleFloatField .value` implementation).

It looks a bit fishy indeed. I conquer that a review would be necessary, to check why and how `value` has ended up as string instead of float. As we have the review task already in mind, I think it is acceptable if we merge these changes now (with a TODO maybe referring to the refactor task number) to avoid the bug, and we come back to it when the refactor is to be done.
It would be great if a unit-test on this mysterious `0` value is added (either now or when the refactoring task will be implemented).

popaula937: The 'how' is as follows: 
- we have `data` and `unvalidated_data` attributes on `Field`
- `unvalidated_data` is what comes from the GUI and gets through validation/transformation
- `data` is the processed value (eg. float value for `FloatField`)
- on `Field.value` we have: `return self.data or self.unvalidated_data`

In the current bug, we have the following values:
- data = 0
- unvalidated_data = '0'

So, `self.data or self.unvalidated_data `evaluates to '0' and this is being returned to the visualizer.

We should review the usage of 'data' and 'unvalidated_data' and better define their roles in order to correct such edge cases. But I think it should be done as a separate task in order to better focus on them.

liadomide: Thanks for documenting how this is happening!
Then to me `self.data or self.unvalidated_data` expression seems inappropriate (and maybe not only for Float).
If the role remains as you suggested (`unvalidated_data` comes from GUI, and `data` is after processing), then I suggest to return on property `value` `self.data` by default (unless otherwise overwritten).
But indeed, this requires a deeper validation and it should be in the separate task.

PR #241, opened 2020-09-11
Improvements to the GPU parameter sweep notebook

maedoc:
This will be a crucial notebook to make look nice, since it would provide a template notebook for many projects relying on parameter sweeps.  I think the CLI is fine.


maedoc:
https://github.com/the-virtual-brain/tvb-root/pull/297 obseletes this PR, so closing

PR #242, opened 2020-09-14
TVB-2745: Fix overriding of tvb folder

adrian-dordea:
This was reported by an external user.
He was trying to use TVB in `~/TVB/TVB_Distribution` folder, and also keeping the default settings for TVB.
This means that folder `~/TVB` is overwritten when TVB starts after initialization, leaving the process in a bad shape (tvb running, but python files partially removed).
With this PR we want to validate on settings page that a user can not select a non empty folder for storing TVB data

### DIFF
--- a/framework_tvb/tvb/core/services/settings_service.py
+++ b/framework_tvb/tvb/core/services/settings_service.py
@@ -33,15 +33,15 @@ Service layer for saving/editing TVB settings.
 .. moduleauthor:: Bogdan Neacsa <bogdan.neacsa@codemart.ro>
 .. moduleauthor:: Lia Domide <lia.domide@codemart.ro>
 """
+
 import os
 import shutil
 import sys
-
 from sqlalchemy import create_engine
 from tvb.basic.config import stored
 from tvb.basic.logger.builder import get_logger
 from tvb.basic.profile import TvbProfile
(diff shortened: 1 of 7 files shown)


framework_tvb/tvb/core/services/settings_service.py
```diff
@@ -237,3 +238,12 @@ def save_settings(self, **data):
...
         return anything_changed, first_run or db_changed
+
+    def check_tvb_folder(self, storage_path):
+        """
+        Check if the storage folder exists and is empty.
+        """
+        if storage_path:
+            if os.path.isdir(storage_path):
+                if os.listdir(storage_path):
+                    raise InvalidStorageException('TVB Storage should be empty, please set another folder.')
```
liadomide, line 249: Shouldn't this exception be raised only in case IS_FIRST_RUN ?
As I see from the unit-tests currently on any settings save this happens. 
We should be allowed to change other TVB settings later on.
Not sure if changing tvb_folder is allowed. If allowed, we should check that the NEW folder selected is empty.

liadomide (changes requested):

liadomide (approved):
Pls test a bit more from web GUI. The code looks good to me now

adrian-dordea:
> Pls test a bit more from web GUI. The code looks good to me now

Checked and tested, all seems to work fine.

PR #244, opened 2020-09-14
TVB-2430: Migration script from old to new storage version

VinczeRobert:
Please ignore the code ugliness of the code in the database migration script for now and the potential duplicate code in the h5 one, I will deal with those soon, but first I wanted to see the whole picture of the idea we discussed last Friday.

In the database migration code I will only migrate the tables that CAN'T BE CREATED FROM H5 (like Users, HasTraitsIndex and a few more).

Then when we reach the H5 migration code you will see that it does it a new thread and it loops over the datatypes in the database. I tried to migrate the HasTraitsIndex and DataTypes tables first, but it gives errors, it doesn't work (I am talking about line 200 in files_update_manager in master). So I thought to get all files with the H5 extension from the storage folder. There might be a better way to do this, but I didn't find one so far.

So I will solve the problems in this order in the following days:
1. Find everything in the database that can't be created from the h5 files and extract/migrate the necessary information. I will write a list with these.
2. Migrate the few H5 files that were left and extract duplicate code in functions. Also write down if there are some H5 files that I can't optain (like TractsH5).
3. Recreate the DB from the H5 files + the extracted information at Step 1.
4. Optimize solution (probably I will get some suggestions on those but I won't wait just for that).
5. Testing.


VinczeRobert:
I decided to first try to implement the whole flow for only a subset of the tables and then just repeat the process for the others.

VinczeRobert:
Since this task has been going on for a while and is approaching it's end, I will explain some of the things I did and then I will mention the problems that still appear in some cases.

The first thing to note is the new method in FilesHelper which collects all the H5 folders in the numerical order of the operation folders (1, 2, 3, etc.). I choose this approach because we want to recreate the database from the H5 files, so we can't use the database to get all datatypes like it was done in the past. Since this change is not necessary for previous migrations, I'll probably make it this way only for migrations from 1.5.8 to 2.0. The numerical order of the operation folders is necessary because we have to migrate the data A which is necessary for data B before data B is migrated (for example we have to migrate a connectivity before a region mapping where the connectivity is used). The alternative I thought about is to put dummy values where the dependent datatype wasn't migrate it yet and then replace it with the correct values later, but I believe that approach would have been even more complicated.

To populate the view model I did a little hack. Many of the attributes that are needed for the view models can be find in the old database, more exactly in the parameters column of the Operations table. Since that column does not exist in the new version, but there is a new one (view_model_gid) I choose to put create the new column named view_model_gid, copy the information from parameters there, then use the information to create the view model and at the end update the field with the gid of the view model. Of course this didn't work for each case, but it worked for most of them, and where it didn't work, I added a dictionary called "changed_values". 

Now I will enumerate the problems that I still have to find a solution to:
1. I couldn't find datatypes to generate a RegionMappingVolumeH5 and TractsH5. 
2. The Users table now has a new column named "display_name". This did not exist before and there is no way I can get it. For now I just put the username value in this column as well.
3. The Stimuli needs the 'weight' attribute and my method populates only the ViewModels not inherited parameters and StimuliRegion and StimuliSurface has this attribute in the superclass. This one I know how to solve though: I will add the parameter directly.
4. The situation with BurstConfiguration is tricky. First I tried to migrate it and then assign the parent_burst attribute whenever it was needed. It seemed like an idea that would work but I realized a little later than I wished that it won't. The reason is that in the new version BurstConfig is included in HasTraitsIndex, so it can't have any id that there is in HasTraitsIndex. Before BurstConfigs were separate, meaning that you could have a BurstConfig with id = 1 let's say and also have another datatype with id = 1 in HasTraitsIndex. I realized this when I wanted to retrieve a BurstConfig from the database with the id = 1 and it got me a surface, cause that had the id = 1 in HasTraitsIndex. I tried to recreate the Burst, but this method also has some problems. You will lose the start_time and finish_time attributes, the simulation name and possibly some others as well. I implemented part of this second scenario but I stopped and ignored it for now so we can discuss it. By the way, for importing separate H5 files I think we don't need to create BurstConfigs, right? I imported a TimeSeriesH5 in the current version and no BurstConfig was created.
5. Uploading StructuralMRI H5 is tricky as well. This is because it depends on data that is created in the same folder: the Volume. Pretty much every data that depends on some other data is migrated in a previous folder, but this is an exception. I thought about creating a dummy Volume and then populate it with the actual values later. StructuralMRI H5 is not the only place where this problem appears, I believe I have seen it in two more places.
6. By recreating the datatypes in the db. the real create_date will be lost. Do we want to keep the actual create dates? If yes, then the Datatype table will have to be migrated which is annoying because you have to migrate each type and module separately. Another option would be to do some trick and store the real create_date somewhere and then transfer it later, like I did with the parameters in Operations table.

These are the problems that come to my mind right now, I will edit this list if I can remember more or find new ones. I will start solving them and then reviewing each file / datatype table again.

framework_tvb/tvb/core/entities/file/file_update_scripts/005_update_files.py
```diff
@@ -32,17 +32,669 @@
...
+        f.load_into(datatype)
+        generic_attributes = f.load_generic_attributes()
+        datatype_index = REGISTRY.get_index_for_datatype(datatype.__class__)()
+        datatype_index.create_date = root_metadata['create_date']
+
+        # Get the dependent datatypes
+        for attr_name, attr_value in dependent_attributes.items():
+            dependent_datatype_index = dao.get_datatype_by_gid(attr_value.replace('-', '').replace('urn:uuid:', ''))
+            dependent_datatype = h5.load_from_index(dependent_datatype_index)
+            setattr(datatype, attr_name, dependent_datatype)
```
popaula937, line 647: take a look at `h5.load_with_references` to reuse here. It loads the whole datatype from a h5 file together with its dependencies. So the code above would turn into one line that loads the datatype and then the lines that prepare the index.

framework_tvb/tvb/core/entities/file/file_update_scripts/005_update_files.py
```diff
@@ -32,17 +32,669 @@
...
+            operation_xml_parameters = eval(operation_xml_parameters)
+        except NameError:
+            operation_xml_parameters = operation_xml_parameters.replace('null', '\"null\"')
+            operation_xml_parameters = eval(operation_xml_parameters)
+    else:
+        has_vm = True
+
+    # HERE STARTS THE PART WHICH IS SPECIFIC TO EACH H5 FILE #
+
+    if class_name == 'Connectivity':
```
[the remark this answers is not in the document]
VinczeRobert: Very nice idea, I will sure do it.

framework_tvb/tvb/core/entities/file/file_update_scripts/005_update_files.py
```diff
@@ -32,17 +32,669 @@
...
+        datatype_class = getattr(sys.modules[root_metadata["module"]],
+                                 root_metadata["type"])
+        h5_class = REGISTRY.get_h5file_for_datatype(datatype_class)
+        root_metadata[H5File.KEY_WRITTEN_BY] = h5_class.__module__ + '.' + h5_class.__name__
+    except KeyError:
+        h5_class = None
+
+    # Other general modifications
+    root_metadata['user_tag_1'] = ''
+    root_metadata['gid'] = GID_PREFIX + root_metadata['gid']
```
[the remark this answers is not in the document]
VinczeRobert: Done

framework_tvb/tvb/core/entities/file/file_update_scripts/005_update_files.py
```diff
@@ -32,17 +32,669 @@
...
+    try:
+        datatype_class = getattr(sys.modules[root_metadata["module"]],
+                                 root_metadata["type"])
+        h5_class = REGISTRY.get_h5file_for_datatype(datatype_class)
+        root_metadata[H5File.KEY_WRITTEN_BY] = h5_class.__module__ + '.' + h5_class.__name__
+    except KeyError:
+        h5_class = None
+
+    # Other general modifications
+    root_metadata['user_tag_1'] = ''
```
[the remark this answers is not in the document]
VinczeRobert: Because as far as I have seen the user_tag_1 in the current version is empty for the datatypes.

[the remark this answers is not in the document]
popaula937: It can be edited from GUI and then it would have a value that should be transferred.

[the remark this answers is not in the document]
VinczeRobert: Okay, I won't remove it then.

framework_tvb/tvb/core/entities/file/file_update_scripts/005_update_files.py
VinczeRobert: Wish I knew sooner, with this there is no need for the dependent_datatype dict...

framework_tvb/tvb/core/entities/file/file_update_scripts/005_update_files.py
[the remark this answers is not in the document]
VinczeRobert: I ended up using a dict instead of an enum because the enum did not want to work for some reason.

framework_tvb/tvb/core/entities/file/file_update_scripts/005_update_files.py
```diff
@@ -32,17 +32,669 @@
...
+    root_metadata[DataTypeMetaData.KEY_DATE] = root_metadata[DataTypeMetaData.KEY_DATE].replace(':', '-')
+    root_metadata[DataTypeMetaData.KEY_DATE] = root_metadata[DataTypeMetaData.KEY_DATE].replace(' ', ',')
+
+    # OBTAIN THE MODULE (for a few data types the old module doesn't exist anymore, in those cases the attr
+    # will be set later
+    try:
+        datatype_class = getattr(sys.modules[root_metadata["module"]],
+                                 root_metadata["type"])
+        h5_class = REGISTRY.get_h5file_for_datatype(datatype_class)
+        root_metadata[H5File.KEY_WRITTEN_BY] = h5_class.__module__ + '.' + h5_class.__name__
```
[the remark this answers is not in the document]
VinczeRobert: What do you mean both places? Which is the second place?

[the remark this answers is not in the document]
popaula937: here and inside _h5_file

framework_tvb/tvb/core/neocom/h5.py
```diff
@@ -309,3 +313,86 @@ def gather_all_references_of_view_model(gid, base_dir, ref_files):
             uuid_files.append(h5_file.path)
             gather_all_references_by_index(h5_file, uuid_files)
         ref_files.extend(uuid_files)
+
+
+# Dependency dict
+h5_files_dependencies = {
+    "RegionVolumeMapping": ["Volume"],
+    "StructuralMRI": ["Volume"],
+    "TimeSeriesVolume": ["Volume"]
```
liadomide, line 322: This is not nice and Could easily become outdated. Can't there be an automatic way?
At import time for example we order by create-date and that proved enough.

liadomide (commented):
This looks promicing.I looked partially and have one question/comment for now

framework_tvb/tvb/core/neocom/h5.py
```diff
@@ -309,3 +311,52 @@ def gather_all_references_of_view_model(gid, base_dir, ref_files):
...
+    projects_folder = FilesHelper.get_projects_folder()
+
+    for project_name in os.listdir(projects_folder):
+        # Getting operation folders inside the current project
+        project_path = os.path.join(projects_folder, project_name)
+        project_operations = os.listdir(project_path)
+        project_operations_base_names = [os.path.basename(op) for op in project_operations]
+
+        # Build a list of the operation numbers and sort them
+        operation_int_names = []
```
popaula937, line 332: This way of iterating through the projects and operations might not cover all cases. 
It would be better to sort the operations by their creation date and iterate over them accordingly, this is what we do also at import.
In your case, since the operations are already in DB, it might be easier to take the creation date from there, instead of reading it from each Operation.xml.

framework_tvb/tvb/core/entities/model/db_update_scripts/helper.py
```diff
@@ -60,3 +61,71 @@ def change_algorithm(module, classname, new_module, new_class):
...
+        return None
+
+    if selected_db == 'sqlite':
+        burst_params_dict = {'datatypes_number': burst_params[0], 'dynamic_ids': burst_params[1],
+                             'range_1': burst_params[2], 'range_2': burst_params[3], 'fk_project': burst_params[5],
+                             'name': burst_params[6], 'status': burst_params[7], 'error_message': burst_params[8],
+                             'start_time': burst_params[9], 'finish_time': burst_params[10],
+                             'fk_simulation': burst_params[12], 'fk_operation_group': burst_params[13],
+                             'fk_metric_operation_group': burst_params[14]}
+        burst_params_dict['start_time'] = datetime.strptime(burst_params_dict['start_time'], '%Y-%m-%d %H:%M:%S.%f')
```
bvalean, line 83: We have the string2date function in the utils.py file. We should use it for these dates conversion.

popaula937:
To be adjusted by @VinczeRobert : a new column `view_model_disk_size` has been added on `Operation` table by PR #264

VinczeRobert:
I only see a few small issues left:
- in DatatypeGroups and OperationGroups the name should be simplified: before it was something like "of DatatypeMeasure varying model_parameters_option_Generic2dOscillator_a x model_parameters_option_Generic2dOscillator_c - 2020-09-27,14-04-54" and now structures like  "model_parameters_option_Generic2dOscillator_a" should be just "model.a"
- disk_size is not computed in the Datatype table
- the visible attribute should be False to DatatypeMeasures which were part of a PSE, but should be True when they are singular

framework_tvb/tvb/core/entities/file/files_update_manager.py
```diff
@@ -195,17 +226,34 @@ def run_all_updates(self):
...
+                file_paths = self._get_all_h5_paths()
+                total_count = len(file_paths)
+                no_ok, no_error = self.__upgrade_h5_list(file_paths)
 
-                self.log.info("Updated H5 files so far: %d [fine:%d, error:%d, ignored:%d of total:%d, in: %s min]" % (
-                    current_idx + len(datatypes_for_page), no_ok, no_error, no_ignored, total_count,
-                    int((datetime.now() - start_time).seconds / 60)))
+                self.log.info("Updated H5 files in total: %d [fine:%d, failed:%d in: %s min]" % (
+                    total_count, no_ok, no_error, int((datetime.now() - start_time).seconds / 60)))
+                delete_old_burst_table_after_migration()
```
liadomide, line 256: This module was supposed to be generic, for all updates. Now it has hard-coded logic for the latest update.
Also the extra param to method upgrade_file is not generic (can't you use self.something for the extra information?)

framework_tvb/tvb/core/entities/file/files_update_manager.py
```diff
@@ -100,38 +103,45 @@ def is_file_up_to_date(self, file_path):
...
-    def upgrade_file(self, input_file_name, datatype=None):
+    def upgrade_file(self, input_file_name, datatype=None, burst_match_dict=None):
         """
         Upgrades the given file to the latest data version. The file will be upgraded
         sequentially, up until the current version from tvb.basic.config.settings.VersionSettings.DB_STRUCTURE_VERSION
-        
+
         :param input_file_name the path to the file which needs to be upgraded
-        :return True, when update was needed and running it was successful.
+        :return 0 when update was not necessary, 1 when update was successful and -1 when it resulted in an error.
```
liadomide, line 112: This is not oop friendly(it is more c style to return error codes)

liadomide (changes requested):
I recommend a more generic style in files_update_manager.py, the rest looks decent

framework_tvb/tvb/core/entities/file/files_update_manager.py
VinczeRobert: It can't be totally generic, because the last update is done in a different way than the old ones. There is an if      TvbProfile.current.version.DATA_CHECKED_TO_VERSION < 4 which separates the 2 styles so the previous updates are still working, I even tested them when doing the tvb data update task.

Also the need of that parameter is a little hard to explain in written form. Perhaps we can discuss it in a short call.

framework_tvb/tvb/core/entities/file/files_update_manager.py
```diff
@@ -195,17 +226,34 @@ def run_all_updates(self):
...
+                    self.log.info(
+                        "Updated H5 files so far: %d [fine:%d, error:%d, ignored:%d of total:%d, in: %s min]" % (
+                            current_idx + len(datatypes_for_page), no_ok, no_error, no_ignored, total_count,
+                            int((datetime.now() - start_time).seconds / 60)))
+            else:
+                # Migrating the configuration file
+                new_stored_settings = {KEY_ADMIN_DISPLAY_NAME: 'Administrator', KEY_ENABLE_KC_LOGIN: False,
+                                       KEY_KC_WEB_CONFIGURATION: '', KEY_KC_CONFIGURATION: 'add_keycloak_path_here'}
+                manager = TvbProfile.current.manager
+                manager.add_entries_to_config_file(new_stored_settings)
```
liadomide, line 247: Instead of the need to add these new Settings, can't we have some defaults for these in the service/config where these are Read first?

VinczeRobert: I added a new upgrade_code script and moved the code there.

VinczeRobert:
I removed the exception raise in 3_update_project because we actually do some project update (switching from Operation.XML to ViewModels), but we don't do it in that script, so I left the script there just for the logs and to still put the project in version 3.

popaula937:
I think this looks decent now, I still have some observations, but they could be tackled in future improvement tasks.

popaula937 (approved):

PR #239, opened 2020-09-10
TVB-2782 Various bugs fixed in web gui

liadomide:
Fixed some of the problems reported on Jira  https://req.thevirtualbrain.org/browse/TVB-2782

As release **2.0.9** is already out, prepare version numbers for the future release.

PR #245, opened 2020-09-14
TVB-2764 conda-forge

liadomide:
- Make pyunicore dependency optional, as it will be very seldom used only, and it does not exist on conda-forge

### DIFF
--- a/framework_tvb/setup.py
+++ b/framework_tvb/setup.py
@@ -40,13 +40,13 @@ import os
 import shutil
 import setuptools
 
-VERSION = "2.0.10"
+VERSION = "2.0.9.1"
 
 TVB_TEAM = "Mihai Andrei, Lia Domide, Stuart Knock, Bogdan Neacsa, Paula Popa, Paula Sansz Leon, Marmaduke Woodman"
 
 TVB_INSTALL_REQUIREMENTS = ["allensdk", "cherrypy", "cryptography", "flask", "flask-restplus", "formencode",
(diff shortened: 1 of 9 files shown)


VinczeRobert (commented):
Everything seems okay as far as I know, I just have a few questions:
1. Why did you delete networkx from the install requirements? I see that it is used in analyers/graph.py.
2. You explained why pyunicore should be an optional dependency, but why BeautifulSoup4 as well?
3. In the PR title the task is TVB-2783, but the branch is TVB-2764 so I don't know which of these two tasks this PR is actually related to. Or both of them?

liadomide:
> Everything seems okay as far as I know, I just have a few questions:
> 
> 1. Why did you delete networkx from the install requirements? I see that it is used in analyers/graph.py.

This should be clear in PR #239 : `networkx` is still needed in `tvb-library` (there it should still be listed as a dependency, and `analyzers.graph` sits under `scientific_library`) but not any more in `tvb-framework` (in the past it was used inside `networkx_parser`, but now we use `pandas` there).

> 2. You explained why pyunicore should be an optional dependency, but why BeautifulSoup4 as well?

I found `bs4` imported only in unit-tests, that is why I placed it in the same bucket as `pytest & co`.

> 3. In the PR title the task is TVB-2783, but the branch is TVB-2764 so I don't know which of these two tasks this PR is actually related to. Or both of them?

You are right to be puzzled. It should only be TVB-2764. I changed the PR title. Thanks!

bvalean (approved):

PR #246, opened 2020-09-17
TVB-2791: Fix metadata lost during export exclude exporting failed operations

adrian-dordea:
When we export a project and import it again, it has been noted that:
- the DT names in the Project -> Data Structure tree are shorter. Investigating shows that `operation_tag` is not correctly preserved in the VM.h5 files.
- if we have operations with errors after export + import, they become with status success.  This happens as we are not longer keeping a dedicated `operation.xml` file. We decided for now to simply do not export operations with errors, as they will most probably also don't have a result DT. We might need to review this in the future, but for now, it should be good enough.

This PR is to ammend these.

### DIFF
--- a/framework_tvb/tvb/adapters/exporters/export_manager.py
+++ b/framework_tvb/tvb/adapters/exporters/export_manager.py
@@ -198,7 +198,7 @@ class ExportManager(object):
         project_datatypes = dao.get_datatypes_in_project(project.id, only_visible=optimize_size)
         to_be_exported_folders = []
         considered_op_ids = []
-        folders_to_exclude = self._get_op_with_errors()
+        folders_to_exclude = self._get_op_with_errors(project.id)
 
         if optimize_size:
             # take only the DataType with visibility flag set ON
@@ -240,15 +240,14 @@ class ExportManager(object):
         return result_path
 
     @staticmethod
-    def _get_op_with_errors():
+    def _get_op_with_errors(project_id):
         """
         Get the operation folders with error base name as list.
         """
-        operations = dao.get_operations_in_group(None)
+        operations = dao.get_operations_with_error_in_project(project_id)
         op_with_errors = []
         for op in operations:
-            if op.status == STATUS_ERROR:
-                op_with_errors.append(op.id)
+            op_with_errors.append(op.id)
         return op_with_errors
 
     def _build_data_export_folder(self, data):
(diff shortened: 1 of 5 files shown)


VinczeRobert (approved):

framework_tvb/tvb/adapters/exporters/export_manager.py
```diff
@@ -235,6 +239,18 @@ def export_project(self, project, optimize_size=False):
 
         return result_path
 
+    @staticmethod
+    def _get_op_with_errors():
+        """
+        Get the operation folders with error base name as list.
+        """
+        operations = dao.get_operations_in_group(None)
```
liadomide, line 247: This query might be inefficient. I think it is currently browsing ALL the projects for ops in error.
Wouldn't be better to have a dedicated query just for the current project ops in error ?

liadomide (commented):
These changes look ok to me.
I have only a small optimization suggestion left bellow

framework_tvb/tvb/adapters/exporters/export_manager.py
```diff
@@ -235,6 +239,18 @@ def export_project(self, project, optimize_size=False):
...
 
+    @staticmethod
+    def _get_op_with_errors(project_id):
+        """
+        Get the operation folders with error base name as list.
+        """
+        operations = dao.get_operations_in_project(project_id)
+        op_with_errors = []
+        for op in operations:
+            if op.status == STATUS_ERROR:
```
popaula937, line 250: The query could return only the operations with error status, so we would not need this check anymore

framework_tvb/tvb/tests/framework/core/services/project_service_test.py
```diff
@@ -300,7 +300,7 @@ def test_project_disk_size(self):
 
             ratio = float(actual_disk_size) / project.disk_size
             msg = "Real disk usage: %s The one recorded in the db : %s" % (actual_disk_size, project.disk_size)
-            assert ratio < 1.4, msg
+            assert ratio < 1.7, msg
```
liadomide, line 303: I do not think this change is healthy. We should rather find why the disk size is so much different. It should be very close to 1.0 ratio

popaula937: Fixed in PR #252. At first, I thought all files on the disk were taken into consideration when computing the disk usage of a project. And since we moved from Operation.xml to ViewModel.h5, the disk usage would have increased.
But now I see that only the H5 files were taken into consideration before, so I made a fix to exclude the view models.

liadomide: Maybe before the operation.xml files were considered irrelevant, or it was a bug. I think VMs should be considered in the calculation, as they can be many and sometimes large.

PR #252, opened 2020-09-30
TVB-2791. Fix disk usage computation test

popaula937:
Exclude view model H5 files from disk usage computation

### DIFF
--- a/framework_tvb/tvb/tests/framework/core/base_testcase.py
+++ b/framework_tvb/tvb/tests/framework/core/base_testcase.py
@@ -88,6 +88,7 @@ from tvb.core.entities.model.model_project import *
 from tvb.core.entities.model.model_datatype import *
 import decorator
 from tvb.core.adapters.abcdisplayer import ABCDisplayer
+from tvb.core.neotraits.h5 import H5File, ViewModelH5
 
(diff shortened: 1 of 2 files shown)


liadomide:
Wouldn't be correct to include all files in the project size, thus not exclude the VMs from the actual size in test, but consider them also for the Project disk size computed field?

popaula937:
> Wouldn't be correct to include all files in the project size, thus not exclude the VMs from the actual size in test, but consider them also for the Project disk size computed field?

I agree but I think that would be a separate task since currently, we determine the project size by adding all Datatype.disk_size values from DB and we do not keep any information about the view_model.h5 size in DB.

liadomide:
> > Wouldn't be correct to include all files in the project size, thus not exclude the VMs from the actual size in test, but consider them also for the Project disk size computed field?
> 
> I agree but I think that would be a separate task since currently, we determine the project size by adding all Datatype.disk_size values from DB and we do not keep any information about the view_model.h5 size in DB.

I think a task already exists in Jira .... 
Could we put disk size of the VM on the Operation table in DB?

popaula937:
Will be fixed by task [TVB-2767](https://req.thevirtualbrain.org/browse/TVB-2767)

PR #254, opened 2020-10-05
HBP-121 Fix keycloak login after latest HBP keycloak server changes.

bvalean:


popaula937 (approved):
As long as this change does not affect our other deploys (eg: FRH), I am ok with it.

PR #256, opened 2020-10-05
Folder restructuring, index of CUDA models change and BOLD kernel addition

DeLaVlag:
Hi @maedoc and @sdiazpier, 
I have restructured the folder of dsl to rateML containing both the python and cuda generators. I changed the index of the models to have variable grid.y size as well and added the BOLD kernel to the templating. I have regenerated all the cuda models which are also part of this PR.


DeLaVlag:
working on a unified generator which incorporates these changes. no need to merge this first

PR #257, opened 2020-10-07
WIP TVB-2747: Implement Import/Export for PSE

adrian-dordea:
add operation group h5 and datatype measure + export op_gr and import the new dt's for pse import


framework_tvb/tvb/core/entities/file/simulator/operation_group_h5.py
```diff
@@ -0,0 +1,84 @@
...
+#   Frontiers in Neuroinformatics (7:10. doi: 10.3389/fninf.2013.00010)
+#
+#
+import uuid
+from tvb.basic.neotraits.api import Attr
+from tvb.core.entities.file.exceptions import MissingDataSetException
+from tvb.core.neotraits.h5 import Json, H5File, Scalar, Reference
+
+
+class OperationGroupH5(H5File):
```
liadomide, line 37: This is a bit controversial.
Have you considered instead that on each ViewModel part of a group to simply store the ranges? This was the previous approach (approx) and the redundancy is not much but it preserves the structure without the need to have an odd Group separate serialization

framework_tvb/tvb/adapters/datatypes/db/mapped_value.py
```diff
@@ -85,3 +86,9 @@ def display_name(self):
                 result = result + " -- " + entry + ' : ' + str(metric_value)
 
         return result
+
+    def fill_from_has_traits(self, datatype):
+        # type: (DatatypeMeasure)  -> None
+        super(DatatypeMeasureIndex, self).fill_from_has_traits(datatype)
```
liadomide, line 92: What is the link with DatatypeMeasure here? Maybe a copy paste mistake?

framework_tvb/tvb/core/services/import_service.py
```diff
@@ -416,10 +450,19 @@ def load_datatype_from_file(self, current_file, op_id, datatype_group=None, curr
...
             h5_file.load_into(burst)
             result = burst
+        elif h5_class is OperationGroupH5:
+            h5_file = OperationGroupH5(current_file)
+            operation_group = OperationGroup(current_project_id)
+            h5_file.load_into(operation_group)
+            result = operation_group
+
+        elif h5_class is DatatypeMeasureH5:
+            h5_file = DatatypeMeasureH5(current_file)
```
popaula937, line 462: The way you are handling the datatype measure entities looks redundant to me. 
Either you keep this `if` and load directly the `DatatypeMeasureIndex` from the `DatatypeMeasureH5`, or you define the new `DatatypeMeasure` object and go through the generic flow of loading from H5 that you do in the `else` below: 
`datatype, generic_attributes = h5.load_with_links(current_file)
`

popaula937: Seeing the code here and knowing that this part is problematic also for the data migration task, I tend to vote for adding the DatatypeMeasure object, so we can use the generic code to go from DatatypeMeasureH5 to DatatypeMeasureIndex and avoid the need to always treat it as a special case.

framework_tvb/tvb/core/entities/file/simulator/operation_group_h5.py
popaula937: Yes, we are still considering simplifying the solution by using only the ViewModel, but I think only ranges are not enough for us to determine which operations are part of the same group, we need some extra information (maybe the GID of the OpGroup).

popaula937:
Reviewed in PR #264, we will merge that one.

PR #260, opened 2020-10-14
TVB-2781: Don't allow to upload H5 files for which TVB doesn't have the dependent datatypes in DB

VinczeRobert:


framework_tvb/tvb/core/services/import_service.py
```diff
@@ -409,6 +409,15 @@ def load_datatype_from_file(self, current_file, op_id, datatype_group=None, curr
         """
         self.logger.debug("Loading DataType from file: %s" % current_file)
         h5_class = H5File.h5_class_from_file(current_file)
+        reference_list = h5_class(current_file).gather_references()
+
+        for reference in reference_list:
```
popaula937, line 414: Access the tuple element you need directly in `for` instead of having to do `reference[1]` later :
`for _, reference_gid in reference_list`

framework_tvb/tvb/core/services/import_service.py
```diff
@@ -409,6 +409,15 @@ def load_datatype_from_file(self, current_file, op_id, datatype_group=None, curr
...
+            if reference[1] is not None:
```
popaula937, line 415:
```
if not reference_gid:
     continue
```

framework_tvb/tvb/core/services/exceptions.py
```diff
@@ -75,6 +75,12 @@ class ImportException(ServicesBaseException):
     """
 
 
+class MissingReferenceException(ImportException):
+    """
+    Exception to be thrown in case there are missing references when importing a file.
```
popaula937, line 80: ... H5 file.

framework_tvb/tvb/core/services/import_service.py
```diff
@@ -409,6 +409,15 @@ def load_datatype_from_file(self, current_file, op_id, datatype_group=None, curr
...
+                ref_index = dao.get_datatype_by_gid(reference[1].hex)
+                if ref_index is None:
+                    raise MissingReferenceException('The file you wish to upload has missing reference files. You ' +
```
popaula937, line 418: shorter? 

> Imported file depends on datatypes that do not exist. Please upload those first!

popaula937 (commented):
Looks good. Do we have optional References? We should handle that case, too.

framework_tvb/tvb/adapters/uploaders/tvb_importer.py
```diff
@@ -131,7 +131,7 @@ def launch(self, view_model):
                         datatype = service.load_datatype_from_file(view_model.data_file, self.operation_id)
                         service.store_datatype(datatype, view_model.data_file)
                         self.nr_of_datatypes += 1
-                    except ImportException as excep:
+                    except (MissingReferenceException, ImportException) as excep:
```
liadomide, line 134: If MissingRef inherits from ImportExc isn't enough to catch the ImportExc here?

framework_tvb/tvb/core/services/import_service.py
```diff
@@ -409,6 +409,15 @@ def load_datatype_from_file(self, current_file, op_id, datatype_group=None, curr
         """
         self.logger.debug("Loading DataType from file: %s" % current_file)
         h5_class = H5File.h5_class_from_file(current_file)
+        reference_list = h5_class(current_file).gather_references()
```
liadomide, line 412: Is this happening late enough in the flow to work correctly in case of a full project import (where multiple linked dts are being imported together)?

VinczeRobert: Hmm, not really. I tested it and if there is a missing referenced datatype, then it doesn't import the project. I guess the project should still be imported, only the datatypes with missing references should be skipped, right?

liadomide: Yes, for partially correct projects we should do a best effort to import what we can, and show a warning that not everything was imported.

What I had in mind was a project PR1 with Conn1, Surface2 and RegMapp (referencing Conn1 and Surface2) datatypes to be imported. Will this import work (if we have an empty db before)?
Could be tested also with the current DefaultProject this situation.

framework_tvb/tvb/adapters/uploaders/tvb_importer.py
```diff
@@ -41,7 +41,7 @@
 from tvb.core.neotraits.forms import TraitUploadField
 from tvb.core.neotraits.uploader_view_model import UploaderViewModel
 from tvb.core.neotraits.view_model import Str
-from tvb.core.services.exceptions import ImportException
+from tvb.core.services.exceptions import ImportException, MissingReferenceException
```
liadomide, line 44: This import could also be reverted.

framework_tvb/tvb/core/services/import_service.py
liadomide: After seeing the entire File again, i would answer this myself: 
in case of a project import with linked dts, we order by operation create date, iterate these ops, store in db all linked dts for one op, then we continue with the next operation. The check recently added by Robert about linked dts happens only late, when we are about to load a dt within the ops loop(previous ops were already stored).
Still we might have a problem if inside one operation folder we have linked dts (both conn1 and regionMapping are in the same operation generated). Could you validate/ check this situation, please?

VinczeRobert: It doesn't work for that cases, because at that point the datatypes are not taken in the order of their creation. I tried an example with a RegionVolumeMapping depending on a Volume and the RVM came before the Volume so the import failed even though the Volume was there.

VinczeRobert: But where could I add the check if not when it iterates through the operations?

popaula937: Sorting based on the creation date should be enough also here, but I think it should be done in `_load_datatypes_from_operation_folder`, could that code be extracted and reused in both places (migration and import)?

framework_tvb/tvb/core/services/import_service.py
```diff
@@ -316,21 +316,29 @@ def import_project_operations(self, project, import_path):
...
-                dao.store_entity(operation_entity)
+                    # Create and store view_model from operation
+                    view_model = self._get_new_form_view_model(operation_entity, operation_data.info_from_xml)
+                    h5.store_view_model(view_model, new_op_folder)
+                    operation_entity.view_model_gid = view_model.gid.hex
+                    dao.store_entity(operation_entity)
 
-                self._store_imported_datatypes_in_db(project, operation_datatypes)
-                imported_operations.append(operation_entity)
+                    self._store_imported_datatypes_in_db(project, operation_datatypes)
```
liadomide, line 336: Only here the dts are to be stored in db, the query to db about possible missing references is much earlier.
Thus, ordering dts by createDate will be necesary But not enough.
Either we check Reference later (in _store_imported_dts_in_db) or we check Reference Both in dao (db) and in the list of dts for the curent operațional (not yet stored in db, only in memory).

framework_tvb/tvb/adapters/uploaders/tvb_importer.py
liadomide: I this this is not yet solved

framework_tvb/tvb/core/services/import_service.py
VinczeRobert: I think it is okay now.

liadomide: Looks promising. Is the order of dts by create-date happening?
If not, then sometimes it might work, other times not

VinczeRobert: It is happening just before the block of code added by me, at the beginning of the method so I think it should work for all cases.

PR #261, opened 2020-10-14
TVB-2642: Show error message when period is bigger than simulation length

VinczeRobert:
I thought that this would be the best way to handle this problem, otherwise we would need an if at simulation launching to prevent it and another if in JS as well for displaying the message and that would disrupt the flow a bit.

### DIFF
--- a/scientific_library/tvb/simulator/monitors.py
+++ b/scientific_library/tvb/simulator/monitors.py
@@ -56,6 +56,7 @@ Conversion of power of 2 sample-rates(Hz) to Monitor periods(ms)
 
 import abc
 import numpy
+from tvb.core.adapters.exceptions import InvalidParameterException
 from tvb.datatypes.time_series import (TimeSeries, TimeSeriesRegion, TimeSeriesEEG, TimeSeriesMEG, TimeSeriesSEEG,
                                        TimeSeriesSurface)
 from tvb.simulator import noise
@@ -856,6 +857,10 @@ class Bold(Monitor):
 
     def config_for_sim(self, simulator):
         super(Bold, self).config_for_sim(simulator)
+
+        if self.period > simulator.simulation_length:
+            raise InvalidParameterException("Sampling period can not be bigger than the simulation length!")
+
         self.compute_hrf()
         sample_shape = self.voi.shape[0], simulator.number_of_nodes, simulator.model.number_of_modes
         self._interim_stock = numpy.zeros((self._interim_istep,) + sample_shape)
--- a/scientific_library/tvb/simulator/monitors.py
+++ b/scientific_library/tvb/simulator/monitors.py
@@ -56,7 +56,6 @@ Conversion of power of 2 sample-rates(Hz) to Monitor periods(ms)
 
 import abc
 import numpy
-from tvb.core.adapters.exceptions import InvalidParameterException
 from tvb.datatypes.time_series import (TimeSeries, TimeSeriesRegion, TimeSeriesEEG, TimeSeriesMEG, TimeSeriesSEEG,
                                        TimeSeriesSurface)
 from tvb.simulator import noise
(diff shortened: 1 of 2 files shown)


scientific_library/tvb/simulator/monitors.py
```diff
@@ -56,6 +56,7 @@
 
 import abc
 import numpy
+from tvb.core.adapters.exceptions import InvalidParameterException
```
liadomide, line 59: This import is invalid. Tvb-framework depends on tvb-library (not the reverse).

scientific_library/tvb/simulator/monitors.py
```diff
@@ -856,6 +857,10 @@ def compute_hrf(self):
 
     def config_for_sim(self, simulator):
         super(Bold, self).config_for_sim(simulator)
+
+        if self.period > simulator.simulation_length:
```
liadomide, line 861: I fear this might be too early as a check for the library mode. See the examples there: config happens early, the the simulator __call__ loop only gets the simulation length. Other examples configure everything, do a small loop (initial results are ignored), then continue the simulation for a longer length.

liadomide (commented):
Have you considered the SimulatorAdapter.launch or configure methods for a similar check?
Could this condition be generic for all monitors?

popaula937:
I think this should be solved at the framework level, as early as possible. Why let the code go through all the flow of preparing an operation when we know this is a condition that breaks the simulation?

liadomide:
> I think this should be solved at the framework level, as early as possible. Why let the code go through all the flow of preparing an operation when we know this is a condition that breaks the simulation?

I suspect Robert tries to avoid an "if monitor==Bold" ugly condition (which is a valid reason). Also a check in the JS won't be consistent with previous code in this area.
But I also agree with Paula to make this check as early as possible.
I still think the SinulatorAdapter could work, and this check to be applied for all selected monitors.
Earlier that the simulatorAdapter would be the simulator forms or the simulatorservice. Not sure how clean that would be...
What do you propose?

liadomide:
> > > I think this should be solved at the framework level, as early as possible. Why let the code go through all the flow of preparing an operation when we know this is a condition that breaks the simulation?
> 
> > 
> 
> > I suspect Robert tries to avoid an "if monitor==Bold" ugly condition (which is a valid reason). Also a check in the JS won't be consistent with previous code in this area.
> 
> > But I also agree with Paula to make this check as early as possible.
> 
> > I still think the SinulatorAdapter could work, and this check to be applied for all selected monitors.
> 
> > Earlier that the simulatorAdapter would be the simulator forms or the simulatorservice. Not sure how clean that would be...
> 
> > What do you propose?
> 
> 
> 
> Well, if the condition is generic for all monitors we won't have the "if monitor==Bold", it will be a validation for the sim_length value in report with all monitor period values. 

Yes, but this is only recently proposed by me above. The original bug only mentions Bold

> Not sure what change we will need on JS side though..

Is not "needed" in the JS, I think we all agree on this. The JS was mentioned only because that would hypothetically be the earliest possible validation code.

> isn't it enough to send the message from server side with 'error' type as we do when some exceptions are raised?

Exactly. The question is where to do the check and raise the exception. Would the simulator specific Forms validation methods work ? That Could be a good place for us to put other validations also .

VinczeRobert:
I am sorry, it escaped my notice that it exists in all monitors, it's just that for the others the default value didn't produce this error so I never payed attention to that. I hope it's okay now.

liadomide (approved):
From my pov this solution is safe and compatible with previous validations. So, I would agree to merge it like this.

But maybe you can suggest a more optimal solution ? (raise this even earlier) ?

VinczeRobert:
> From my pov this solution is safe and compatible with previous validations. So, I would agree to merge it like this.
> 
> But maybe you can suggest a more optimal solution ? (raise this even earlier) ?

Well the earliest that we could raise it is just before launching the simulation I guess, because only then we have the simulation length. The next option would be in the simulator service, but it would mean that we need to put the validation in two places (for normal simulations and for PSE ones).

I was also trying to think of a way to make the simulation work if there is at least one monitor when the period is okay (in the case we choose more). I thought that we could just "delete" the monitors for which the period is not okay and if there is no valid monitor, only then we raise the exception. But in that case I don't know if there is a way to let the user know this.

PR #262, opened 2020-10-15
TVB-2798 Init tvb-framework storage and db independently of TVB GUI for the command profile

bvalean:


liadomide (approved):
These changes could work. 

Pls do manual tests:
- start clean framework cherrypy
- restart framework 
- start command examples on clean (no ~\.tvb.configuration)
- start command examples in the same time while cherrypy is running (ops created in command should be visible in web)
- cherry settings edit something and restart 

These are the scenarios I can think of... :-)

framework_tvb/tvb/interfaces/command/demos/project/project_export.py
```diff
@@ -57,10 +51,9 @@ def run_export(project_id, loose_irrelevant=False):
 
 
 if __name__ == '__main__':
+    from tvb.config.init.initializer import command_initializer
 
-    if len(argv) < 2:
-        print("You should specify a project ID to be exported!")
-
-    print("We will try to export project with ID: " + str(argv[1]))
+    command_initializer()
```
liadomide, line 56: Should we consider importing lab.py instead?
And promote usage of lab in all command examples as a good practice?

liadomide (approved):

PR #263, opened 2020-10-17
Intermediate state vars

dionperd:
In this pull request from branch intermediate_state_vars, Lionel and I introduce state variables that:
- are not integrated,
- but correspond to intermediate results (e.g., the current I and rate R of the ReducedWongWang models),
- which can be computed by model methods either before or after integration of a time step.

To this end, we had to modify 
- the Base model to include 
     - a list of variables that have to be integrated, which defaults to all the state variables, 
     - and a related state variables mask, which is True (False) for the state variables that ARE (NOT) integrated, respectively,
     - as well as possible methods to update such non-integrated state variables either before integration (in which case all the arguments of the Integrator.scheme are also considered here (i.e., coupling, local coupling, stimulus), or after the integration, in which case the only argument is the state itself,
- the integrator to add methods that compute the next time step with or without updating such state variables,
- the simulator to configure the integrator by choosing the correct method for integration, with or without updating such state variables, and to use the resulting method inside the __call__ method.

I have added the following tests:
- a specific test model with 3 state variables and 2 non-integrated state variables, which are given before integration following:
state[3] += state[0]
state[4] += state[1] + state[2]
and after integration reversing the above:
state[3] -= state[0]
state[4] -= state[1] + state[2],
so that together with a dfun: 0.0*state,
the state should be left unaltered if the test is succesful,
- an integration test testing all integrators in computing 1 step following the above model methods/equations,
- a simulator test configuring the simulator, and the model, for the above scenario, and testing that the configuration is successful, e.g. the correct integrator method is selected and the state variables mask is formed correctly.

Given that this addition to TVB is considered necessary for CoSimulation, we would appreciate a quick review!


scientific_library/tvb/simulator/simulator.py
```diff
@@ -203,6 +212,7 @@ def _configure_integrator_boundaries(self):
         else:
             self.integrator.bounded_state_variable_indices = None
             self.integrator.state_variable_boundaries = None
+        self.bound_and_clamp = self.integrator.bound_and_clamp
```
popaula937, line 215: Why keep the `bound_and_clamp` on the simulator?

popaula937 (commented):
How about minimizing the changes needed at the simulator level by moving the condition for update inside the `Integrator` and define a single method to `integrate()` that would wrap over `scheme()`?

scientific_library/tvb/simulator/simulator.py
dionperd: So, I removed the bound_and_clamp form the simulator. Using the one of integrator, and applying it to the history as well by default.
As for the two distinct methods for integrate() and integrate_with_update() they are necessary to avoid an if statement for every time step of integration, I guess.

popaula937:
Another idea would be to define a decorator for the `Integrator`, which would have the same API as an `Integrator` (and would get the current integrator as a parameter in the constructor), but would enhance the `integrate()` method with the pre/post integrate steps you are doing in `integrate_with_update()`.
In the `Simulator.configure()`, you would apply the decorator on the integrator in case your condition is False, and inside `__call__` you would call directly `self.integrator.integrate()`
This would be a way to keep the same API for integrators and separate the new functionality intended to be used for co-simulations.

dionperd:
> Another idea would be to define a decorator for the `Integrator`, which would have the same API as an `Integrator` (and would get the current integrator as a parameter in the constructor), but would enhance the `integrate()` method with the pre/post integrate steps you are doing in `integrate_with_update()`.
> In the `Simulator.configure()`, you would apply the decorator on the integrator in case your condition is False, and inside `__call__` you would call directly `self.integrator.integrate()`
> This would be a way to keep the same API for integrators and separate the new functionality intended to be used for co-simulations.

I kind of like this idea, but I don't know exactly what a decorator is and how it could be implemented.

maedoc:
Please don’t jump on the decorator idea before we have a chance to review.  The pattern I would prefer to see is generating a closure (nested function) which can be called instead of a object method.  This would allow for data driven selection of algorithm but in line in a way that leaves performance improvements available later on.

Secondly at the risk of being pedantic, integration per se does not involve this extra stuff like bounding and clamping but just the differential equation integration.  I would prefer to keep a separation of concerns and not push a lot of extraneous stuff into integrators.

Sorry for the latency on review I will try to provide more feedback tomorrow.

maedoc (approved):
I think the changes are OK.  There's a lot happening at runtime with bounds and clamps but as long as the logic is sound and well tested, and static i.e. doesn't change during the simulation, we can clean up performance-wise afterward.  Given TVB performance is not an issue for cosim, I think we can merge.

maedoc:
@dionperd if you don't have any updates ot push here, I think we should move ahead and merge, ok?

dionperd:
Yes. Please merge!

dionperd:
I have merged the master branch and I am not sure what the tests are telling me, i.e., whether something has failed, and if yes, what that is....

maedoc:
The tests passed, the GitHub build config is confused right now.

PR #264, opened 2020-10-19
TVB-2747. Review and simplify import/export of PSE.

popaula937:
This is a review for PR #257 that simplifies the procedure of export/import PSE by using only the ViewModel instead of adding an extra OperationGroupH5.
@adrian-dordea pls review and test various scenarios for this, you know better where the possible problems are.

### DIFF
--- a/framework_tvb/tvb/core/entities/file/simulator/datatype_measure_h5.py
+++ b/framework_tvb/tvb/core/entities/file/simulator/datatype_measure_h5.py
@@ -48,6 +48,6 @@ class DatatypeMeasureH5(H5File):
     def __init__(self, path):
         super(DatatypeMeasureH5, self).__init__(path)
         # Actual measure (dictionary Algorithm: single Value)
-        self.metrics = Json(Attr(str), self, name='metrics')
+        self.metrics = Json(DatatypeMeasure.metrics, self)
         # DataType for which the measure was computed.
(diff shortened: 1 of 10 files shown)


liadomide (approved):
These look ok as changes.

The 'is-metric' flag seems a bit odd on the ViewModel. Theoretically we can compute it at import time, but I understand its usage, and could also stay.

adrian-dordea (approved):

popaula937:
> These look ok as changes.
> 
> The 'is-metric' flag seems a bit odd on the ViewModel. Theoretically we can compute it at import time, but I understand its usage, and could also stay.

I think it will be a bit tricky to compute it in order to differentiate between operations that are metrics of a simulation PSE (to know when BurstConfiguration should be updated) and operations that were generated in an analyzer PSE. We have info: operation_group, ranges and parent_burst which will be the same for both cases.

liadomide:
> > These look ok as changes.
> 
> > 
> 
> > The 'is-metric' flag seems a bit odd on the ViewModel. Theoretically we can compute it at import time, but I understand its usage, and could also stay.
> 
> 
> 
> I think it will be a bit tricky to compute it in order to differentiate between operations that are metrics of a simulation PSE (to know when BurstConfiguration should be updated) and operations that were generated in an analyzer PSE. We have info: operation_group, ranges and parent_burst which will be the same for both cases. 

First: I agree to keep this flag.

When suggesting to compute it at import time, I was considering to look at the operation/VM type (if equals something, then it means is metric).

PR #266, opened 2020-10-20
TVB-2790: Make PSE Simulations run with surfaces as gid params

VinczeRobert:
The first problem is that there should be a filter to only run the simulations with cortical surfaces.

The second and bigger problem is that the way the code is written, in _set_simulator_range_parameter uuid's are expected, but the SimulatorAdapterModel not only wants CortexViewModels, but on those we also need to set region mappings and these are required, not optional. So taking these into consideration I guess that we can't run PSE on surfaces if we didn't choose a surface at the respective fragment. Also, if we have more cortical surfaces, we will just add the same region mapping even if the dimensions don't match?

If I know how this is supposed to work, I can finish this task.

### DIFF
--- a/framework_tvb/tvb/interfaces/web/templates/jinja2/visualizers/commons/scripts/regionSelector.js
+++ b/framework_tvb/tvb/interfaces/web/templates/jinja2/visualizers/commons/scripts/regionSelector.js
@@ -255,6 +255,7 @@ RegionSelectComponent.prototype.selectedIndices = function(arg){
         this._set_indices(arg);
         this._dropDown.val("[]");
         this.$dom.trigger("selectionChange", [this._selectedValues.slice()]);
+
     }
 };
 
--- a/framework_tvb/tvb/adapters/simulator/range_parameters.py
+++ b/framework_tvb/tvb/adapters/simulator/range_parameters.py
@@ -53,11 +53,9 @@ class SimulatorRangeParameters(object):
                                           Range(lo=0.01, hi=100.0, step=1.0),
                                           isinstance(Simulator.conduction_speed, NArray))
         connectivity = RangeParameter(Simulator.connectivity.field_name, Connectivity, self.connectivity_filters)
-        surface = RangeParameter(Simulator.surface.field_name, Surface, self.connectivity_filters)
 
         return OrderedDict({Simulator.conduction_speed.field_name: conduction_speed,
-                            Simulator.connectivity.field_name: connectivity,
-                            Simulator.surface.field_name: surface})
+                            Simulator.connectivity.field_name: connectivity})
 
     def _ensure_correct_prefix_for_param_name(self, prefix, param):
         prefix = prefix + '.'


liadomide:
These are very good remarks.
We could choose answers to all these, but won't be straight use-cases, some times not expected by the user.

As surface simulations are rare and very long now anyway, I suggest we don't allow pse over surfaces, thus to close this issue with won't fix (copy your finding on this pr also on the task as explanation).

liadomide:
Something should still be done in the code, that pse on surfaces can not be chosen from the GUI

popaula937:
Currently, we display surface as a PSE param option, even when we don't configure a surface-simulation. I think it does not make sense to display it if we don't want a surface-simulation.

Indeed this topic raises many questions.
Not sure whether you mean:
1. don't allow PSE over surface-simulations at all (because they are rare)
2. allow PSE over surface-simulations but don't allow to vary the surfaces as PSE params

For the second option, when allowing PSE over surface-simulations, not only the surface is linked to the RegionMapping, but also the connectivity. So, if we decide to hide the surface from PSE params, then we should do the same for connectivity. We would allow PSE only over scalars.

liadomide:
I meant the second case in your description.

Varying the connectivity is a scenario I heard of being used intensively (on a group of patients, or the same subject but after disease in various stages), so I would suggest we struggle to keep pse by ranging connectivities. The problem with region mapping when ranging over connectivities is only at the projection monitors level, right?

popaula937:
> I meant the second case in your description.
> 
> Varying the connectivity is a scenario I heard of being used intensively (on a group of patients, or the same subject but after disease in various stages), so I would suggest we struggle to keep pse by ranging connectivities. The problem with region mapping when ranging over connectivities is only at the projection monitors level, right?

My previous comment refers strictly to surface-simulations, and the fact that we choose a Surface and a RegionMapping and then we want to vary the Connectivity during PSE. That means we should filter connectivities and use only the ones that are linked to same RegionMapping and it will be similar with the Surface problems mentioned by Robert. But we can support it in some way.
For region-simulations, I think we use the RM only at monitors level. We should define that use-case better, too, maybe in a separate task.

liadomide:
I understand now what you mean. Good findings to be remembered.

I suggest to don't show the surfaces as param to range over. 

The region mapping incompatibilities might not be that bad, if the ranged connectivities have the same number of regions. I would leave these unrestricted: let it fail at runtime, and we will see how the user's feedback if for these scenarios.

VinczeRobert:
So in conclusion I should just make the surfaces not appear as a range param and close the task with 'Won't Fix'?

framework_tvb/tvb/interfaces/web/templates/jinja2/visualizers/commons/scripts/regionSelector.js
```diff
@@ -254,7 +254,7 @@ RegionSelectComponent.prototype.selectedIndices = function(arg){
     }else{
         this._set_indices(arg);
         this._dropDown.val("[]");
-        this.$dom.trigger("selectionChange", [this._selectedValues.slice()]);
```
liadomide, line 257: This removal is from another task, right?

liadomide (changes requested):
I think it is ok for now, except the extra commit in regionsassosciator js

liadomide (approved):

popaula937 (approved):
Maybe squash the commits before merge

framework_tvb/tvb/interfaces/web/templates/jinja2/visualizers/commons/scripts/regionSelector.js
VinczeRobert: Yes

PR #267, opened 2020-10-20
Rawvoi coupling monitors

dionperd:
In this pull request from rawvoi_coupling_monitors branch I introduce the following new monitors:
- RawVoi, i.e., a Raw monitor for which the user can choose Vois
- Coupling, a Raw monitor for node_coupling,
- CouplingTemporalAverage, a TemporalAverage monitor for node_coupling.

The Coupling monitors set vois from cvars:
self.voi = self.variables_of_interest
if self.voi is None or self.voi.size == 0:
      self.voi = numpy.r_[:len(simulator.model.cvar)]

In order to shorten code, I have introduced _config_vois and _config_time methods to the base class Monitor, and to other monitors accordingly.

Some changes were needed in simulator.py:

The _loop_monitor_output() had to change to take node_coupling as argument and to choose whether to use the observed state or node_coupling, depending on whether the monitor is a Coupling monitor or not.

Moreover, we needed to compute the node_coupling of step + 1 just after looping history inside the __call__() and compute the node_coupling for start_step = self.current_step + 1 before getting in the integration loop, i.e., for the initial condition.

Finally, I have added tests for configuration in monitor_tests and for integration in simulator_test, following the example of other monitors.


maedoc:
sorry for delay, was out of office for last two weeks. I'll review within the week

scientific_library/tvb/simulator/monitors.py
```diff
@@ -174,19 +179,43 @@ class Raw(Monitor):
...
+
+    _ui_name = "Selected State Variables' Raw recording"
+
+    def _config_vois(self, simulator):
+        self.voi = self.variables_of_interest
+        if self.voi is None or self.voi.size == 0:
+            self.voi = numpy.r_[:len(simulator.model.variables_of_interest)]
+
+    def sample(self, step, state):
+        return super(RawVoi, self).sample(step, state[self.voi])
```
maedoc, line 216: Python3 lets you write just `super().sample()` now, which helps avoid errors with copy paste.

scientific_library/tvb/simulator/monitors.py
```diff
@@ -374,6 +401,49 @@ def sample(self, step, state):
             return [time, avg_stock]
 
 
+class Coupling(RawVoi):
```
maedoc, line 404: I think Coupling is getting a bit overloaded in TVB, and it would be preferable to make the name more specific.  For instance, are you montioring the afferent or efferent coupling variables?

maedoc (approved):
looks ok to me, I would mainly rename to AfferentMonitor or similar

scientific_library/tvb/simulator/monitors.py
dionperd: I just renamed to AfferentCoupling. I am not sure that Afferent is clear enough. However, I can also rename to Afferent if you insist.

scientific_library/tvb/simulator/monitors.py
dionperd: There are more super() calls there, so, I would suggest that we do a dedicated correction to a larger part of the code if we wish to change this.

maedoc: I agree

scientific_library/tvb/simulator/monitors.py
maedoc: Yes agreeed that afférent alone is insufficient.  How about PostSynapticTrace?

scientific_library/tvb/simulator/simulator.py
```diff
@@ -386,16 +388,18 @@ def __call__(self, simulation_length=None, random_state=None):
...
-        for step in range(self.current_step + 1, self.current_step + n_steps + 1):
-            # needs implementing by hsitory + coupling?
-            node_coupling = self._loop_compute_node_coupling(step)
+        for step in range(start_step, start_step + n_steps):
+            # needs implementing by history + coupling?
             self._loop_update_stimulus(step, stimulus)
             state = self.integrator.scheme(state, self.model.dfun, node_coupling, local_coupling, stimulus)
             self._loop_update_history(step, n_reg, state)
-            output = self._loop_monitor_output(step, state)
+            node_coupling = self._loop_compute_node_coupling(step + 1)
```
lionelkusch, line 401: I think this modify cation is not correct because this will give the coupling for the next time step and not for the current time step.
What do you think?

scientific_library/tvb/simulator/monitors.py
dionperd: To be honest, I don't think that switching to a terminology that doesn't include "Coupling" is an accurate description.
Many of our models are anyway too phenomenological for "postsynaptic" to really mean something.
Everywhere in TVB we call "Coupling", actually node_coupling, or local_coupling this quantity, and therefore, the Monitor is better to use the same wording.
I agreed though with the Afferent distinction because it has to be made against EfferentCoupling.

maedoc:
@dionperd can you rebase this on master and fix conflict? or I can do it, and you update the PR so we can merge?

dionperd (commented):
> @dionperd can you rebase this on master and fix conflict? or I can do it, and you update the PR so we can merge?

So, I think I have done so, and I don't really understand if the tests have passed or failed, or what has failed...

scientific_library/tvb/simulator/simulator.py
maedoc: interesting to shuffle around this way, but should be right

scientific_library/tvb/simulator/monitors.py
maedoc: The pre/post synaptic distinction is held if loosely by whether something occurs before or after the summation, but if we can agree to use AfferentCoupling that's fine too.

maedoc:
> I don't really understand if the tests have passed or failed

the GitHub testing is a little confused it seems, but the library tests passed, which is enough for a PR.  The Jenkins testing is a little more complete and will catch anything else.

maedoc:
@liadomide @popaula937 you were both requested to review; I would say we can merge unless either of you want to complete a review first?

liadomide (approved):
Fine with me to merge, from a "reading the code" level. I hope Jenkins is also happy :-)

maedoc:
thanks @dionperd !

PR #269, opened 2020-10-21
TVB-2787: Fix LargeScaleConnectivity in case of a single hemisphere Connectivity

adrian-dordea:


adrian-dordea:
The problem starts in the case when we don't have a hemisphere and there is no connection between the hemispheres.

Because of this when we try to get ordered_centers of the input hemispheres we get a empty array on which min and max are applied and there is the problem.

popaula937:
The Large Scale Connectivity viewer throws an error in case we want to visualize a Connectivity that has only one hemisphere. And this is happening when we prepare the hemispheres for separate display in the tabs: Hemisphere1 and Hemisphere2. 
We did not have this problem before because we used to group regions into hemispheres by dividing them in half, so, we used to treat a single hemisphere as a full brain. 
Now, we group the regions using the hemispheres array. So, in case of a single-hemisphere-Connectivity, one of the tabs won't have data to display.
We are considering to hide the empty tab or displaying a message when the user clicks on it.

bvalean (approved):
Did a small cleanup

PR #270, opened 2020-10-22
TVB-2751 Migrate simulate_for_mouse and interacting_with_Allen notebooks

bvalean:


popaula937 (commented):
Some comments related to interating_with_Allen notebook:
 
Pls remove the TODO from the first cell.

Also, it would be good to shortly explain the role of the view_model since this is a new concept for the users.

How about doing: `view_model = builder_instance.get_view_model_class()()` instead of `view_model = AllenConnectModel()`, so we would show there is an automated way to get the view_model for an adapter, users don't need to learn the correspondence between these adapter-view_model in order to use it. Again, there could be a comment that would say: _the corresponding view_model is `AllenConnectModel`. You can either instantiate it directly or get it by calling `get_view_model_class`_.

One mention for making this command interface nicer: could help to have a method on the ViewModel to display only the parameters that can be changed by users together with their defaults&explanations (doc on the Attr). What do you think? We can also do this in a next task/sprint.

popaula937:
Also, mind that there already is a method `wait_to_finish` in lab.py to use instead of re-writing the code in many notebooks.
We could also add a method in lab.py to display the operation results for the last cell.

popaula937:
For the simulate_for_mouse notebook: Most of it is based on the tvb-framework now, only the first part that loads a Connectivity in memory seems to use only tvb-library. 
How about changing that to use tvb-framework, too? It could be another method in lab.py that tries to import the chosen Connectivity.h5 and if it is already imported, just loads it.

bvalean:
> For the simulate_for_mouse notebook: Most of it is based on the tvb-framework now, only the first part that loads a Connectivity in memory seems to use only tvb-library.
> How about changing that to use tvb-framework, too? It could be another method in lab.py that tries to import the chosen Connectivity.h5 and if it is already imported, just loads it.

Yep, good idea. I would need the tvb-data package updated on zenodo for this change, because currently I get a migration error. I see that Robert changes also the mouse h5 files in the tvb-data package (https://req.thevirtualbrain.org/browse/TVB-2778?focusedCommentId=18399&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-18399)

popaula937 (approved):
The notebooks that interact with the tvb-framework look easy to understand and use now, will merge this state.

PR #271, opened 2020-10-23
TVB-2767: Review flow for calculating and displaying the disk usage

adrian-dordea:
add a new disk_size column to operation table for view_models and take them into account when calculating the project disk size

### DIFF
--- a/framework_tvb/tvb/core/entities/file/files_helper.py
+++ b/framework_tvb/tvb/core/entities/file/files_helper.py
@@ -423,7 +423,7 @@ class FilesHelper(object):
         """
         Computes the disk usage of all h5 files under the given directory.
         :param start_path:
-        :return: A tuple of size in kiB and number of files
+        :return: A tuple of size in kiB
         """
         total_size = 0
         n_files = 0
@@ -433,7 +433,7 @@ class FilesHelper(object):
                     fp = os.path.join(dir_path, f)
                     total_size += os.path.getsize(fp)
                     n_files += 1
-        return int(round(total_size / 1024.)), n_files
+        return int(round(total_size / 1024.))
 
 
 class TvbZip(ZipFile):
--- a/framework_tvb/tvb/core/entities/file/files_helper.py
+++ b/framework_tvb/tvb/core/entities/file/files_helper.py
@@ -418,6 +418,23 @@ class FilesHelper(object):
             return int(os.path.getsize(file_path) / 1024)
         return 0
 
+    @staticmethod
+    def compute_recursive_h5_disk_usage(start_path='.'):
+        """
+        Computes the disk usage of all h5 files under the given directory.
+        :param start_path:
+        :return: A tuple of size in kiB and number of files
+        """
+        total_size = 0
+        n_files = 0
+        for dir_path, _, file_names in os.walk(start_path):
+            for f in file_names:
+                if f.endswith('.h5'):
+                    fp = os.path.join(dir_path, f)
+                    total_size += os.path.getsize(fp)
+                    n_files += 1
+        return int(round(total_size / 1024.)), n_files
+
 
 class TvbZip(ZipFile):
     def __init__(self, dest_path, mode="r"):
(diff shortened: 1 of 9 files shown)


framework_tvb/tvb/core/entities/model/model_operation.py
```diff
@@ -252,6 +252,7 @@ class Operation(Base, Exportable):
     user_group = Column(String, default=None)
     range_values = Column(String, default=None)
     estimated_disk_size = Column(Integer)
+    disk_size = Column(Integer, default=0)
```
popaula937, line 255: more clear to call it `view_model_disk_size`?

framework_tvb/tvb/core/neocom/h5.py
```diff
@@ -223,8 +225,19 @@ def store_view_model(view_model, base_dir):
                     store_view_model(model_attr[idx], base_dir)
             else:
                 store_view_model(model_attr, base_dir)
+
+        if not references and isinstance(view_model, ViewModel) and not file_exist_before_save:
+            operation = dao.get_operation_by_id(os.path.basename(base_dir))
+            disk_size = get_view_model_disk_size(h5_path)
+            operation.disk_size = operation.disk_size + disk_size
+            dao.store_entity(operation)
+
```
popaula937, line 234: I would keep disk_usage computation and update in DB somewhere outside this method because we might need to use it from places that do not have access to DB.

One idea would be: after `h5.store_view_model()` is called from a service, we have on disk the operation folder with only view models inside. We could call at this point a method to compute the disk_size on the entire operation folder and then store it in DB?

framework_tvb/tvb/core/neocom/h5.py
```diff
@@ -223,8 +225,19 @@ def store_view_model(view_model, base_dir):
...
     return h5_path
 
+def get_view_model_disk_size(path):
```
popaula937, line 237: There is already a similar method:` FilesHelper.compute_size_on_disk`
Maybe it could be extended to also compute the size of an entire folder?

framework_tvb/tvb/tests/framework/core/services/project_service_test.py
```diff
@@ -296,11 +296,11 @@ def test_project_disk_size(self):
             assert '0.0 KiB' != project.disk_size_human
 
             prj_folder = self.structure_helper.get_project_folder(project)
-            actual_disk_size = self.compute_recursive_h5_disk_usage(prj_folder)[0]
+            actual_disk_size = FilesHelper.compute_recursive_h5_disk_usage(prj_folder)[0]
 
             ratio = float(actual_disk_size) / project.disk_size
             msg = "Real disk usage: %s The one recorded in the db : %s" % (actual_disk_size, project.disk_size)
-            assert ratio < 1.7, msg
+            assert ratio < 1.4, msg
```
liadomide, line 303: Is this the minimum ratio?
I wonder where from we have differences left.

framework_tvb/tvb/core/entities/file/files_helper.py
```diff
@@ -418,6 +418,23 @@ def compute_size_on_disk(file_path):
...
+        total_size = 0
+        n_files = 0
+        for dir_path, _, file_names in os.walk(start_path):
+            for f in file_names:
+                if f.endswith('.h5'):
+                    fp = os.path.join(dir_path, f)
+                    total_size += os.path.getsize(fp)
+                    n_files += 1
+        return int(round(total_size / 1024.)), n_files
+
```
liadomide, line 437: Is the return as tuple necessary here?
I could only see the first element used (I might be wrong ...)

popaula937: It is not necessary, it escaped my notice, will remove it soon

framework_tvb/tvb/tests/framework/core/services/project_service_test.py
popaula937: I don't think it is the minimum, I just reverted to the ratio we used to check. I could make it smaller.

PR #273, opened 2020-10-26
TVB-2788. Region Stimulus - click on a cube should toggle the corresponding checkbox

popaula937:
Correctly sync between left and right side. Ensure data is sent in correct order to the server for the region stimulus page.


VinczeRobert (approved):
It looks fine to me too. This is how I imagined it should work, bidirectionally. Seems like Setup Noise and Setup Region Model works the same way. Maybe in a different task we could consider adding this bidirectionality to the Large Scale Connectivity Page as well.

Also, I noted that the left side is refreshed only if you move the Mouse. I think there is a function somewhere in JS which refreshes it without needing to move the mouse, I will try to find and add it.

liadomide (approved):

PR #274, opened 2020-10-27
TVB-2785: Made TSs appear instead of DatatypeMeasures when clicking on PSE circles

VinczeRobert:


liadomide (approved):

liadomide:
Could you please test some extreme scenarios also, for example when the Measures group is being removed from the project ?
We had some bug reported in the past on some similar situation, not sure if it is still in the backlog or was solved....

VinczeRobert:
It gives an error if I remove the DatatypeMeasure group: No datatypes were generated due to simulation errors. Nothing to display.

popaula937:
What if you remove only one measure from the measures group? (should be enough change directly in DB the fk_datatype_group) The viewer should still be able to display circles for the remaining ones.

VinczeRobert:
If I only set fk_datatype_group to NULL for one DatatypeMeasureIndex, then it still displays all circles. However, if I delete one from the DB, it will throw an error at this line in pse.py:  sizes_array = [metrics[self.size_metric] for metrics in self.get_all_metrics().values()] (self.size_metric will be None).

liadomide:
I think you can merge these, and put all the missing measures comments on bug tvb-2344. It will need to be fixed (can be done separately).

PR #276, opened 2020-10-27
TVB-2754: Fix issue with Topographic Visualizer

VinczeRobert:
The positions of the ears and nose  are hardcoded (section_visualizers.css lines 178-203). I thought not to mess with the css and try to rotate the matrix of values correctly. 

Paula gave me this idea to take a connectivity and hardcode 0s to the first 38 regions and 1 to the other 1 and this is how I reproduced the issue. The 1s are red, the 0s are white. I tried it out with 4 different connectivities and every time in the region_labels attribute the first half of the values where the right hemisphere nodes and then came the ones from the left hemisphere, but then there is an ordered_labels param where it is vice-versa. So this is why I think a right rotation of the matrix would solve the problem but I'm more than 50% sure that there could be some other cases as well, which I didn't find yet.

### DIFF
--- a/framework_tvb/tvb/adapters/visualizers/topographic.py
+++ b/framework_tvb/tvb/adapters/visualizers/topographic.py
@@ -277,7 +277,8 @@ class TopographicViewer(ABCDisplayer):
         for i, array_data in enumerate(arrays):
             try:
                 data_array = TopographyCalculations.compute_topography_data(array_data, sensor_locations)
-                if h5.load_from_index(connectivities_idx[0]).region_labels[0][0] == 'r':
+                first_label = h5.load_from_index(connectivities_idx[0]).region_labels[0]
+                if first_label.lower().startswith('r'):
                     data_array = numpy.rot90(data_array, k=1, axes=(0, 1))
                 else:
                     data_array = numpy.rot90(data_array, k=-1, axes=(0, 1))
--- a/framework_tvb/tvb/adapters/visualizers/topographic.py
+++ b/framework_tvb/tvb/adapters/visualizers/topographic.py
@@ -277,7 +277,10 @@ class TopographicViewer(ABCDisplayer):
         for i, array_data in enumerate(arrays):
             try:
                 data_array = TopographyCalculations.compute_topography_data(array_data, sensor_locations)
-                data_array = numpy.rot90(data_array, k=1, axes=(0, 1))
+                if h5.load_from_index(connectivities_idx[0]).region_labels[0][0] == 'r':
+                    data_array = numpy.rot90(data_array, k=1, axes=(0, 1))
+                else:
+                    data_array = numpy.rot90(data_array, k=-1, axes=(0, 1))
                 if numpy.any(numpy.isnan(array_data)):
                     titles[i] = titles[i] + " - Topography contains nan"
                 if not numpy.any(array_data):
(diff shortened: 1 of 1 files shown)


VinczeRobert:
If there exists some connectivities for which in the ordered labels array the first half are the right nodes and then the left, we could add an if-else there to rotate to the left or right, I just don't know if there are such connectivities.

framework_tvb/tvb/adapters/visualizers/topographic.py
```diff
@@ -277,6 +277,11 @@ def launch(self, view_model):
         for i, array_data in enumerate(arrays):
             try:
                 data_array = TopographyCalculations.compute_topography_data(array_data, sensor_locations)
+                first_label = h5.load_from_index(connectivities_idx[0]).region_labels[0]
```
liadomide, line 280: Isn't it better to use connectivity.hemisphere array of Booleans instead of hard coding the Letter R?

framework_tvb/tvb/adapters/visualizers/topographic.py
```diff
@@ -277,6 +277,11 @@ def launch(self, view_model):
...
+                if first_label.lower().startswith('r'):
+                    data_array = numpy.rot90(data_array, k=1, axes=(0, 1))
+                else:
+                    data_array = numpy.rot90(data_array, k=-1, axes=(0, 1))
```
liadomide, line 284: The rotation seems to rely on the first connectivity only, but we could have different connectivities for the different arrays to display, isn't it?

liadomide:
I think this should be improved (I left two small comments)

framework_tvb/tvb/adapters/visualizers/topographic.py
VinczeRobert: Hmm. Can  you please tell me how to generate an FCD with more connectivities? And then would it be enough just to add connectivities_idx[i] instead of connectivities_idx[0]?

popaula937: I think Lia refers to the viewer itself which is capable of displaying more than one ConnMeaures at once. Good point, I missed the connectivities_idx[0] there.

liadomide: Yes, I believe using connectivities_idx[i] would solve it.

To reproduce, you will need to have more than one Connectivity in your project, then produce ConnMeasures for both, and in the Topo viewer simply choose ConnMeasures in a mixed combination (linked with both the initial Connectivity dts).

VinczeRobert: I tried a mixed combination, but the code says it is not possible and that we should choose a single connectivity for displaying. Please check lines 267-268 in topographic.py. Also, just removing that condition will produce further errors.

I suggest we leave this as is it now, maybe open a new task if we want to make the display work for different connectivities.

liadomide: Thanks for investigating. If only one conn can be chosen for all inputs, then it is ok for the code to remain unchanged. Pls just comment the code why conn _idx[0] is ok

framework_tvb/tvb/adapters/visualizers/topographic.py
liadomide: How about this improvement?

PR #277, opened 2020-10-28
TVB-2761: Export/Import of DT with related entities

adrian-dordea:
Implement exporting DT with references from DT tab

### DIFF
--- a/framework_tvb/tvb/adapters/exporters/export_manager.py
+++ b/framework_tvb/tvb/adapters/exporters/export_manager.py
@@ -94,52 +94,30 @@ class ExportManager(object):
 
         return results
 
-    def export_data_with_references(self, dt_list, project, exporter):
-        files_helper = FilesHelper()
-        considered_op_ids = []
-        to_be_exported_folders = []
+    def get_export_data_zip_path(self, data, exporter):
+        zip_file_name = exporter.get_export_file_name(data)
+        zip_file_name = zip_file_name.replace('.h5', '.zip')
+        return os.path.join(self.export_folder, zip_file_name)
 
-        data_export_folder = self._build_data_export_folder(dt_list[0])
-        name = exporter.get_export_file_name(dt_list[0], False)
+    def export_data_with_references(self, export_data_zip_path, data_export_folder):
+        with TvbZip(export_data_zip_path, "w") as zip_file:
+            for filename in os.listdir(data_export_folder):
+                zip_file.write(os.path.join(data_export_folder, filename), filename)
 
-        zip_file_name = "%s.%s" % (name, self.ZIP_FILE_EXTENSION)
-        result_path = os.path.join(data_export_folder, zip_file_name)
+        return None, export_data_zip_path, True
 
-        for dt in dt_list:
-            op_id = dt.fk_from_operation
-            if op_id not in considered_op_ids:
-                to_be_exported_folders.append({'folder': files_helper.get_project_folder(project, str(op_id)),
-                                               'archive_path_prefix': str(op_id) + os.sep,
-                                               'exclude': ''})
-                considered_op_ids.append(op_id)
-
-        with TvbZip(result_path, "w") as zip_file:
-            # Pack project [filtered] content into a ZIP file:
-            self.logger.debug("Done preparing, now we will write folders " + str(len(to_be_exported_folders)))
-            self.logger.debug(str(to_be_exported_folders))
-            for pack in to_be_exported_folders:
-                zip_file.write_folder(**pack)
-            self.logger.debug("Done exporting files, now we will export linked DTs")
-            self.logger.debug("Done, closing")
-
-        return None, result_path, name
-
-    def get_data_with_references_list(self, data, text_dict=None):
+    def copy_dt_to_export_folder(self, data, data_export_folder):
         data_path = h5.path_for_stored_index(data)
-
-        if not text_dict:
-            text_dict = dict()
-        if data.gid and data.gid not in text_dict.keys():
-            text_dict[data.gid] = data
-
         with H5File.from_file(data_path) as f:
+            file_destination = os.path.join(data_export_folder, os.path.basename(data_path))
+            if not os.path.exists(file_destination):
+                FilesHelper().copy_file(data_path, file_destination)
             sub_dt_refs = f.gather_references()
 
             for reference in sub_dt_refs:
                 if reference[1]:
                     dt = dao.get_datatype_by_gid(reference[1].hex)
-                    self.get_data_with_references_list(dt, text_dict)
-        return list(text_dict.values())
+                    self.copy_dt_to_export_folder(dt, data_export_folder)
 
     def export_data(self, data, exporter_id, project):
         """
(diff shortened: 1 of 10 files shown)


framework_tvb/tvb/adapters/exporters/export_manager.py
```diff
@@ -46,6 +46,7 @@
 from tvb.core.entities.model.model_operation import STATUS_ERROR
 from tvb.core.entities.storage import dao
 from tvb.core.neocom import h5
+from tvb.core.neotraits._h5core import H5File
```
liadomide, line 49: We should not import private modules (name starting with _)but use the main api

framework_tvb/tvb/adapters/exporters/export_manager.py
```diff
@@ -124,19 +167,23 @@ def export_data(self, data, exporter_id, project):
...
+            # This will imply to generate a folder which is unique for each export
+            data_export_folder = None
+            try:
+                data_export_folder = self._build_data_export_folder(data)
+
+                # Here is the real export
+                self.logger.debug("Start export of data: %s" % data.type)
+                export_data = exporter.export(data, data_export_folder, project)
+            finally:
+                # In case export did not generated any file delete folder
```
liadomide, line 184: Maybe this part can be enhanced and both scenarios (single file or zip refences) have a try/finally?

framework_tvb/tvb/core/services/import_service.py
```diff
@@ -91,6 +91,10 @@ def is_old_form(self):
     def order_field(self):
         return self.operation.create_date if (self.operation is not None) else datetime.now()
 
+    @property
+    def order_field_by_folder(self):
+        return int(os.path.basename(self.operation_folder)) if (self.operation_folder is not None) else datetime.now()
+
```
liadomide, line 97: This conversion to int is not reliable. We might have links, for example.
Can't we continue to order by operation create date?

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
```diff
@@ -305,23 +305,25 @@ def execute_post(self, project_id, submit_url, step_key, algorithm, **data):
...
-                adapter_instance.user_id = common.get_logged_user().id
-                result = adapter_instance.launch(view_model)
-                if isinstance(result, dict):
-                    return result
-                else:
-                    common.set_error_message("Invalid result returned from Displayer! Dictionary is expected!")
-                return {}
-
-            result = self.operation_services.fire_operation(adapter_instance, common.get_logged_user(),
+            dts_imported = self.operation_services.import_dts(view_model, project_id)
```
liadomide, line 308: This is a very odd change. Why? The export was functioning before, can't we reuse the mechanism?

framework_tvb/tvb/interfaces/web/controllers/project/project_controller.py
```diff
@@ -705,7 +705,10 @@ def downloaddata(self, data_gid, export_module):
             self.mark_file_for_delete(file_path, True)
 
         self.logger.debug("Data exported in file: " + str(file_path))
-        return serve_file(file_path, "application/x-download", "attachment", file_name)
+        if file_name:
+            return serve_file(file_path, "application/x-download", "attachment", file_name)
+        else:
+            return serve_file(file_path, "application/x-download", "attachment")
```
liadomide, line 711: Wouldn't be better if we always provide a file name?

framework_tvb/tvb/core/services/operation_service.py
```diff
@@ -486,6 +488,27 @@ def __expand_arguments(arguments_list, range_values, range_title):
                 result.append((kw_new, range_new))
         return result
 
+    def import_dts(self, view_model, project_id):
```
liadomide, line 491: Can't we work with the current adapters mechanism? This method seems to break the patterns

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
adrian-dordea: In order to test what happens on the mechanism before, the line in question can be commented and dts_imported given the value False. The change was necessary because if we try with the old version a additional empty TVB importer operation will be created dew to the flow(this could not be ommited whit the old flow, tested and researched during debug), also the old flow seems to ommit the time information on the Data Structure page in UI.

framework_tvb/tvb/core/services/operation_service.py
adrian-dordea: this questions answer is related to the answer to the above question about: 
"dts_imported = self.operation_services.import_dts(view_model, project_id)"

framework_tvb/tvb/core/services/import_service.py
adrian-dordea: I think I created some faulty zip data, but after further testing the initial solution is the wanted one.

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
liadomide: A TVBImport operation will be needed, right? This is what we execute.

ViewModel files should not be exported in this flow, so all linked DTs will not have any operation information with them, thus at import time, we execute a TVBImport.

liadomide (changes requested):
I think the current implementation is not correctly reusing previous export/import mechanism. If we miss params there, we should add some, or adjust that code, instead of creating another completely different flow

framework_tvb/tvb/interfaces/web/controllers/flow_controller.py
adrian-dordea: The export is not the problem here, the import is...when we uploas a file in TVB HDF5/ZIP importer overlay the first thing that is created is a TVBImporterModel and it creates a operation for that model, and then the flow continues as in the case of the import project creating a new operation for each folder in the zip file we want to import -> leaving us with the first operation folder containing only the TvbImporterModel and the rest are imported normaly...

liadomide: I think the "imported normally like at import project" is the problem here.
Why not export the dts as a zip with no extra folders. Then import with a tVBimporter operation created.

adrian-dordea: I tested with the mentioned solution for the export (all h5 in the zip file no subfolder), and with the current implementation of the TVB HDF5/ZIP importer, we still have the problem that a extra operation with only TVBImporterModel will be first created, and then all the other h5's will be imported in the same operation... tested with zip containing conn, surface and region mapping, the result looked a bit unusual/wierd to me in the operations dashboard... is this what we want? or separate operations for each file in the zip?

liadomide: I think just one operation will be ok.
So it is left to put all imported dts to the first operation that you mention (the one that you Say it is now empty)

VinczeRobert (commented):
If I understood correctly, then we want to export DT with related entities as a ZIP folder, and then the mirrored import would be a TVB Import of the ZIP. 

I tested the export of a several datatypes which had one or more related entities and they are all exported correctly, I just noticed one thing, not necessarily related to this task: if I want to import a ZIP with related entities and they exist, I don't get a suggestive error message like I get with other imports, instead.

framework_tvb/tvb/adapters/exporters/export_manager.py
```diff
@@ -130,9 +155,14 @@ def export_data(self, data, exporter_id, project):
         try:
             data_export_folder = self._build_data_export_folder(data)
 
-            # Here is the real export                    
-            self.logger.debug("Start export of data: %s" % data.type)
-            export_data = exporter.export(data, data_export_folder, project)
+            self.copy_dt_to_export_folder(data, data_export_folder)
+            if data_export_folder is not None and len(os.listdir(data_export_folder)) > 1:
```
liadomide, line 159: Instead of this check, wound't be better to have a new button in web GUI "Export with references"  to differentiate between the 2 scenarios ?
In case a user wants to export a TimeSeries and only use it in Matlab, for example, a simple export (old form) could work better

liadomide:
> if I want to import a ZIP with related entities and they exist, I don't get a suggestive error message like I get with other imports, instead.

In case some entities exist, we should not end in error but rather pass over, as the reference can be reused.
In case the entity with the same GUID exists in another project, we should create a link in the current project.

popaula937:
Import only works if the datatypes have been deleted before. It should not fail in case datatypes already exist, but pass over them.
This does not handle the DataTypeGroup entities, it throws an exception when you try to export them. 
Please take another look.

VinczeRobert:
This PR needs a rebase.

PR #278, opened 2020-10-28
TVB-2553 Fix integrator configuration in the Phase Plane page.

bvalean:


popaula937 (approved):
Tested this state in several scenarios and it seems to draw the same trajectories as 1.5.* now. 
We believe the big difference was that we have lost the reset of RandomStream which was always done on the `Noise.__init__`, while now it is done only if it is `None`. I suspect that In the old input tree, the Noise was always re-instantiated when a param changed, while now, we only re-inistantiate it when another type of Noise is chosen, and when the params change, we just update them on the current Noise kept in session.

PR #280, opened 2020-10-29
TVB-2794: Don't display question marks inside of the brain submenu

VinczeRobert:


framework_tvb/tvb/interfaces/web/static/js/genericTVB.js
```diff
@@ -1137,7 +1137,11 @@ let activeMenu = null;
 function openMenu(selector) {
     hideMenus();
     activeMenu = $(selector);
+    extension =  activeMenu.find('.extension')
     activeMenu.find('.extension').show();
+    if(extension.length > 1) {
+        $(".helper").remove();
```
liadomide, line 1143: Could you please explain what is intended to happen here?
Is it really to remote the green bubbles from some places? If so, why? Task tvb-2794 suggests to show these more, not less.

popaula937: The intention was to hide the green bubbles in the upper-right menu, as we did in the past.
Still, not all green bubbles were shown (eg: on the simulator page).

I did a refactor to add all of them and defined display rules on the server-side to take into account both isCallout (upper-right menu) and showOnlineHelp which the user can choose from GUI.
@VinczeRobert pls test this is complete.

liadomide: I think the help not being displayed in the upper right callout was a bug in version 1.5.*
Would there be a problem if we display help there also (when help is enabled for the curent user)?

popaula937: In the upper right callout, it looks like some CSS classes change the design of the help and also displays all of them by default. We thought that is the reason why the help was hidden, but we could investigate a bit more.

VinczeRobert: More deeply, the problem is the following: both that brain submenu and the help question marks have the html class "extension". The code in JS is built in such way that it expects only one element with that class to exist when you click on it, so when you click on the brain, it takes all elements with "extension" class and it displays them. I tried for some time to change that behaviour yesterday but I couldn't do it with small fixes and I thought it was not worth to do major changes there just because of this.

@popaula937 there is a problem with your solution as I tested it: clicking on the question marks from the Simulation Page don't display the help text.

PR #286, opened 2020-11-06
TVB-2523 Prepare MAC distribution

bvalean:


popaula937:
Haven't looked yet at the changes, but tested the distribution and it works on my machine. Will take a look soon.

popaula937 (commented):
This looks good, I have tested it also on my older MacOS version.
Few remarks that used to work in the older distribution:
- jupyter_notebook.command does not work for me: `Error executing Jupyter command 'notebook': [Errno 2] No such file or directory`
- tvb_start.command does not show the tvb icon in the app bar
- tvb_stop.command stops the process but does not remove the tvb icon from the app bar

popaula937 (approved):

PR #287, opened 2020-11-06
Issue encountered using stimulator on a neural fields model

BalanceKey:
During the configuration of the simulator, the region mapping that the stimulator uses does not take into account the subcortical regions. Indeed, this is due to the fact that the stimulator uses a region mapping provided by the surface data, which only contains the cortical areas. This issue can be solved easily by specifying explicitly the region mapping the stimulator should use, but this is not user-friendly.


BalanceKey:
@maedoc could you review this ?

maedoc:
the ratio of test lines increased to code lines increased is fantastic here.  The fix looks fine but  I'm a little surprised the test suite didn't run here on github though.. I will check this out first to see why.

maedoc (approved):

PR #289, opened 2020-11-09
TVB-2792: Solve some problems regarding PSE Viewers

VinczeRobert:
I explained why I did what I did more detailed here: https://req.thevirtualbrain.org/browse/TVB-2792

Shortly:
1) I don't think it makes sence to launch a PSE with only one operation. Neither Visualizer will work on it. To solve this, I chose to not launch a PSE simulation and display a message from JS.
2) There are some JS errors in the Isocline Viewer when one of the two range params have only one value. I chose to add an extra param in the database which will tell us wether we display the Isocline Viewer or not.

If the solution gets accepted, I will extend the migration code with this extra db param as well.

### DIFF
--- a/framework_tvb/tvb/interfaces/web/static/js/bursts.js
+++ b/framework_tvb/tvb/interfaces/web/static/js/bursts.js
@@ -382,7 +382,8 @@ function _computeRangeNumberForParamPrefix(prefix){
         return _getRangeValueForGuidParameter(pse_param_guid);
     }
 
-    return 1;
+    // We don't have a value chosen for this range param
+    return 0;
 }
 
 function _displayPseSimulationMessage() {
@@ -393,6 +394,19 @@ function _displayPseSimulationMessage() {
     pse_param2_number = _computeRangeNumberForParamPrefix('pse_param2');
 
     let nrOps = pse_param1_number * pse_param2_number;
+
+    // Only the first range been chosen
+    if(nrOps == 0){
+        nrOps = pse_param1_number;
+    }else{
+        if(pse_param1_number == 1 || pse_param2_number == 1){
+            message = "Can't launch PSE when one of the parameters has only one value," +
+                " instead of a range of values!";
+            displayMessage(message, "errorMessage");
+            throw message;
+        }
+    }
+
     let className = "infoMessage";
 
     if (nrOps > THREASHOLD_WARNING) {
@@ -404,6 +418,10 @@ function _displayPseSimulationMessage() {
     if (nrOps > 1) {
         // Unless greater than 1, it is not a range, so do not display a possible confusing message.
         displayMessage("Range configuration: " + nrOps + " operations.", className);
+    }else{
+        message = "Can't launch PSE with only one  operation!"
+        displayMessage(message, "errorMessage");
+        throw message;
     }
 }
 


framework_tvb/tvb/interfaces/web/static/js/bursts.js
```diff
@@ -404,6 +404,10 @@ function _displayPseSimulationMessage() {
     if (nrOps > 1) {
         // Unless greater than 1, it is not a range, so do not display a possible confusing message.
         displayMessage("Range configuration: " + nrOps + " operations.", className);
+    }else{
+        message = "Can't launch PSE with only operation!"
```
liadomide, line 408: ... With only ONE operation ...

framework_tvb/tvb/adapters/visualizers/pse_isocline.py
```diff
@@ -129,7 +129,8 @@ def get_input_name():
 
     @staticmethod
     def get_filters():
-        return FilterChain(fields=[FilterChain.datatype + ".no_of_ranges"], operations=["=="], values=[2])
+        return FilterChain(fields=[FilterChain.datatype + ".no_of_ranges", FilterChain.datatype + ".display_isocline"],
+                           operations=["==", "=="], values=[2, 1])
```
liadomide, line 133: It seems to me like a double negation here... in this situation in fact we do not have a 2D range, only 1D. 
I would not add the extra parameter.
How about checking the ranges before submit to execution? When one of the two (range-param-1 or range-param-2) range choices returns a single value, cut it as a range, and convert it into a static value. I am not convinced when this should happen. It looks like a validation process, so maybe it belongs in the controllers layer?

VinczeRobert: Why would it be a double negation? There are cases when the first condition can be true, but the second one false.

I am not sure I understand this:  "cut it as a range, and convert it into a static value". Does that mean I should set it to one value and let the no_of_ranges be 1? Otherwise without my new condition the Isocline Viewer would still show up.

framework_tvb/tvb/interfaces/web/static/js/bursts.js
VinczeRobert: Oops

framework_tvb/tvb/adapters/visualizers/pse_isocline.py
popaula937: I agree it would be better to have a stronger validation on the ranges (and ensure that no single value range is generated instead of keeping extra fields in DB). Currently, we just do the `fill_from_post` which prepares the `RangeParameter` object and passes it along to the execution, without checking the actual values.
Maybe there we could also check in there how many values would each range have and in case there is only one, don't create the `RangeParameter` object, but directly set the single value on the session simulator? I think we have enough information to do that.

popaula937: > Why would it be a double negation? There are cases when the first condition can be true, but the second one false.
> 
> I am not sure I understand this: "cut it as a range, and convert it into a static value". Does that mean I should set it to one value and let the no_of_ranges be 1? Otherwise without my new condition the Isocline Viewer would still show up.

Yes, I think that is what Lia meant.

popaula937 (approved):

PR #290, opened 2020-11-09
TVB-2538: Enhance Branch simulation use-case

adrian-dordea:
move branch button next to copy + creating new endpoint and make sure it works as current branch implementation

### DIFF
--- a/framework_tvb/tvb/interfaces/web/controllers/common.py
+++ b/framework_tvb/tvb/interfaces/web/controllers/common.py
@@ -87,7 +87,7 @@ KEY_LAST_LOADED_FORM_URL = 'last_loaded_form_url'
 KEY_BACK_PAGE = "back_page_link"
 KEY_SECTION_TITLES = "section_titles"
 KEY_SUBSECTION_TITLES = "sub_section_titles"
-KEY_BRANCH = "is_branch"
+KEY_IS_SIMULATOR_BRANCH = "is_branch"
 
 # Overlay specific keys
 KEY_OVERLAY_TITLE = "overlay_title"
--- a/framework_tvb/tvb/interfaces/web/controllers/common.py
+++ b/framework_tvb/tvb/interfaces/web/controllers/common.py
@@ -88,7 +88,6 @@ KEY_BACK_PAGE = "back_page_link"
 KEY_SECTION_TITLES = "section_titles"
 KEY_SUBSECTION_TITLES = "sub_section_titles"
 KEY_IS_SIMULATOR_BRANCH = "is_branch"
-KEY_NUMBER_OF_REGIONS = "number_of_regions"
 
 # Overlay specific keys
 KEY_OVERLAY_TITLE = "overlay_title"
--- a/framework_tvb/tvb/interfaces/web/controllers/common.py
+++ b/framework_tvb/tvb/interfaces/web/controllers/common.py
@@ -87,6 +87,7 @@ KEY_LAST_LOADED_FORM_URL = 'last_loaded_form_url'
 KEY_BACK_PAGE = "back_page_link"
 KEY_SECTION_TITLES = "section_titles"
 KEY_SUBSECTION_TITLES = "sub_section_titles"
+KEY_BRANCH = "is_branch"
 
 # Overlay specific keys
 KEY_OVERLAY_TITLE = "overlay_title"
--- a/framework_tvb/tvb/interfaces/web/controllers/common.py
+++ b/framework_tvb/tvb/interfaces/web/controllers/common.py
@@ -88,6 +88,7 @@ KEY_BACK_PAGE = "back_page_link"
 KEY_SECTION_TITLES = "section_titles"
 KEY_SUBSECTION_TITLES = "sub_section_titles"
 KEY_IS_SIMULATOR_BRANCH = "is_branch"
+KEY_NUMBER_OF_REGIONS = "number_of_regions"
 
 # Overlay specific keys
 KEY_OVERLAY_TITLE = "overlay_title"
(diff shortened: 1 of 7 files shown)


framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -1133,3 +1147,10 @@ def load_simulator_configuration_from_zip(self, **data):
             common.set_warning_message(excep.message)
 
         raise cherrypy.HTTPRedirect('/burst/')
+
+    @staticmethod
+    def srt_to_bool(str):
```
popaula937, line 1152: This is not used anymore, right?

framework_tvb/tvb/interfaces/web/controllers/common.py
```diff
@@ -87,6 +87,7 @@
 KEY_BACK_PAGE = "back_page_link"
 KEY_SECTION_TITLES = "section_titles"
 KEY_SUBSECTION_TITLES = "sub_section_titles"
+KEY_BRANCH = "is_branch"
```
popaula937, line 90: maybe keep consistency between names here: KEY_IS_SIMULATOR_BRANCH?

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -204,7 +204,8 @@ def hide_launch_and_setup_pse_button(self):
 
     @property
     def include_branch_button(self):
-        if self.is_launch_fragment and self.is_simulation_copy and not self.is_pse_launch:
+        is_branch = common.get_from_session(common.KEY_BRANCH)
```
popaula937, line 207: Let's not link `SimulatorFragmentRenderingRules` object with the session, send the `is_branch` flag as a parameter to keep consistency. We might refactor this part soon, so it will be easier to keep it separate.

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -204,7 +204,8 @@ def hide_launch_and_setup_pse_button(self):
...
+        if is_branch and self.is_launch_fragment and self.is_simulation_copy and not self.is_pse_launch:
```
popaula937, line 208: I think this condition can be simplified now, as now we want to show the branch button only when `is_branch` and `is_launch_fragment`  are true, we don't want it on the 'copy' page anymore

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -1030,6 +1031,23 @@ def load_burst_read_only(self, burst_config_id):
...
+    @expose_fragment('simulator_fragment')
+    def branch_simulator_configuration(self, burst_config_id):
+        form = self.get_first_fragment(burst_config_id)
+        common.add2session(common.KEY_BRANCH, True)
+        rendering_rules = SimulatorFragmentRenderingRules(form, SimulatorWizzardURLs.SET_CONNECTIVITY_URL,
+                                                          is_simulation_copy=True, is_simulation_readonly_load=True,
+                                                          is_first_fragment=True)
+        return rendering_rules.to_dict()
+
+    def get_first_fragment(self, burst_config_id):
```
popaula937, line 1050: this can be private and maybe use a more specific name to what its intention is. Also, the line         `common.add2session(common.KEY_IS_SIMULATOR_COPY, True)` is not common for both copy&branch methods

popaula937:
Clicking on `Previous` looks close to what we planned but mind that the `Next` button for disabled fragments does not work because no data is sent to the server for disabled fields. 
Also, if you click Next on an enabled fragment (eg: Integration-step size) it will show the next fragment (eg: Monitors) as enabled, but it should remain disabled.

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -465,6 +477,9 @@ def set_stimulus(self, **data):
                                                           is_simulator_load, self.last_loaded_form_url,
                                                           cherrypy.request.method, is_model_fragment=True,
                                                           is_surface_simulation=session_stored_simulator.is_surface_simulation)
+        is_branch = common.get_from_session(common.KEY_IS_SIMULATOR_BRANCH)
+        if is_branch:
+            rendering_rules.enable_branch_class = "enable_branch"
```
popaula937, line 482: Why do this if block everywhere? You could keep 'enable_branch' class directly in the html template and add it to the fieldset only when rendering_rules.is_branch is true

framework_tvb/tvb/interfaces/web/controllers/common.py
```diff
@@ -87,6 +87,8 @@
 KEY_BACK_PAGE = "back_page_link"
 KEY_SECTION_TITLES = "section_titles"
 KEY_SUBSECTION_TITLES = "sub_section_titles"
+KEY_IS_SIMULATOR_BRANCH = "is_branch"
+KEY_NUMBER_OF_REGIONS = "number_of_regions"
```
popaula937, line 91: Is this used?

framework_tvb/tvb/interfaces/web/controllers/simulator_controller.py
```diff
@@ -313,21 +318,23 @@ def index(self):
...
                                                           last_form_url=self.last_loaded_form_url,
                                                           last_request_type=cherrypy.request.method,
-                                                          is_first_fragment=True)
-
+                                                          is_first_fragment=True, is_branch=is_branch)
         template_specification.update(**rendering_rules.to_dict())
         return self.fill_default_attributes(template_specification)
 
-    def prepare_first_fragment(self):
+    def prepare_first_fragment(self, burst_config=None):
```
popaula937, line 332: burst_config is already in session at this point, right? Then, there is no need to pass it as an argument here, just take it from the session.

popaula937:
The code changes look good to me now, I left just 2 small comments. But, after testing it seems that this part is not yet fixed:

> Also, if you click Next on an enabled fragment (eg: Integration-step size) it will show the next fragment (eg: Monitors) as enabled, but it should remain disabled.

Is all the code pushed?

popaula937 (approved):

PR #293, opened 2020-11-16
TVB-2834 Import tumor data from ebrains dataset

bvalean:
Import TVB Tumor Data demos

### DIFF
--- a/framework_tvb/tvb/core/services/operation_service.py
+++ b/framework_tvb/tvb/core/services/operation_service.py
@@ -119,7 +119,8 @@ class OperationService:
             return self._send_to_cluster(operations, adapter_instance, current_user.username)
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


liadomide (approved):
As far as i can judge this looks ok

popaula937:
Should we import also the surface and region mapping from tvb_data/berlinSubjects from within the script?

popaula937:
Maybe we should not introduce in the command API the possibility to receive dictionaries that have to be transformed to view_models. I believe preparing a dict (ensuring all attributes are written correctly as strings) is not easier than preparing directly the view_model object.

bvalean:
> Should we import also the surface and region mapping from tvb_data/berlinSubjects from within the script?

Did it

popaula937 (approved):

popaula937:
PR checks are failed because Zenodo is down.

PR #295, opened 2020-11-18
TVB-2705 Decouple forms from dao 

bvalean:


liadomide (approved):
From a top architecture pov these changes look ok.

popaula937:
I am wondering whether it would be worth adding another layer to avoid the need of making 2 calls (instantiate form and fill the datatype select fields), something like a FormBuilder.

popaula937 (commented):
@bvalean I rebased this branch and extracted the project_id dependency out of the forms, it is now used only at the alg_service level. What do you think? 
I will look a bit later at the alg_service changes in more detail.

PR #296, opened 2020-11-23
TVB-2786: Review metadata of Timeseries generated from PSE

adrian-dordea:


popaula937:
After looking over the flow, it does not make much sense to me to pass the operation_group name on the operation user_tag_3 here as it will always have the value 'incomplete' and we end up with the 'incomplete' value on both operation and resulted index. We don't seem to be using it. And we already have operation.status in order to inform us about the current operation status.

liadomide:
> After looking over the flow, it does not make much sense to me to pass the operation_group name on the operation user_tag_3 here as it will always have the value 'incomplete' and we end up with the 'incomplete' value on both operation and resulted index. We don't seem to be using it. And we already have operation.status in order to inform us about the current operation status. 

It used to be the range param names on that tag, nothing regarding "incomplete". It was something like "Pse of coupling x conductionspeed"

popaula937:
> > After looking over the flow, it does not make much sense to me to pass the operation_group name on the operation user_tag_3 here as it will always have the value 'incomplete' and we end up with the 'incomplete' value on both operation and resulted index. We don't seem to be using it. And we already have operation.status in order to inform us about the current operation status.
> 
> It used to be the range param names on that tag, nothing regarding "incomplete". It was something like "Pse of coupling x conductionspeed"

Then, maybe it should be set in `ABCAdapter._capture_operation_results` since there is where the `OperationGroup.name` is currently computed?

PR #301, opened 2020-11-26
TVB-2670: Review Code in Simulator/Burst Controllers and Simulator/Burst Services

VinczeRobert:


VinczeRobert:
I know there is still too much code in the SimulatorController. The problem with moving a lot of code parts in the service layer is that there are a lot of lines of code that use the session and those should stay in the Controller.

For example in each set method there is a part of code that looks somehow like this:

       ```
            self._update_last_loaded_fragment_url(SimulatorWizzardURLs.SET_COUPLING_PARAMS_URL)
            form.fill_from_post(data)
            self.simulator_service.reset_at_connectivity_change(is_simulator_copy, form, session_stored_simulator)
            form.fill_trait(session_stored_simulator)

next_form = get_form_for_coupling(type(session_stored_simulator.coupling))()
next_form.fill_from_trait(session_stored_simulator.coupling)
```
Structures like this could be pushed in the service by adding methods of like 4/5 parameters, but the problem is that self._update_last_loaded_fragment that is used everywhere. Alternatively, we could add a bunch of small methods that only have like 2 or 3 lines of code with the parts that come after _update_last_loaded_fragment_url.

Also, what about moving the SimulationFragmentRenderingRules to a new file? That would reduce the dimension of this file by ~150 lines of code.

**Comment by VinczeRobert** [comment]
I think I managed to move most of the things that where possible to the simulator/burst service.
I would appreaciate further suggestions as I seem to be out of ideas.

There have been a few chunks of code that I hesitated to move, like for example in set_monitors everything that comes after cherrypy.request.method == POST_REQUEST. If I were to move that part of code, then I have to move prepare_final_fragment as well and that method uses cherrypy ("cherrypy.request.method != POST_REQUEST"). I think cherrypy must remain at the controller level, so maybe we can send cherrypy.request.method as a parameter to the service? In that case there is still some code that I could move.

Also, I believe that there are a lot of methods with very similar names, so maybe I could make some names more clear, even if they end up to be longer.

**Comment by liadomide** [comment]
Indeed: do not move cheerypy stuff into the service, not even as params.

I would support more explicit method names, even if a bit longer, but try not to be extremeeely long.

Further suggestion: what is the logic separation between burst and simulator services? Which does what? At some point duplicate logic existed between these two.

**Comment by VinczeRobert** [comment]
The simulator service has the two big methods for launching simulations (async_launch_and_prepare_simulation and async_launch_simulation_on_server) and other than those two, the burst service is not used there. The other methods do not use the burst service (for preparing different fragments of the Simulation Page, building urls etc).

The burst service is only related to bursts.

We talked with Paula that we could even merge the two services into one, but then again we would have a another pretty big file with too many methods, so I think it is better to keep them like this even if they are not 100% separated.

**Comment by VinczeRobert** [comment]
There is a TO-DO in the prepare_metrics_operation method of the BurstService class. I tried moving that code to OperationService to somehow merge it with _prepare_metric_operation, but it is not possible. The reason is that you can't import OperationService from hpc_scheduler_client, the place where prepare_metrics_operation is used (circular dependency I think).

File: framework_tvb/tvb/interfaces/web/entities/context_simulator.py [hunk 1]
```diff
@@ -0,0 +1,138 @@
...
+"""
+
+from tvb.core.entities.file.simulator.view_model import SimulatorAdapterModel
+from tvb.core.entities.model.model_burst import BurstConfiguration
+from tvb.interfaces.web.controllers import common
+
+
+class SimulatorContext(object):
+    def __init__(self):
+        self.session_stored_simulator = None
```
popaula937, line 42: Why keep these attributes also on SimContext if we already keep them in session?

framework_tvb/tvb/interfaces/web/entities/context_simulator.py
```diff
@@ -0,0 +1,138 @@
...
+    def init_session_at_sim_config_from_zip(self, burst_config, simulator):
+        self.burst_config = burst_config
+        self.session_stored_simulator = simulator
+        self.is_simulator_copy = True
+        self.is_simulator_load = False
+
+        common.add2session(common.KEY_BURST_CONFIG, burst_config)
+        common.add2session(common.KEY_SIMULATOR_CONFIG, simulator)
+        common.add2session(common.KEY_IS_SIMULATOR_COPY, True)
+        common.add2session(common.KEY_IS_SIMULATOR_LOAD, False)
```
popaula937, line 134: reset also the branch flag, make sure you set all the flags correctly in other methods, too

framework_tvb/tvb/interfaces/web/entities/context_simulator.py
VinczeRobert: I thought the point was to keep all session related parameters on the context to not have to get them from the session in each method where they  are used.  So I am a little confused now about how I should change the code.

framework_tvb/tvb/interfaces/web/entities/context_simulator.py
VinczeRobert: I tested the functionality and I didn't find any problems so far but I will test again when there are no more changes to be made.

framework_tvb/tvb/interfaces/web/entities/context_simulator.py
popaula937: I don't think session access is a problem, you have to do it anyway, in each endpoint. 
Introducing this SimulatorContext would improve the code by handling all the necessary flags together, in one place, instead of having to handle each flag separately in each endpoint. So, instead of having multiple lines of code that get values from the session, you would have only one that calls a method on the SimulatorContext. For me, keeping these values both on the session and on the context seems a bit much since you always have to make sure they are synchronized.
What I would expect to have here are the keys from common.py and methods that add/get those to/from the session for different use-cases.

framework_tvb/tvb/interfaces/web/entities/context_simulator.py
```diff
@@ -0,0 +1,129 @@
...
+        self.project = None
+        self.last_loaded_fragment_url = None
+
+    def add_session_stored_simulator(self, session_stored_simulator):
+        common.add2session(self.KEY_SIMULATOR_CONFIG, session_stored_simulator)
+
+    def get_session_stored_simulator(self):
+        return common.get_from_session(self.KEY_SIMULATOR_CONFIG)
+
+    def get_current_project(self):
```
bvalean, line 57: Change this method into a setter

framework_tvb/tvb/core/services/burst_service.py
```diff
@@ -33,20 +33,21 @@
 from datetime import datetime
 
 from tvb.adapters.datatypes.db.mapped_value import DatatypeMeasureIndex
-from tvb.adapters.datatypes.db.time_series import TimeSeriesIndex
-from tvb.adapters.datatypes.h5.time_series_h5 import TimeSeriesH5
```
liadomide, line 37: The remaining import from adapters.datatypes should also be avoided from the service package

framework_tvb/tvb/core/services/simulator_service.py
```diff
@@ -37,18 +37,29 @@
...
 from tvb.core.entities.storage import dao
 from tvb.core.neocom import h5
-from tvb.core.neocom.h5 import DirLoader
+from tvb.core.services.algorithm_service import AlgorithmService
 from tvb.core.services.burst_service import BurstService
-from tvb.core.services.import_service import ImportService
+from tvb.core.services.exceptions import BurstServiceException
 from tvb.core.services.operation_service import OperationService
+from tvb.interfaces.web.controllers.simulator.simulator_wizzard_urls import SimulatorWizzardURLs
 from tvb.simulator.integrators import IntegratorStochastic
```
liadomide, line 52: Dependencies from adapters and controllers here are bad. Please reconsider

framework_tvb/tvb/core/services/simulator_service.py
```diff
@@ -226,13 +237,200 @@ def async_launch_and_prepare_pse(self, burst_config, user, project, simulator_al
...
+        equation_form.equation.data = simulator.integrator.noise.b.equation
+        equation_form.fill_from_trait(simulator.integrator.noise.b)
+
+        rendering_rules.form = equation_form
+        rendering_rules.form_action_url = SimulatorWizzardURLs.SET_NOISE_EQUATION_PARAMS_URL
+        return rendering_rules.to_dict()
+
+    @staticmethod
+    def prepare_monitor_fragment(simulator, rendering_rules, form_action_url):
+        monitor_fragment = SimulatorMonitorFragment(simulator.is_surface_simulation)
```
liadomide, line 302: I see "fragments" related code moved here. It seems a bit counter intuitive, as it is GUI related code, which should not really be in the service layer. I suggest code near the simulator fragments classes, near the simulator form classes or in the controller

framework_tvb/tvb/core/services/simulator_service.py
```diff
@@ -226,13 +237,200 @@ def async_launch_and_prepare_pse(self, burst_config, user, project, simulator_al
...
+
+    def build_list_of_monitors_from_view_models(self, monitor_view_models):
+        self.monitors_dict = dict()
+        count = 0
+        for monitor_model in monitor_view_models:
+            self.monitors_dict[monitor_model.__class__.__name__] = (monitor_model, count + 1)
+            count = count + 1
+
+    @staticmethod
+    def build_monitor_url(fragment_url, monitor):
```
liadomide, line 329: URL build in the service does not seem right. Should be in the controllers, because there we have the code responding to such urls, isn't it?

liadomide (changes requested):
I see good changes here (e.g. the new contex class) but also some regressions or left over imports with the wrong dependencies direction (from service towards adapters or web). I think it is imperative to avoid and fix those also, as they show a break in the architecture.

framework_tvb/tvb/core/services/burst_service.py
VinczeRobert: I will use h5.index_for_h5_file(result_filename)() in that case.

framework_tvb/tvb/adapters/simulator/monitor_forms.py
```diff
@@ -84,6 +85,23 @@ def get_form_for_monitor(monitor_class):
...
 
+def build_list_of_monitors_from_names(monitor_names, is_surface_simulation):
+    next_monitors_dict = dict()
+    monitor_dict = get_ui_name_to_monitor_dict(is_surface_simulation)
+
+    count = 0
+    for monitor_name in monitor_names:
+        monitor = monitor_dict[monitor_name]
+        next_monitors_dict[monitor.__name__] = (monitor(), count + 1)
+        count = count + 1
```
popaula937, line 96: I see this mechanism in 2 places, could it be extracted to a common method?

popaula937:
Pls do a rebase for this PR, from my side we can merge it after my last comment is addressed.
I think we could extract the code that handles the monitors into a separate class `MonitorsWizzardHandler` and use that in the `SimulatorController` instead of the `next_monitor_dict` and the other methods involved. But we could do that in a separate task.

framework_tvb/tvb/adapters/simulator/monitor_forms.py
VinczeRobert: I don't know, I think it would be a pretty forced extraction because one of them iterates over names and builds the list (when configuring a simulation) and the other one from view models (when loading a simulation). Also, in the first one I mentioned get_ui_name_to_monitor_dict is also called and the code inside the for loop is not really the same, except for the fact that we have a counter variable that keeps increasing.

PR #303, opened 2020-11-27
WIP simulator refactor

maedoc:
This PR refactors the simulator, in order to reduce complexity of adding new features (fields, cosim, TF support) and debugging the current bugs such as surface region mapping conventions.  It also introduces the notion of pluggable numerical backends, following roughly the DAO pattern in the framework, which will have our existing NumPy code as a reference backend, a second one using Numba for accelerations and later others.  Refactoring the simulator dovetails with the backend intro, so it's been started together.

- [x] move monitor, model, integrator et al specific logic out of `Simulator` methods into their own classes
- [ ] move NumPy/Numba array code into backend (primarily code which is called during simulation, not datatype prep)
- [ ] unify region mapping manipulations into `RegionMapping` methods


maedoc:
will reopen this with the correct branch name

PR #304, opened 2020-11-27
TVB-2827: migrate h5py>3

adrian-dordea:
update h5py + minor fix at migration + testing


popaula937:
Pls rebase this and then we can merge

PR #306, opened 2020-12-01
TVB-2842 sim refactor and start reference backend

maedoc:
This PR initiates some refactoring as part of TVB-2842.  To avoid keeping it open too long, this PR shall focus on the Simulator class only.


maedoc:
Had to rebase some commits and force push, then resolve a bunch of conflicts from master, but it should now be in sync with master.

maedoc:
The CI check here is running both library and framework tests.  It's should be caching the pip install and tab-data downloads 🤞 .  This wasn't the goal of the PR, I got a bit distracted there.

i-Zaak:
Cool, that's really good to have for the refactor anyway :) Do you want to continue in this PR?

maedoc:
No, i already lost 2 hours in merge conflicts.  The next PRs should be smaller and happening more quickly.  I think we can do better also by planning a module by module sweep in Jira with dates so that other team members can have a clearer idea when things may be shifting beneath their feet.

i-Zaak (approved):

PR #307, opened 2020-12-02
TVB-2841 tf

maedoc:
This PR explores TF2 as a potential second backend for TVB simulations.


maedoc:
Closing this WIP as we will not move forward with this in the short term, and I'll reopen later.

PR #308, opened 2020-12-08
TVB-2843 Fix delete project for incomplete PSEs

bvalean:


popaula937 (commented):
The removal of projects looks good, but when I remove only a PSE operation, not everything is deleted from the disk: some operation folders with only view models remain.  Pls check this case again.

framework_tvb/tvb/core/services/project_service.py
```diff
@@ -650,13 +660,33 @@ def remove_datatype(self, project_id, datatype_gid, skip_validation=False):
             if len(dependent_dt) > 0:
                 # Do not remove Operation in case DataType still exist referring it.
                 continue
+            op_burst = dao.get_burst_for_operation_id(operation_id)
+            if op_burst:
+                correct = correct and dao.remove_entity(BurstConfiguration, op_burst.id)
             correct = correct and dao.remove_entity(Operation, operation_id)
             # Make sure Operation folder is removed
             self.structure_helper.remove_operation_data(project.name, datatype.fk_from_operation)
```
popaula937, line 668: Might be related to this line, where the same operation id is sent, but I haven't looked into more details.

bvalean: It was related to the line 655. We passed the wrong datatype.fk_from_operation op id instead of the for loop operation id. Fixed

popaula937 (approved):

PR #309, opened 2020-12-08
Intermediate state vars

dionperd:
Some corrections to the modifications already merged that need to be updated asap, because now master leads to errors.

The main correction is that we cannot have dfun return only the integrated variables' derivatives because integration schemes with intermediate steps fail, because they rely to state variable X having the same size as the intermediate derivatives dX.
 So, I have made it to return 0 derivatives for non-integrated intermediate state variables, until some other solution is found. 
We can't give a reduced (masked) state vector to dfun() either, which would be an alternative solution, because then dfun cannot use the intermediate state variables for its computations.
Mind also that although we can modify the integration schemes to use the state_variables_mask, we cannot do so for ScipyODE solvers...

Accordingly we don't need a mask of integrated variables in the base Model class anymore.
Instead, I added a _nintvar and a respective property method @nintvar to count the integrated variables during Model.configure.
I also added a check of assert to compare integration_variables with state_variables.

So, now, the Simulator checks if nintvar < nvar to decide about the integrate_next_step method.

Moreover, we need to reduce redundancy of node_coupling computation in the master branch. It needs to be computed only after integration of step "step" with argument "step+1".

A possible AfferentCoupling Monitor, though, will still get the node_coupling of step "step", and therefore, node_coupling needs to be recomputed after the computation of outputs, and just before we move to next step.

Also, I use start_step = Simulator.current_step + 1 to reduce a bit the redundancy of this +1 used 3 times.

Finally, I corrected the simulator_test substituting asserts of state_variables_mask with 
assert simulator.model.nintvar == len(simulator.model.integration_variables)

as well as models_test and integrators_test for minor errors.

I cannot understand the last error the tests lead to...


maedoc:
> Mind also that although we can modify the integration schemes to use the state_variables_mask, we cannot do so for ScipyODE solvers

We can break or remove the SciPy solvers, as they are not used, as far as I know.

I will try to read/review a little more closely.

dionperd:
> > Mind also that although we can modify the integration schemes to use the state_variables_mask, we cannot do so for ScipyODE solvers
> 
> We can break or remove the SciPy solvers, as they are not used, as far as I know.
> 
> I will try to read/review a little more closely.

In the meantime, can you explain the error we get now?..

maedoc:
> In the meantime, can you explain the error we get now?..

I am not sure, the GitHub test suite is failing, but on my computer I checked out the branch, ran `purest -n8` and see `501 passed, 435 warnings in 156.98s`.  I think there's something wrong with the GitHub tests, and I'm not really sure what it is.  If you are running test suite locally and me too, then we can just drop the `.github` folder and avoid this false red flag entirely.

lionelkusch:
In my opinion, you need all this modification. We need only to modify the line 143 of integrator file :
`X = self.integrate(X, model, coupling, local_coupling, stimulus)`
by 
`X[model.state_variables_mask] = self.scheme(X[model.state_variables_mask], model.dfun, coupling, local_coupling, stimulus)`

dionperd:
> In my opinion, you need all this modification. We need only to modify the line 143 of integrator file :
> `X = self.integrate(X, model, coupling, local_coupling, stimulus)`
> by
> `X[model.state_variables_mask] = self.scheme(X[model.state_variables_mask], model.dfun, coupling, local_coupling, stimulus)`
> This solve almost all the problems.

Hi Lionel,

this is what I explained above that the differential equations for the variables integrated need the non-integrated variables in order to be computed, and therefore, we need the whole X as input to the dfun, and accordingly into the integration scheme.
e.g. In the case of the extension of ReducedWongWang model, I am integrating S and Rint, but Rint uses Rin (inputs from NEST) and S uses R, where Rin and R are variables not integrated.

lionelkusch:
If you need Rin during the integration step, you can save it in a private variable of the model. This private variable is updated before the integration step by the function update_state_variables_before_integration .

I don't see where is your problem.

dionperd:
> If you need Rin during the integration step, you can save it in a private variable of the model. This private variable is updated before the integration step by the function update_state_variables_before_integration .
> 
> I don't see where is your problem.

So, let's make it very specific:

The X vector consists of [S, Rint, R, Rin, I].

Of those 5, the integrated variables are Xint = [S, Rint].

S follows the known model equation that is:
dS/dt = f(S, R)
and 
dRint/dt = -(Rint - Rin)/tau_rint

Therefore, the dfun of the model is a function:

dXint/dt = dfun(X), 

because the non integrated variables Rin and R are necessary.

i.e., it needs the whole X vector to return dXint/dt.

Rin and R are computed indeed in the update_state_variables_before_integration(), via the computation of I, and then only used, not recomputed, in dfun.

This is a specific example of course.

In the general case, though, the intermediate results, as represented by the non integrated state variables, are used in the final dfun, in order to be exactly that, **intermediate**, . If they were not necessary for the computation of the dfun, then they wouldn't be intermediate results, but only play the role of a monitor.

And of course, all this is in the general case irrelevant to co-simulation.

dionperd:
OK. After last changes all tests are passing succesfully.

After talking with Lionel, I opted for another solution where the state vector getting inside the dfun consists only of the integrated variables, and the non integrated ones can be transferred by setting a transitory local attribute of the model.

Still, I had to delete or comment out all the bound and clamping of state variables performed within the integrator schemes.
This might be correct anyway, i.e., choose to better do not interfere with the state variables during an integration scheme's steps.

I have some minor additional modifications, i.e., introducing a n_nonintvar attribute and reversing the order of configuration and derived parameters computation in the Model base.py etc

and have also fixed the tests of course accordingly.

So, I would suggest now to merge, especially if we agree on not bounding or clamping during the integration schemes.
Then, we could also delete completely the commented lines as well.

i-Zaak:
Unless I'm missing something, removing the checking of bounds from the intermediate steps is a bad idea. The bounds define the domain of the dfun, and have to be enforced before it is evoked. If you need to get rid of them for your use case, I'd suggest either using single-step integrator, or a custom integrator without bounds on intermediate values (which explicitly says so in the doc string).

maedoc:
Bound and clamp are required for some use cases (bounds are for e.g. variables representing probabilities must be between 0 and 1, clamping is for the phase plane trajectories on models with more than two variables) so they need to stay in there despite the inelegance.

dionperd:
I understand Jan's argument and he is probably right.
However, think about the following aspect as well:
an integration schemes starts from point x(t) and takes us to point x(t+h) by computing the dfun(x).
When there are intermediate steps in such a scheme, points in between are also computed, e.g., x(t+h/2). However, the desired solution is needed only for points x(t) and x(t+h).
So, I understand that boundaries determine the domain of dfun(), but should they also applied for the point x(t+h/2)?
If not, Jan will say that we are integrating not the correct dfun() but another with a different domain.
On the other hand, if yes, then, we are not using the point x(t+h/2), but potentially some other point, and therefore, we are interfering with the integration scheme, its stability/convergence properties etc...
The objective is to compute x(t+h) with the minimal error. When does this happen? When we respect the boundaries for the x(t+h/2) or when we respect the integration scheme? 
In both cases, boundaries are applied to the point x(t) and x(t+h) that form the solution.

So, please give me a last vote, and if you insist I will work out a solution for bound and clamp to read a mask of state variables in order to bound and clamp only the integrated state variables within the integration schemes.

However, mind that this solution again cannot be applied for the SciPy solvers...

maedoc:
Bounds are applied so that dfun’s domain is satisfied, and the correctness of the integration scheme depends upon it, not the other way around as you seem to suggest.  The clamping is applied for the phase plane viewer and is intentionally “wrong” in light of that use case.  

> SciPy solvers

were addressed by previous comment, they can be dropped for now. 

> read a mask of state variables in order to bound and clamp only the integrated state variables within the integration schemes

seems you already have the solution :)

maedoc:
Small erratum to my previous comment: clamping could be dropped for intermediate steps.  Bounds may not for the reasons I stated.

dionperd:
I have made the following changes:
- separated the bounding and clamping of integrated only state variables within the integration schemas
- moved the configuration of bounding and clamping to Integrator methods
**- adjusted the noise sig size to the number of integrated only state variables (using the state_variable_mask if the user provides nsig according to nvars instead of n_intvars)**
- using n_intvar for Model instead of n_nonintvar
- created a configure_integration_for_model method in Simulator to configure the boundaries and next_step_integration
**- corrected the error in the master branch of computing twice node_coupling**
**- removed Simulator.bound_and_clamp, and using Simulator.Integrator.bound_and_clamp instead**
- adjusted tests that seem to pass now!

The solution has a bit more configuration for the integrator, due to the interaction of boundaries/clamping and non integrated state variables, but it will not have and redundant masking or if statements during integration.

dionperd:
Hello everybody and happy new year!

Could we please move forward with this now?

It would be nice to merge this before the consensus cosimulation branch...

scientific_library/tvb/tests/library/simulator/integrators_test.py
```diff
@@ -179,40 +182,55 @@ def test_clamp(self):
...
                 # Set noise to 0 if this is a stochastic integrator
-                integrator.noise.nsig = numpy.array([0.0])
-                integrator.noise.dt = dt
+                integrator.noise.configure_white(dt, x[model.state_variable_mask].shape)
+                integrator.noise.nsig = numpy.array([0.0] * self.model.nintvar)
             except:
                 pass
+            integrator.configure()
+            integrator.configure_boundaries(model)
+            if model.nvar - model.nintvar:
```
maedoc, line 205: this is an error prone thingy that had better be wrapped as a property on the base model class, something like
```python
@property
def has_interm_vars(self):
    return len(model.nvar) > len(model.nintvar)
```

maedoc (approved):
small comment on property aside this is ok for me.

scientific_library/tvb/tests/library/simulator/integrators_test.py
dionperd: > this is an error prone thingy that had better be wrapped as a property on the base model class, something like

I have added

nnonintvar (= _nvar - _nintvar)

and 

has_nonint_vars (= nnonintvar > 0)

properties.

Using has_nonint_vars in the integrators_test now.

maedoc: thanks

PR #313, opened 2020-12-11
TVB-2743: Refactor BurstService.prepare_indexes_for_simulation_results

VinczeRobert:


VinczeRobert:
I don't like the way I computed the time_series_type, I just didn't know how to do it otherwise.

popaula937 (commented):
This could be a direction to pass over the need to load the whole Timeseries in memory, but mind that you are preparing only the base TimeSeriesIndex, which will not put the complete information in DB, you should deal with the subtypes, too.

This is part of a larger story in TVB: avoid loading large datatypes in-memory. I am thinking maybe we could make a generic change for this. Maybe on the base `H5File.load_into()` we could avoid loading in memory the `Dataset` types that have the `expand_dimension` attribute set to a larger value than -1. But that means, we need to analyse the usages to see what other adjustments are necessary.

framework_tvb/tvb/adapters/datatypes/db/time_series.py
```diff
@@ -83,13 +83,19 @@ def fill_from_has_traits(self, datatype):
...
+    def fill_from_h5(self, h5_file):
+        self.time_series_type = type(h5_file).__name__.replace('H5', '')
+        self.title = h5_file.title.load()
+        self.start_time = h5_file.start_time.load()
+        self.sample_period_unit = h5_file.sample_period_unit.load()
+        self.sample_period = h5_file.sample_period.load()
+        self.sample_rate = h5_file.sample_rate.load()
+        self.labels_ordering = json.dumps(h5_file.labels_ordering.load())
+        self.labels_dimensions = json.dumps(h5_file.labels_dimensions.load())
+        self.data_ndim = len(h5_file.data.shape)
```
popaula937, line 97: this is set twice, here and inside `fill_shape()`

popaula937 (approved):

liadomide (approved):

liadomide:
> This could be a direction to pass over the need to load the whole Timeseries in memory, but mind that you are preparing only the base TimeSeriesIndex, which will not put the complete information in DB, you should deal with the subtypes, too.
> 
> 
> 
> This is part of a larger story in TVB: avoid loading large datatypes in-memory. I am thinking maybe we could make a generic change for this. Maybe on the base `H5File.load_into()` we could avoid loading in memory the `Dataset` types that have the `expand_dimension` attribute set to a larger value than -1. But that means, we need to analyse the usages to see what other adjustments are necessary.

I think the current solution is decent.

I agree on a more uniform solution need for chunked datatypes. One first thought is to introduce a new method 'load_partial' to avoid confusion. But that should be done in another step, together with the main usages of such cases in analyzers

PR #320, opened 2020-12-16
TVB-2852 Fix Step1 build

bvalean:


tvb_build/Jenkinsfile-Step1
```diff
@@ -26,6 +34,7 @@ pipeline {
             steps {
                 sh '''#!/bin/bash
                     source activate tvb-docs
+                    pip install docutils==0.12
```
liadomide, line 37: The installation of docutils here is odd. 
I saw yesterday it was removed from dockerfile, where this env is first created. I think it should stay in the dockerfile.
Maybe this is only temporary, until we find a good version?

PR #330, opened 2020-12-21
TVB-2827: revert changes to previous version due to incompatibility

adrian-dordea:


popaula937:
AllenSDK does not support yet h5py > 2.10, so we decided to keep using that version for now.

PR #343, opened 2021-01-11
TVB-2855: Fixed Tracts Importers and Visualizer

VinczeRobert:
I seemingly managed to fix the importers and the visualizer partially. One of the first problems I encountered is that judging by how the code is written I am pretty sure that the Region Volume Mapping when importing a Tracts should be a required, not an optional field. Paula sent me a RVM and a connectivity which I was able to import with the 5M.trk file and I also changed the code until I was able to see some tracts in the viewer. The problem is that it strongly seems that not only are the vertices not read in slices, but there are actually more vertices read than they are said to be in the H5 which clearly means that the reading of vertices in slices is what still needs to be fixed.

### DIFF
--- a/framework_tvb/tvb/adapters/uploaders/tract_importer.py
+++ b/framework_tvb/tvb/adapters/uploaders/tract_importer.py
@@ -233,8 +233,8 @@ class TrackvizTractsImporter(_TrackImporterBase):
             datatype.vertices = vox2ras.transform(vertices)
             tracts_h5.write_vertices_slice(datatype.vertices)
 
-        datatype.tract_start_idx = numpy.array(tract_start_indices, dtype=numpy.int16)
-        datatype.tract_region = numpy.array(tract_region, dtype=numpy.int16)
+        datatype.tract_start_idx = numpy.array(tract_start_indices)
+        datatype.tract_region = numpy.array(tract_region)
 
         tracts_index = TractsIndex()
         tracts_index.fill_from_has_traits(datatype)
@@ -281,5 +281,5 @@ class ZipTxtTractsImporter(_TrackImporterBase):
                 vertices_file.close()
 
         datatype.tract_start_idx = tract_start_indices
-        datatype.tract_region = numpy.array(tract_region, dtype=numpy.int16)
+        datatype.tract_region = numpy.array(tract_region)
         return datatype
--- a/framework_tvb/tvb/adapters/uploaders/tract_importer.py
+++ b/framework_tvb/tvb/adapters/uploaders/tract_importer.py
@@ -34,13 +34,14 @@
 import numpy
 from abc import ABCMeta
 from nibabel import trackvis
+from tvb.adapters.datatypes.h5.region_mapping_h5 import RegionVolumeMappingH5
 from tvb.adapters.datatypes.h5.tracts_h5 import TractsH5
 from tvb.core.adapters.abcuploader import ABCUploader, ABCUploaderForm
 from tvb.core.adapters.exceptions import LaunchException
 from tvb.core.entities.file.files_helper import TvbZip
 from tvb.adapters.datatypes.db.tracts import TractsIndex
 from tvb.core.entities.generic_attributes import GenericAttributes
-from tvb.core.entities.storage import transactional
+from tvb.core.entities.storage import transactional, dao
 from tvb.core.neocom import h5
 from tvb.core.neocom.h5 import path_for
 from tvb.core.neotraits.uploader_view_model import UploaderViewModel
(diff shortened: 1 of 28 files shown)


popaula937:
The problem arises from the importer, where you have chosen `tract_start_idx` to have `dtype=numpy.int16` and you end up with negative numbers in `tract_start_idx` because of int16 overflows. Thus, in the viewer, the chunks are not computed correctly (`count` end up being a negative number and conditions based on it will be broken).

framework_tvb/tvb/adapters/uploaders/tract_importer.py
```diff
@@ -148,15 +149,20 @@ def _attempt_to_cache_regionmap(self, region_volume):
...
+                self.region_volume = rvm
+
+            region_volume = h5.load_from_index(rvm_index)
+
             # this is called here and not in _get_tract_region because it goes to disk to get this info
             # which would massively dominate the runtime of the import
-            self.region_volume_shape = region_volume.read_data_shape()
 
         datatype = Tracts()
         datatype.storage_path = self.storage_path
```
popaula937, line 168: `storage_path` attribute does not exist anymore on `Tracts`, pls remove it.

framework_tvb/tvb/adapters/uploaders/tract_importer.py
```diff
@@ -148,15 +149,20 @@ def _attempt_to_cache_regionmap(self, region_volume):
...
+            rvm_index = dao.get_datatype_by_gid(region_volume.hex)
+            rvm_h5_path = h5.path_for_stored_index(rvm_index)
+            with RegionVolumeMappingH5(rvm_h5_path) as rvm:
+                self._attempt_to_cache_regionmap(rvm)
+                self.region_volume_shape = rvm.read_data_shape()
+                self.region_volume = rvm
+
+            region_volume = h5.load_from_index(rvm_index)
+
             # this is called here and not in _get_tract_region because it goes to disk to get this info
```
popaula937, line 164: this comment seems to be left out of its purpose

framework_tvb/tvb/adapters/uploaders/tract_importer.py
```diff
@@ -265,15 +271,15 @@ def launch(self, view_model):
...
 
-                if region_volume_index is not None:
-                    tract_region.append(self._get_tract_region(tract_vertices[0]))
+                if view_model.region_volume is not None:
+                    tract_region.append(self._get_tract_region(datatype.tract_vertices[0]))
                 vertices_file.close()
 
         datatype.tract_start_idx = tract_start_indices
-        datatype.tract_region = numpy.array(tract_region, dtype=numpy.int16)
+        datatype.tract_region = numpy.array(tract_region)
```
popaula937, line 284: Why not keep int16 here? I believe the region numbers will fit inside int16

popaula937 (commented):
Looks good, few comments from my side.

framework_tvb/tvb/core/adapters/abcdisplayer.py
```diff
@@ -183,8 +195,22 @@ def dump_with_precision(xs, precision=3):
         format_str = "%0." + str(precision) + "g"
         return "[" + ",".join(format_str % s for s in xs) + "]"
 
-    def _load_h5_of_gid(self, entity_gid):
+    def load_h5_of_gid(self, entity_gid):
```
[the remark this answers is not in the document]
VinczeRobert: Because it is used outside of it's class in several places

framework_tvb/tvb/adapters/uploaders/tract_importer.py
VinczeRobert: I thought we discussed to not explicitly put the size and let numpy choose the smallest size that it can fit the data into.

popaula937: That was available only for `tract_start_idx` where values won't fit into int16. But for `tract_region` we know values would fit and numpy probably won't use dtype=int16 by default.

framework_tvb/tvb/core/adapters/abcdisplayer.py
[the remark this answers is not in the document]
popaula937: That's ok, I missed it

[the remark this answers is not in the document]
liadomide: I think this method deserves a TODO in all its usages. Can't we avoid using it, as it exposes too much of the internal structures in viewers? For example, small objects (like Volume hasTraits) could be loaded entirely (with the generic api) instead of having to use the file path and h5class in the viewer.
For larger objects (surface, timeseries) maybe we could have helper methods on these datatype classes, and also avoid using the file path in the viewer

PR #348, opened 2021-01-18
Region mapping refactoring

BalanceKey:
Refactoring the region mapping definition, by using the method datatypes.RegionMapping.full_region_mapping(surf, conn), which uses the surface of the cortex and the connectivity matrix, thus mapping both the cortical and the subcortical regions. The goal of this is to make it easier in the future to add modifications to the way the region mapping is realized, by just changing this one method.

Actually, the method datatypes.RegionMapping.from_file is also used in other cases where the region mapping is retrieved from a .txt file. These are at the moment the two ways to define a region mapping.


BalanceKey:
@maedoc could you review this ? 
(P.s. sorry for the last two commits, was just trying to undo the unnecessary modifications to gitignore...)

scientific_library/tvb/simulator/monitors.py
```diff
@@ -261,7 +261,9 @@ def config_for_sim(self, simulator):
         # setup given spatial mask or default to region mapping
         if self.spatial_mask is None:
             self.is_default_special_mask = True
-            if not (simulator.surface is None):
+            if simulator.surface is not None and simulator.connectivity is not None:
```
liadomide, line 264: I think 'simulator.conn' is always not None.

scientific_library/tvb/simulator/simulator.py
```diff
@@ -546,7 +543,7 @@ def _configure_stimuli(self):
         if self.stimulus is not None:
             if self.surface:
                 # NOTE the region mapping of the stimuli should also include the subcortical areas
-                self.stimulus.configure_space(region_mapping=numpy.r_[self.surface.region_mapping, self.connectivity.unmapped_indices(self.surface.region_mapping)])
+                self.stimulus.configure_space(region_mapping=region_mapping.RegionMapping.full_region_mapping(self.surface, self.connectivity))
```
liadomide, line 546: Can't we reuse here 'self._regmap'?

scientific_library/tvb/datatypes/region_mapping.py
```diff
@@ -65,6 +65,23 @@ def from_file(source_file="regionMapping_16k_76.txt"):
         result = RegionMapping()
         result.array_data = reader.read_array(dtype=numpy.int32)
         return result
+    
+    @staticmethod
+    def full_region_mapping(surf, conn):
```
liadomide, line 70: Could we maybe receive as input param directly surf.region-_mapping?

Otherwise Here on a theoretical level we have a circular dependency (sim.surface class and RegionMapping)

liadomide (commented):
Could you please do a rebase with master and thus fix the conflicts?

scientific_library/tvb/datatypes/region_mapping.py
BalanceKey: actually it's not the sim.surface class, but the Cortex class that this input param 'surf' represents... 

I know the name is confusing, but I didn't want to change it since in the Simulator class the attribute surface corresponds to an object from the Cortex class.

scientific_library/tvb/simulator/monitors.py
BalanceKey: I would say so too, thanks

scientific_library/tvb/datatypes/region_mapping.py
BalanceKey: but the remark remains valid... I can change the input param in that case... @maedoc any remarks on this ?

BalanceKey:
I updated the region mapping refactoring, using the method defined in the backend by @maedoc .
I ran the pytests locally (having some issues with the automatic run).

scientific_library/tvb/simulator/monitors.py
```diff
@@ -291,8 +293,8 @@ def config_for_sim(self, simulator):
         # setup given spatial mask or default to region mapping
         if self.spatial_mask is None:
             self.is_default_special_mask = True
-            if not (simulator.surface is None):
-                self.spatial_mask = simulator.surface.region_mapping
+            if simulator.surface is not None:
+                self.spatial_mask, _, _ = self.backend.full_region_map(simulator.surface, simulator.connectivity)
```
liadomide, line 297: Could it be that the `backend. full_region_map` implementation is missing from the current committed state ?

maedoc: it's in the branch [here](https://github.com/BalanceKey/tvb-root/blob/reg-map-fix/scientific_library/tvb/simulator/backend/ref.py#L112)

maedoc:
@BalanceKey how's it going ? do you want to have a chat about this PR?

BalanceKey:
@maedoc sure, let's chat

scientific_library/tvb/tests/library/simulator/regionmapping_test.py
```diff
@@ -0,0 +1,56 @@
...
+# import TVB modules
+from tvb.simulator.lab import *
+from tvb.datatypes.region_mapping import RegionMapping
+from tvb.tests.library.base_testcase import BaseTestCase
+
+class SetUpSimulator(object):
+    """
+    Simulator test class for region mapping
+    """
+    def __init__(self, monitors=(monitors.Raw(),)):
```
maedoc, line 18: instead of placing this in a separate class, can you just make this the `setUp` method on the TestCase below?

maedoc (changes requested):
just a small change to do then we can merge

maedoc (approved):

PR #349, opened 2021-01-19
TVB-2862 Clean files uploaded via REST uploaders

bvalean:


framework_tvb/tvb/interfaces/rest/server/facades/operation_facade.py
```diff
@@ -107,10 +107,8 @@ def launch_operation(self, current_user_id, model_file, project_gid, algorithm_m
                 for key, value in adapter_instance.get_form_class().get_upload_information().items():
                     data_file = fetch_file(request_file_key=key, file_extension=value)
                     data_file_path = save_temporary_file(data_file, temp_folder)
-                    file_name = os.path.basename(data_file_path)
                     upload_field = getattr(view_model_h5, key)
-                    upload_field.store(os.path.join(storage_path, file_name))
-                    shutil.move(data_file_path, storage_path)
+                    upload_field.store(data_file_path)
```
popaula937, line 111: Won't be enough to do this change because uploaders run async and the temp folder is cleaned up in the `finally` block.

popaula937:
Discussed with Bogdan and we will remove the `finally` block from here because temp data will be cleaned up by our mechanism in operation service.

popaula937 (approved):

PR #350, opened 2021-01-19
TVB-2858 Fix TS importing

bvalean:
TS import was failing on postgres if we export a TS (with links/ without links), then delete all projects, create a new project where we are trying to import this TS.

### DIFF
--- a/framework_tvb/tvb/adapters/exporters/TVBLinkedExporter.py
+++ b/framework_tvb/tvb/adapters/exporters/TVBLinkedExporter.py
@@ -81,12 +81,19 @@ class TVBLinkedExporter(ABCExporter):
             file_destination = os.path.join(data_export_folder, os.path.basename(data_path))
             if not os.path.exists(file_destination):
                 FilesHelper().copy_file(data_path, file_destination)
+
             sub_dt_refs = f.gather_references()
 
             for reference in sub_dt_refs:
                 if reference[1]:
                     dt = dao.get_datatype_by_gid(reference[1].hex)
                     self.copy_dt_to_export_folder(dt, data_export_folder)
+
+        with H5File.from_file(file_destination) as dest_file:
+            if 'parent_burst' in dest_file.storage_manager.get_metadata():
+                dest_file.storage_manager.remove_metadata('parent_burst')
+
+
 
     def get_export_file_extension(self, data):
         return "zip"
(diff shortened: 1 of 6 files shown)


framework_tvb/tvb/adapters/exporters/tvb_export.py
```diff
@@ -38,6 +38,7 @@
 from tvb.adapters.exporters.exceptions import ExportException
 from tvb.core.entities.model.model_datatype import DataType
 from tvb.core.neocom import h5
+from tvb.core.neotraits._h5core import H5File
```
liadomide, line 41: Importing modules with underscore is not advices outside their own package. Can't you find h5file in "api"?

framework_tvb/tvb/adapters/exporters/TVBLinkedExporter.py
```diff
@@ -81,13 +81,20 @@ def copy_dt_to_export_folder(self, data, data_export_folder):
             file_destination = os.path.join(data_export_folder, os.path.basename(data_path))
             if not os.path.exists(file_destination):
                 FilesHelper().copy_file(data_path, file_destination)
+
```
liadomide, line 84: Please rename this file to match the pattern of the other .py files

liadomide (commented):
So the approach now is to remove the "parent_burst" reference from any individual DT exported (with or without links). "Parent_burst" will be preserved only when we export and import an entire project. Is this correct?

popaula937:
> So the approach now is to remove the "parent_burst" reference from any individual DT exported (with or without links). "Parent_burst" will be preserved only when we export and import an entire project. Is this correct?

Yes, only at individual export, we remove the "parent_burst"

framework_tvb/tvb/adapters/exporters/tvb_export.py
```diff
@@ -81,9 +82,20 @@ def export(self, data, export_folder, project):
...
 
+    def copy_dt_to_export_folder(self, data, data_export_folder):
+        data_path = h5.path_for_stored_index(data)
+        file_destination = os.path.join(data_export_folder, os.path.basename(data_path))
+        if not os.path.exists(file_destination):
+            FilesHelper().copy_file(data_path, file_destination)
+
+        with H5File.from_file(file_destination) as dest_file:
+            if 'parent_burst' in dest_file.storage_manager.get_metadata():
+                dest_file.storage_manager.remove_metadata('parent_burst')
```
popaula937, line 96: Could this code be simplified to avoid accessing `storage_manager` directly from here by adding a `remove_metadata_param` method on `H5File`, and handling inside that method the `if` condition? We won't need to open the file here either. Would be similar to the `H5File.get_metadata_param` static method.

PR #352, opened 2021-01-20
Tvb cosim consensus

lionelkusch:
Hi all,
@dionperd  and I have finished creating the cosimulation interface for TVB.

Can you review it for including this new contribution?

Best regard,
Lionel


lionelkusch:
Can you include the co-simulation test in the check of future commit?
Thank you in advance,
Lionel

maedoc:
408 commits, 5700 lines added, 49 files modified 😅

dionperd:
> 408 commits, 5700 lines added, 49 files modified 😅

Almost all of it is cosimulation related and shouldn't affect the tvb library.

I see that there are some minor changes to the sensors.py and time_series.py datatypes, that probably originate to my other tvb-multiscale-cosim branch.

Unless there is a test failing, I would suggest to not worry about them.

dionperd:
Maybe also adding @popaula937  as a reviewer here.

scientific_library/tvb/simulator/coupling.py
```diff
@@ -501,3 +501,6 @@ def pre(self, x_i, x_j):
 
     def post(self, gx):
         return self.a / gx.shape[0] * gx
+
+
+CouplingWithCurrentState = [Difference, Kuramoto]
```
maedoc, line 506: undocumenated global `list` that anyone could modify during runtime, please use an `enum` or something more appropriate

scientific_library/tvb/simulator/models/wong_wang_exc_inh.py
```diff
@@ -182,9 +182,15 @@ class ReducedWongWangExcInh(ModelNumbaDfun):
...
+    I_ext = NArray(
+        label=":math:`I_{ext}`",
+        default=numpy.array([0.0, ]),
+        domain=Range(lo=0.0, hi=1.0, step=0.001),
+        doc="""[nA]. Effective external stimulus input""")
+
     G = NArray(
         label=":math:`G`",
-        default=numpy.array([2.0, ]),
+        default=numpy.array([20.0, ]),
```
maedoc, line 193: If this model is present in a current TVB release, please do not change the default parameter values; a jump from 2 to 20 is pretty big.

scientific_library/tvb/simulator/monitors.py
```diff
@@ -54,6 +54,7 @@
 
 """
 
+import os
```
maedoc, line 57: why do monitors need access to `os` module? please remove

tvb_build/docker/Dockerfile-multiscale-build
```diff
@@ -0,0 +1,193 @@
...
+    mkdir $VENV/build
+WORKDIR $VENV/build
+RUN mkdir libneurosim; \
+    cd libneurosim; \
+    PYTHON=$VENV/bin/python $HOME/packages/libneurosim/configure --prefix=$VENV; \
+    make; make install; ls $VENV/lib $VENV/include; \
+    ln -s /usr/lib/python3.7/config-x86_64-linux-gnu/libpython3.7.so $VENV/lib/
+
+# Install NEST, latest stable version (July 2019)
+ARG NEST_VER=2.18.0
```
maedoc, line 59: please collect all build args at the top and namespace them  (here they're all `NEST_XYZ` but other below like `LAST_SHA_ROOT` should be `TVB_SHA` etc)

tvb_contrib/tvb/contrib/cosimulation/cosim_history.py
```diff
@@ -0,0 +1,99 @@
...
+    @property
+    def nbytes(self):
+        return self.buffer.nbytes
+
+    def __init__(self,  n_time, n_var, n_node, n_mode):
+        self.n_time = n_time  # state buffer has n past steps and the current one
+        self.n_var = n_var
+        self.n_node = n_node
+        self.n_mode = n_mode
+        self.buffer[:] = numpy.NAN
```
maedoc, line 64: why NaNs here? is there an assert somewhere else that ensures that the buffer is completely and correctly filled?

tvb_contrib/tvb/contrib/cosimulation/cosim_monitors.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def get_sample(self, start_step, n_steps, history,cosim):
+        times = []
+        values = []
+        for step in range(start_step, start_step + n_steps):
+            if cosim:
+                state = history.query(step)
+            else:
+                state = history.query(step)[0]
+            tmp = super(self.__class__, self).sample(step, state)
```
maedoc, line 54: If this is a base class, and you want to call the subclass sample method, just write `self.sample(step, state)`

tvb_contrib/tvb/contrib/cosimulation/cosim_monitors.py
```diff
@@ -0,0 +1,185 @@
...
+    """
+    Abstract base class for cosimulation monitors implementations.
+    """
+
+    def get_sample(self, start_step, n_steps, history,cosim):
+        times = []
+        values = []
+        for step in range(start_step, start_step + n_steps):
+            if cosim:
+                state = history.query(step)
```
maedoc, line 51: For repeated calls to query your cosim history, why not just implement a new method on that history instance to retrieve data for the `start_step` to `start_step+n_steps` interval at once?

tvb_contrib/tvb/contrib/cosimulation/cosim_monitors.py
```diff
@@ -0,0 +1,185 @@
...
+        """
+        This method provides monitor output, and should be overridden by subclasses.
+        Use the original signature.
+        """
+        pass
+
+class CosimMonitorFromCoupling(CosimMonitor):
+    """
+       Abstract base class for a monitor that records the future coupling values.
+       !!!WARNING don't use this monitor for a time smaller than the synchronization variable!!!
```
maedoc, line 71: Docstring warnings will not show up in a debugger; please insert appropriate `assert`s in the initialization or `get_sample` method to ensure the use is correct.

tvb_contrib/tvb/contrib/cosimulation/cosim_monitors.py
```diff
@@ -0,0 +1,185 @@
...
+class CosimCoupling(AfferentCoupling, CosimMonitorFromCoupling):
+    """
+    A monitor that records the future coupling of selected variables:
+    It collects:
+
+        - selected coupling values and all modes from class :Model:
+        - all nodes of a region or surface based
+        - all the integration time steps
+
+    !!!WARNING don't use this monitor for a time smaller than the synchronization variable!!!
```
maedoc, line 178: as before please add runtime check, not docstring warning

tvb_contrib/tvb/contrib/cosimulation/cosim_monitors.py
```diff
@@ -0,0 +1,185 @@
...
+                state = history.query(step)
+            else:
+                state = history.query(step)[0]
+            tmp = super(self.__class__, self).sample(step, state)
+            if tmp is not None:
+                times.append(tmp[0])
+                values.append(tmp[1])
+        return [numpy.array(times), numpy.array(values)]
+
+    @abc.abstractmethod
```
maedoc, line 60: This has no effect if the metaclass is not ABCMeta.  Also, this class is used as a mix-in so I would suggest dropping this decorator, and just `raise NotImplemented` instead of `pass` in the method body below.

maedoc:
13/49 files.. I suggest you give me time to go through all of them before replying to comments, otherwise I will probably get confused

dionperd:
> 13/49 files.. I suggest you give me time to go through all of them before replying to comments, otherwise I will probably get confused

Sure. OK. Thanks for looking at them thoroughly.

However, it is really not 49 files. Actually most files are tests.

The problem is that there are some remnants of older versions in some scientific_library files after consecutive merges from master or other branches, like the import os in the monitors or the not-used list in coupling.py...
Sorry about this...

tvb_build/docker/Dockerfile-multiscale-build
```diff
@@ -0,0 +1,193 @@
...
+          -Dwith-python=3 \
+          ###-Dwith-music=ON \
+          -Dwith-libneurosim=$VENV \
+          -DPYTHON_LIBRARY=/usr/lib \
+          -DPYTHON_INCLUDE_DIR=/usr/include/python3.7m \
+          ${NEST_SRC}; \
+    make; make install
+
+# Install nestml
+RUN $VENV/bin/pip3 install setuptools; \
```
maedoc, line 81: this line fails with the error
```
error: numpy 1.16.2 is installed but numpy>=1.17 is required by {'astropy'}
```

tvb_contrib/tvb/contrib/cosimulation/cosimulator.py
```diff
@@ -0,0 +1,310 @@
...
+from tvb.simulator.common import iround
+from tvb.simulator.simulator import Simulator, math
+
+from tvb.contrib.cosimulation.cosim_history import CosimHistory
+from tvb.contrib.cosimulation.cosim_monitors import CosimMonitor
+
+
+class CoSimulator(Simulator):
+
+    exclusive = Attr( #Todo need to test it
```
maedoc, line 46: to do?

tvb_contrib/tvb/contrib/cosimulation/models/generic_2d_oscillator_multiscale.py
```diff
@@ -0,0 +1,292 @@
...
+        vw_ = vw.reshape(vw.shape[:-1]).T
+        c_ = c.reshape(c.shape[:-1]).T
+        deriv = _numba_dfun_g2d(vw_, c_, self.tau, self.I, self.a, self.b, self.c, self.d, self.e, self.f, self.g,
+                                self.beta, self.alpha, self.gamma, lc_0)
+        return deriv.T[..., numpy.newaxis]
+
+
+@guvectorize([(float64[:],) * 16], '(n),(m)' + ',()'*13 + '->(n)', nopython=True)
+def _numba_dfun_g2d(vw, c_0, tau, I, a, b, c, d, e, f, g, beta, alpha, gamma, lc_0, dx):
+    "Gufunc for reduced Wong-Wang model equations."
```
maedoc, line 286: looks like copy paste here, this is not the Wong-Wang model.  Why the need for a new model if it's the same as the existing one?

tvb_contrib/tvb/contrib/cosimulation/models/reduced_wong_wang_exc_io.py
```diff
@@ -0,0 +1,334 @@
...
+@guvectorize([(float64[:],)*11], '(n),(m)' + ',()'*8 + '->(n)', nopython=True)
+def _numba_update_non_state_variables_before_integration(S, c, a, b, d, w, jn, r, g, io, newS):
+    "Gufunc for reduced Wong-Wang model equations."
+
+    newS[0] = S[0]  # S
+    newS[1] = S[1]  # Rint
+
+    cc = g[0]*jn[0]*c[0]
+    newS[4] = w[0] * jn[0] * S[0] + io[0] + cc  # I
+    if r[0] <= 0.0:
```
maedoc, line 47: This and line 56 below are a little hard to understand.  Are these conditionals expressing different numerics or different code paths for different components?  What does `r[0] <= 0.0` mean? Why below branch on `S[1] < 40.0`?

tvb_contrib/tvb/contrib/cosimulation/models/reduced_wong_wang_exc_io_inh_i.py
```diff
@@ -0,0 +1,317 @@
...
+#   CITATION:
+# When using The Virtual Brain for scientific publications, please cite it as follows:
+#
+#   Paula Sanz Leon, Stuart A. Knock, M. Marmaduke Woodman, Lia Domide,
+#   Jochen Mersmann, Anthony R. McIntosh, Viktor Jirsa (2013)
+#       The Virtual Brain: a simulator of primate brain network dynamics.
+#   Frontiers in Neuroinformatics (7:10. doi: 10.3389/fninf.2013.00010)
+
+"""
+Models based on Wong-Wang's work.
```
maedoc, line 28: same docstring as the other module, please just add a few words to distinguish them

tvb_contrib/tvb/contrib/scripts/utils/data_structures_utils.py
```diff
@@ -288,7 +288,7 @@ def list_of_dicts_to_dict_of_lists(lst):
 
 
 def list_of_dicts_to_dicts_of_ndarrays(lst, shape=None):
```
maedoc, line 290: docstrings please, with [doctest](https://docs.python.org/3/library/doctest.html) since that would help a lot to illustrate the intended usage

tvb_contrib/tvb/contrib/scripts/utils/data_structures_utils.py
```diff
@@ -868,24 +871,28 @@ def data_xarray_from_continuous_events(events, times, senders, variables=[],
         from xarray import DataArray
```
maedoc, line 871: Return types depending on the installed set of packages is a difficult debug session waiting to happen, unless the two (dict and DataArray) behave the same, in which case, only the dict is required.  Please move this out to the call site

tvb_contrib/tvb/contrib/scripts/utils/data_structures_utils.py
```diff
@@ -868,24 +871,28 @@ def data_xarray_from_continuous_events(events, times, senders, variables=[],
...
         return DataArray(data, dims=list(coords.keys()), coords=coords, name=name)
     except:
+        # Return a dictionary as a plan B'
         return {"data": data, "dims": list(coords.keys()), "coords": coords, "name": name}
 
 
-def concatenate_heterogeneous_DataArrays(data, dim_name, data_name="", fill_value=np.nan, dims=None):
+def concatenate_heterogeneous_DataArrays(data, concat_dim_name,
+                                         data_keys=None, name=None, fill_value=np.nan, transpose_dims=None):
+    from pandas import Series
```
maedoc, line 880: ok more libs here, neither xarray nor pandas are on the dependency list for TVB.  since it's in contrib, we can leave it but consider how to handle this in production, as an optional dependency.

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/ReducedWongWang.py
```diff
@@ -0,0 +1,87 @@
...
+#
+#   CITATION:
+# When using The Virtual Brain for scientific publications, please cite it as follows:
+#
+#   Paula Sanz Leon, Stuart A. Knock, M. Marmaduke Woodman, Lia Domide,
+#   Jochen Mersmann, Anthony R. McIntosh, Viktor Jirsa (2013)
+#       The Virtual Brain: a simulator of primate brain network dynamics.
+#   Frontiers in Neuroinformatics (7:10. doi: 10.3389/fninf.2013.00010)
+
+"""
```
maedoc, line 27: This module doesn't provide any unit tests, so some module level comments are in order: what is going on here, why is it not just in the contrib model module if it's a model, etc.

My guess is that this model is primarily for mocking up a parallel cosim and it's going to make tests easier to run.  But again, please write out an explanation here.

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/double_proxy_precision_complex_delay_update_test.py
```diff
@@ -0,0 +1,98 @@
...
+from tvb.tests.library.base_testcase import BaseTestCase
+from tvb.contrib.tests.cosimulation.parallel.function_tvb import TvbSim
+
+
+class TestDoubleProxyPrecisionComplexDelayUpdate(BaseTestCase):
+    """
+    test the transmission of information between two models with proxy in most complex case and different delay
+    """
+    def test_double_precision_complex_delay_update(self):
+        weight = np.array([[5, 2, 4, 0], [8, 5, 4, 1], [6, 1, 7, 9], [10, 0, 5, 6]])
```
maedoc, line 44: NumPy will use integer weights here, probably not a big deal, but will affect testing precision below.  Use `]], np.float64)` at the end, for isntance.

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/double_proxy_precision_complex_delay_update_test.py
```diff
@@ -0,0 +1,98 @@
...
+#       The Virtual Brain: a simulator of primate brain network dynamics.
+#   Frontiers in Neuroinformatics (7:10. doi: 10.3389/fninf.2013.00010)
+
+"""
+.. moduleauthor:: Lionel Kusch <lkusch@thevirtualbrain.org>
+.. moduleauthor:: Dionysios Perdikis <dionperd@gmail.com>
+"""
+
+import numpy as np
+import numpy.random as rgn
```
maedoc, line 33: since this is mispelled, might as well drop this import and just reference `np.random` below.

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/double_proxy_precision_complex_delay_update_test.py
```diff
@@ -0,0 +1,98 @@
...
+
+        # full simulation
+        rgn.seed(42)
+        sim_ref = TvbSim(weight, delay, [], resolution_simulation, time_synchronize)
+        time, result_ref = sim_ref(time_synchronize)
+
+        # COMPARE PROXY 1
+        diff_1 = np.where(np.squeeze(result_ref[:, proxy_id_2, :], axis=2)[0] !=
+                          np.squeeze(result_1[0][:, proxy_id_2, :], axis=2)[0])
+        assert diff_1[0].size == 0
```
maedoc, line 70: Please use NumPy testing functions, e.g. [`np.testing.assert_allclose`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html) with a reasonable tolerance.

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/double_proxy_precision_complex_test.py
```diff
@@ -0,0 +1,104 @@
+# -*- coding: utf-8 -*-
```
maedoc, line 1: similar comments as the other test, watch numbers and precision, use numpy testing functions etc

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/function_tvb.py
```diff
@@ -0,0 +1,208 @@
...
+
+import tvb.simulator.lab as lab
+from tvb.contrib.tests.cosimulation.parallel.ReducedWongWang import \
+    ReducedWongWangProxy, _numba_dfun_proxy
+
+from tvb.contrib.cosimulation.cosim_monitors import RawCosim
+from tvb.contrib.cosimulation.cosimulator import CoSimulator
+
+
+rgn.seed(42)
```
maedoc, line 43: nope, don't do this.  seed inside the test function.

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/function_tvb.py
```diff
@@ -0,0 +1,208 @@
...
+def tvb_simulation(time, sim, data_proxy):
+    """
+    Simulate t->t+dt:
+    :param time: the time of simulation
+    :param sim: the simulator
+    :param data_proxy : the firing rate of the next steps
+    :return:
+        the time, the firing rate and the state of the network
+    """
+    if sim.__class__ == lab.simulator.Simulator:
```
maedoc, line 133: `isinstance(sim, lab.simulator.Simulator)`

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/function_tvb.py
```diff
@@ -0,0 +1,208 @@
...
+        result = sim.run(simulation_length=time)
+        time = result[0][0]
+        s = result[0][1][:, 0]
+        rate = result[0][1][:, 1]
+    else:
```
maedoc, line 138: `elif isinstance(sim,???)`

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/modify_wongwang_precision_test.py
```diff
@@ -0,0 +1,130 @@
...
+
+import tvb.simulator.lab as lab
+from tvb.tests.library.base_testcase import BaseTestCase
+from tvb.contrib.tests.cosimulation.parallel.ReducedWongWang import ReducedWongWangProxy
+
+from tvb.contrib.cosimulation.cosim_monitors import RawCosim
+from tvb.contrib.cosimulation.cosimulator import CoSimulator
+
+
+SIMULATION_LENGTH = 3.0
```
maedoc, line 42: why is this a global? why not a method arg below?

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/modify_wongwang_precision_test.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def test_with_no_cosimulation(self):
+        connectivity, coupling, integrator, monitors, sim, result, result_all = self._reference_simulation()
+        np.random.seed(42)
+        init = np.concatenate((np.random.random_sample((385, 1, 76, 1)),
+                               np.random.random_sample((385, 1, 76, 1))), axis=1)
+        np.random.seed(42)
+        result_2 = self._reference_simulation(model_class=ReducedWongWangProxy, simulator=CoSimulator, init=init)[5]
+        diff = result - result_2
+        assert np.sum(diff) == 0.0
```
maedoc, line 86: `np.testing.assert_allclose` please

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/monitors_test.py
```diff
@@ -0,0 +1,138 @@
...
+import tvb.simulator.lab as lab
+from tvb.tests.library.base_testcase import BaseTestCase
+from tvb.contrib.tests.cosimulation.parallel.ReducedWongWang import ReducedWongWangProxy
+
+from tvb.contrib.cosimulation.cosim_monitors import RawCosim, RawVoiCosim, RawDelayed,\
+    RawVoiDelayed, CosimCoupling
+from tvb.contrib.cosimulation.cosimulator import CoSimulator
+
+
+SIMULATION_LENGTH = 3.0
```
maedoc, line 43: should be a test parameter

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/monitors_test.py
```diff
@@ -0,0 +1,138 @@
...
+        for i in range(int(SIMULATION_LENGTH/integrator.dt)):
+            diff = result_all[0][1][i][0][1:] - result_1_all[1][i+sync_steps, 0, 1:]
+            diff_2 = result_all[0][1][i][0][:1] - result_1_all[1][i+sync_steps, 0, :1]
+            assert np.sum(diff, where=np.logical_not(np.isnan(diff))) == 0.0 and \
+                   np.sum(diff_2, where=np.logical_not(np.isnan(diff_2))) == 0.0
+
+        for i in range(sim_to_sync_time):
+            result_step = result_cosim_monitors[i]
+            # check the dimension of the monitors
+            assert result_step[0][1].shape == (sync_steps, 2, 76, 1)
```
maedoc, line 129: `np.testing.assert_....` please

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/update_function_test.py
```diff
@@ -0,0 +1,66 @@
...
+        for i in range(0, 100):
+            time, result = sim(synchronization_time,
+                               [np.arange(i * synchronization_time, (i + 1) * synchronization_time,
+                                          resolution_simulation),
+                                np.repeat(firing_rate.reshape(1, 2),
+                                          int(synchronization_time / resolution_simulation), axis=0)])
+        assert True
+
+        # Test a fail function due to the time of simulation too long
+        try:
```
maedoc, line 62: please assert the exception raised, https://docs.pytest.org/en/stable/assert.html#assertions-about-expected-exceptions

tvb_contrib/tvb/contrib/tests/cosimulation/simple/update_function_test.py
```diff
@@ -0,0 +1,65 @@
...
+        firing_rate = np.array([[20.0, 10.0]]) * 10 ** -3  # time units in tvb is ms so the rate is in KHz
+
+        sim = TvbSim(weight, delay, proxy_id, resolution_simulation, synchronization_time)
+        time, result = sim(resolution_monitor, [np.array([resolution_simulation]), firing_rate])
+        for i in range(0, 100):
+            time, result = sim(synchronization_time,
+                               [time + resolution_monitor,
+                                np.repeat(firing_rate.reshape(1, 2),
+                                          int(resolution_monitor / resolution_simulation), axis=0)])
+        assert True
```
maedoc, line 65: what's being tested here?

maedoc (changes requested):
It looks good to me, and a lot of my comments are advisory more than necessary.  I would though like to see a working Dockerfile and check test coverage before moving on.  Additionally I would like to see the tests done with floats instead of integers and allclose, not plain asserts.

tvb_contrib/tvb/contrib/cosimulation/cosim_monitors.py
dionperd: This base class doesn't inherit the base Monitor class because we want to avoid the diamond class inheritance.

tvb_contrib/tvb/contrib/cosimulation/cosim_monitors.py
dionperd: I don't really see the point here since anyway we need to loop for each step in case history is a normal TVB history and in all cases for the sampling with the TVB monitor.

tvb_contrib/tvb/contrib/cosimulation/cosim_monitors.py
dionperd: I have removed the warning from the docstrings.
The ideal place for a check is upon configuration, since then simulator.synchronization_time will be available, e.g., by overwriting the Monitor._config_time() method. 
However, I am not sure what kind of assertion is due here.
@lionelkusch could you help with this?

tvb_contrib/tvb/contrib/cosimulation/cosim_monitors.py
dionperd: Done

tvb_build/docker/Dockerfile-multiscale-build
dionperd: I have removed completely this Dockerfile. It shouldn't have been there in the first place.

tvb_contrib/tvb/contrib/cosimulation/cosimulator.py
dionperd: Removed todo

tvb_contrib/tvb/contrib/cosimulation/models/generic_2d_oscillator_multiscale.py
dionperd: I deleted this model from tvb_contrib and corrected the docstring for the original TVB model.

tvb_contrib/tvb/contrib/cosimulation/models/reduced_wong_wang_exc_io.py
dionperd: The first one is indeed a particular implementation for different code for the nodes that are updated by NEST.
The second one has to do with a very particular use-case.
I moved this whole models folder under tvb.contrib.scripts, because they are too specific to particular use-cases of TVB-multiscale, which are due to change also upon refactoring TVB-multiscale to use this code..

tvb_contrib/tvb/contrib/scripts/utils/data_structures_utils.py
dionperd: The tvb_contrib scripts module is indeed in need of a lot of things. Currently it is used only by me and TVB-multiscale.
I promise to deal with it at a later stage.

dionperd:
I have dealt with most of the comments, except for tests (@lionelkusch will work on them) and tvb_contrib scripts module that anyway requires a lot to be done.
Hoping that this will be enough to merge! Thanks Marmaduke for looking at it!

scientific_library/tvb/simulator/coupling.py
lionelkusch: The variable is not used and it can be remove.

scientific_library/tvb/simulator/models/wong_wang_exc_inh.py
lionelkusch: ok

scientific_library/tvb/simulator/monitors.py
lionelkusch: ok

tvb_contrib/tvb/contrib/cosimulation/cosim_history.py
lionelkusch: The initialization of the NAN values is for tracking the possibility of bugs for the calculation of the coupling by the users before the start of the simulation.
There are not an assert somewhere but it can be add in the function  _update_cosim_history of the cosimulator to validate the update  of the simulator history.
Do you want to add it?

tvb_contrib/tvb/contrib/cosimulation/cosim_monitors.py
lionelkusch: There are a need to include this new function to the scientific library.
We doesn't include it for the moment.
We can do it if you want.
if yes, what name do you propose for this new function?

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/ReducedWongWang.py
lionelkusch: This is the model used by the other tests for testing the cosimulation.
If you check the name, there are not test in it.

lionelkusch: I add the comments

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/double_proxy_precision_complex_delay_update_test.py
lionelkusch: ok

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/double_proxy_precision_complex_delay_update_test.py
lionelkusch: By looking the code, I find that the weight are define by np.float instead of np.float64 .
It's a bug or not?
I prefer to add np.float as the code of the connectivity

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/double_proxy_precision_complex_delay_update_test.py
lionelkusch: We can discuss about it but I prefer to not using the assert_allclose because the precison need to be the same for the cosimulation and the normal simulator with the same input.
Actually it works because the precision of the cosimhistory is the same as the precision of the output of TVB.

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/function_tvb.py
lionelkusch: sorry, it was a left code

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/function_tvb.py
lionelkusch: ok

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/function_tvb.py
lionelkusch: ok

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/modify_wongwang_precision_test.py
lionelkusch: I add it in the attribute of the test class.
It's not an argument of _reference_simulation because it's need to in the test for some of assertions.

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/modify_wongwang_precision_test.py
lionelkusch: see my comments above on this topics

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/monitors_test.py
lionelkusch: ok

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/monitors_test.py
lionelkusch: see above

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/update_function_test.py
lionelkusch: ok

tvb_contrib/tvb/contrib/tests/cosimulation/simple/update_function_test.py
lionelkusch: it's a running test, a test which test if a simple co-simulation can be simulate the simulaiton until the end without error.

lionelkusch:
I fix most of the other comments.
I let them some open because there are a need a discussion about it.

liadomide:
Please don't add init__.py files under `scientific_library`, `tvb_contrib` of `framework_tvb` folders

lionelkusch:
I didn't get the answer for including the tests of the co-simulation in the check of github.

lionelkusch:
> Please don't add init__.py files under `scientific_library`, `tvb_contrib` of `framework_tvb` folders

I remove them

dionperd:
Hi, I am fine with Lionel's changes.
I am sorry I forgot to correct some of the changes in scientific_library and contrib, but Lionel took care of them.
I only reverted one of them, which adds DEFAULT_SIZE = SMALL_SIZE to plot.config of scientific_library.
I hope you don't mind. Some of my code in tvb-multiscale might fail without it and it doesn't harm anyone!
If you DO mind, just revert my last commit.
So, as far as I am concerned, it is ready to merge.

tvb_build/docker/Dockerfile-multiscale-build
lionelkusch: We remove the docker file.

tvb_contrib/tvb/contrib/cosimulation/cosim_monitors.py
lionelkusch: Denis refactor the code for having more clarity about it.

tvb_contrib/tvb/contrib/cosimulation/cosim_monitors.py
lionelkusch: I did the modification about this comment in this commit : 28dac89
I am not sure it's the best way. If you have more comment about do it.

tvb_contrib/tvb/contrib/cosimulation/cosim_monitors.py
lionelkusch: see above

lionelkusch:
There is still some reviewed are not handle because I miss information or it can be discuss. I wait the input from @maedoc

tvb_contrib/tvb/contrib/cosimulation/cosim_history.py
maedoc: It's a good idea, but yes please add the check.  Ideally, there would be an exception specific to the condition, 
```python
class NumericalInstability(ValueError):
    pass
```
which you can raise,
```
if nan_detected(): 
    raise NumericalInstability
```
During testing or parameter tuning process, the caller can catch this specific exception and handle it more intelligently (like reducing coupling or time step), than just aborting completely.

tvb_contrib/tvb/contrib/cosimulation/cosim_monitors.py
maedoc: no it's ok, was minor comment

tvb_contrib/tvb/contrib/cosimulation/cosim_monitors.py
maedoc: That commit adds a current_step argument to a bunch of methods, but there's no assert or exception raised, which is what I meant, something like 
```python
if t0 < t1:
    raise ValueError("don't use this monitor for a time smaller than the synchronization variable")
```

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/double_proxy_precision_complex_delay_update_test.py
maedoc:
```python
>>> np.float==float
True
```
I guess that means that the weights even if given as integers will be coerced to floats, which is what I meant so it's fine.

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/double_proxy_precision_complex_delay_update_test.py
maedoc: Then `np.testing.assert_array_equal` would be appropriate.

maedoc:
Thanks @lionelkusch and @dionperd (and @liadomide) for the replies so far.  I am fine with nearly everything except two

- [ ] handle inappropriate conditions by raising appropriate exceptions
- [ ] use NumPy testing functions instead of custom equals tests

When these are addressed we can merge

tvb_contrib/tvb/contrib/tests/cosimulation/parallel/double_proxy_precision_complex_delay_update_test.py
lionelkusch: I did it for most of the assert.
I didn't do it for the assert where the delay make the difference.

tvb_contrib/tvb/contrib/cosimulation/cosim_monitors.py
lionelkusch: I add all the exception in the abstract class.

tvb_contrib/tvb/contrib/cosimulation/cosim_history.py
lionelkusch: I add the assert
I saw, there is a special class for TVB exception : TVBException
Want you to inheritance from this class for the different type of exception in the co-simulation class?

lionelkusch:
I modify the code for the last missing reviews.

maedoc (approved):

maedoc:
@popaula937 @bvalean do you want to review this?

popaula937:
Looks good from my side, looked mostly at how traits are used and the file structure. Wouldn't the models be better placed under a cosimulation/models folder?

We could also have a setup to automatically run these tests, I think we can constrain them to run only when changes are made under the tvb-contrib folder with github actions. I could help with that.

dionperd:
> Looks good from my side, looked mostly at how traits are used and the file structure. Wouldn't the models be better placed under a cosimulation/models folder?
> 
> We could also have a setup to automatically run these tests, I think we can constrain them to run only when changes are made under the tvb-contrib folder with github actions. I could help with that.

The problem with these cosimulation models is that I don't know yet their final form, after adjusting tvb-multiscale to use tvb-cosim-consensus. Once I adjust them and test them, we can either move them under tvb.contrib.cosimulation.models or remove them completely from here and move them to tvb_multiscale.core.tvb.cosimulator.models.

Let me make a specific pull request about this at a later stage.

dionperd:
Marmaduke, please wait also for Lionel to check my last changes -possibly today- before merging.

lionelkusch:
It's OK for and thank you for having taken in charge of the particularity of the coupling monitor.

maedoc:
@lionelkusch @dionperd last check are we ok to merge?

dionperd:
> @lionelkusch @dionperd last check are we ok to merge?

Yes!

There can be an enhancing/correcting pull request later on when we start using it.

lionelkusch:
It's ok for merging

maedoc:
🍻 thanks both of you

PR #353, opened 2021-01-26
TVB-2870: Move everything related to monitors handling in a new class

VinczeRobert:
Except the url handling (which I think it must stay in the controller) I moved everything else related to monitors (preparing the fragments and the monitor dict) in this one new class. I chose to send context instead of 4 params to handle_next_fragment_for_monitors, otherwise we woul end up with 10 parameters for that method and I wanted to avoid that many sent params.

### DIFF
--- a/framework_tvb/tvb/interfaces/web/controllers/simulator/monitors_wizard_handler.py
+++ b/framework_tvb/tvb/interfaces/web/controllers/simulator/monitors_wizard_handler.py
@@ -34,7 +34,7 @@
 
 from tvb.adapters.simulator.equation_forms import get_form_for_equation
 from tvb.adapters.simulator.monitor_forms import AdditiveNoiseViewModel, get_ui_name_to_monitor_dict, \
-    get_form_for_monitor, prepare_monitor_legend, BoldViewModel
+    get_form_for_monitor, BoldViewModel, get_monitor_to_ui_name_dict
 from tvb.adapters.simulator.simulator_fragments import SimulatorMonitorFragment, SimulatorFinalFragment
 from tvb.core.services.algorithm_service import AlgorithmService
 from tvb.interfaces.web.controllers.simulator.simulator_wizzard_urls import SimulatorWizzardURLs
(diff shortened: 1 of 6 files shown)


framework_tvb/tvb/interfaces/web/controllers/simulator/monitors_wizard_handler.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def build_list_of_monitors_from_view_models(self, monitor_view_models):
+        self.next_monitors_dict = dict()
+        count = 0
+        for monitor_model in monitor_view_models:
+            self.next_monitors_dict[monitor_model.__class__.__name__] = (monitor_model, count + 1)
+            count = count + 1
+
+    def build_list_of_monitors_from_names(self, monitor_names, is_surface_simulation):
+        monitor_dict = get_ui_name_to_monitor_dict(is_surface_simulation)
```
popaula937, line 63: Could be private

framework_tvb/tvb/interfaces/web/controllers/simulator/monitors_wizard_handler.py
```diff
@@ -0,0 +1,139 @@
...
+    def clear_next_monitors_dict(self):
+        if self.next_monitors_dict:
+            self.next_monitors_dict.clear()
+
+    def build_list_of_monitors_from_view_models(self, monitor_view_models):
+        self.next_monitors_dict = dict()
+        count = 0
+        for monitor_model in monitor_view_models:
+            self.next_monitors_dict[monitor_model.__class__.__name__] = (monitor_model, count + 1)
+            count = count + 1
```
popaula937, line 60: Looking at the usages of the `next_monitors_dict` it seems we could simplify it to hold only classname -> index pairs. What do you think?

framework_tvb/tvb/interfaces/web/controllers/simulator/monitors_wizard_handler.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def clear_next_monitors_dict(self):
+        if self.next_monitors_dict:
+            self.next_monitors_dict.clear()
+
+    def build_list_of_monitors_from_view_models(self, simulator):
+        monitor_names = []
+        monitor_dict = get_monitor_to_ui_name_dict(simulator.is_surface_simulation)
+        for monitor in simulator.monitors:
+            monitor_names.append(monitor_dict[monitor.__class__])
```
popaula937, line 60: Better use `type(monitor)` instead of `monitor.__class__`

framework_tvb/tvb/tests/framework/interfaces/web/controllers/simulator_controller_test.py
```diff
@@ -578,7 +578,7 @@ def test_set_seeg_monitor_params(self):
             self.simulator_controller.context.add_session_stored_simulator(self.session_stored_simulator)
             self.simulator_controller.context.add_burst_config_to_session(BurstConfiguration(self.test_project.id))
             self.simulator_controller.set_monitors(**self.sess_mock._data)
-            self.simulator_controller.set_monitor_params('iEEGViewModel', **self.sess_mock._data)
+            self.simulator_controller.set_monitor_params('Intracerebral / Stereo EEG', **self.sess_mock._data)
```
popaula937, line 581: We will have a problem with using the monitors UI names inside the URLs because in the case of names like `Intracerebral / Stereo EEG` we have a '/' which will be considered part of the URL and treated like there are 2 URL params sent here instead of 1. Try it in the GUI to see how it behaves.

PR #354, opened 2021-01-26
Tvb 2854 - Operation overlay fix in case of surface simulations

adrian-dordea:
Fix CortexViewModel attributes display for simulations

### DIFF
--- a/framework_tvb/tvb/core/adapters/inputs_processor.py
+++ b/framework_tvb/tvb/core/adapters/inputs_processor.py
@@ -49,7 +49,7 @@ def _review_operation_inputs_for_adapter_model(form_fields, form_model, view_mod
         if not hasattr(view_model, field.name):
             continue
         attr_vm = getattr(view_model, field.name)
-        if type(field) == TraitUploadField:
+        if attr_vm and type(field) == TraitUploadField:
             attr_vm = os.path.basename(attr_vm)
 
         if isinstance(field, TraitDataTypeSelectField):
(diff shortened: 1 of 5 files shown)


framework_tvb/tvb/core/adapters/inputs_processor.py
```diff
@@ -49,18 +50,19 @@ def _review_operation_inputs_for_adapter_model(form_fields, form_model, view_mod
         if not hasattr(view_model, field.name):
             continue
         attr_vm = getattr(view_model, field.name)
-        if type(field) == TraitUploadField:
+        if attr_vm and type(field) == TraitUploadField:
             attr_vm = os.path.basename(attr_vm)
 
-        if isinstance(field, TraitDataTypeSelectField):
+        if isinstance(attr_vm, CortexViewModel):
+            for attr, value in attr_vm.__dict__.items():
```
popaula937, line 57: Is it really necessary to do this CortexViewModel check? I thought we could keep this method generic and prepare the surface data before calling this (in get_adapter_fragments).

adrian-dordea: I initialy tryed so, as you said, but in that case there is a problem that not all the fields that are missing can be added to the overlay, more precisely "region_mapping", due to the fact that in view model there are 2 fields "region_mapping" that is empty when a field check is performed on it whitch will skip it's display in the overlay and "region_mapping_data" whitch is a uuid but the field check is for "region_mapping" not "region_mapping_data", so the current solution came to mind...

popaula937 (approved):

PR #355, opened 2021-01-26
TVB-2871: Replace usage of load_h5_of_gid with something more suitable.

VinczeRobert:


framework_tvb/tvb/adapters/visualizers/complex_imaginary_coherence.py
```diff
@@ -94,9 +95,9 @@ def get_required_memory_size(self, view_model):
         """
         Return the required memory to run this algorithm.
         """
-        input_data_h5_class, input_data_h5_path = self.load_h5_of_gid(view_model.input_data.hex)
-        with input_data_h5_class(input_data_h5_path) as input_data_h5:
-            required_memory = numpy.prod(input_data_h5.read_data_shape()) * 8
+        input_data_h5 = h5.h5_file_for_gid(view_model.input_data.hex)
+        required_memory = numpy.prod(input_data_h5.read_data_shape()) * 8
+        input_data_h5.close()
```
liadomide, line 100: Good replacement in itself, but i suggest keeping the "with" syntax

VinczeRobert: Can I use the "with" syntax if the input_data_h5 is already a created object (and the H5 file is opened)? h5_file_for_gid returns an object.

liadomide: Yes, any Object that has Method "close" can be used with "with"

VinczeRobert: I hope I didn't miss anything... I didn't put with where they would've been to much code under it. Also, there where a few more small files like the CorrelationCorrelationsH5 for which we could upload the HasTraits data but we also need the source gid from it, which we can't directly do so I decided to leave the usage of H5 in those cases.

framework_tvb/tvb/adapters/visualizers/time_series_volume.py
```diff
@@ -224,8 +205,6 @@ def _get_voxel_time_series_region(self, ts_h5, x, y, z, var=0, mode=0):
             background = numpy.ones((time_length, 1)) * ts_h5.out_of_range(back_min)
             label = 'background'
 
-        volume_rm_h5.close()
-        connectivity_h5.close()
-
+        ts_h5.close()
```
popaula937, line 208: Same here

popaula937 (commented):
We have `ABCAdapter.load_traited_by_gid` with the same intention as `h5.load_from_gid`. We should keep only one version, pls check it out.
Before closing this, could you analyze the usages of the `h5.h5_file_for_index` and `h5.load_from_index` and replace them with `h5.h5_file_for_gid` and `h5.load_from_gid`  where better fitted?

framework_tvb/tvb/adapters/visualizers/time_series_volume.py
```diff
@@ -187,24 +177,16 @@ def get_voxel_time_series(self, entity_gid, **kwargs):
...
+            if isinstance(ts_h5, TimeSeriesRegionH5):
                 return self._get_voxel_time_series_region(ts_h5, **kwargs)
 
-            return ts_h5.get_voxel_time_series(**kwargs)
-
-    def _get_voxel_time_series_region(self, ts_h5, x, y, z, var=0, mode=0):
-        region_mapping_volume_gid = ts_h5.region_mapping_volume.load()
-        if region_mapping_volume_gid is None:
-            raise Exception("Invalid method called for TS without Volume Mapping!")
+        return ts_h5.get_voxel_time_series(**kwargs)
```
popaula937, line 185: At this point the `ts_h5` is closed, maybe do: `data = ts_h5.get_voxel_time_series(**kwargs)` inside `with` and then `return data` here.

framework_tvb/tvb/adapters/visualizers/region_volume_mapping.py
```diff
@@ -216,12 +215,30 @@ def compute_params(self, region_mapping_volume=None, measure=None, data_slice=''
...
+
+    @staticmethod
+    def prepare_space_parameters(ts_h5, x_plane, y_plane, z_plane):
+        region_mapping_volume_gid = ts_h5.region_mapping_volume.load()
+
+        if region_mapping_volume_gid is None:
+            raise Exception("Invalid method called for TS without Volume Mapping!")
+
+        volume_rm_h5 = h5.h5_file_for_gid(region_mapping_volume_gid)
+        volume_rm_shape = volume_rm_h5.array_data.shape
```
popaula937, line 235: This method is hard to follow and one could easily forget to close the `volume_rm_h5` outside it.
We could do a generic method for both visualizers that handles the opening of `volume_rm_h5` using `with` and then calls `_get_volume_view_region` which receives also `volume_rm_h5` as arg but does something different for each viewer. What do you think?

VinczeRobert: I have done it. I needed to change more things in order to make it work, so please have a look to make sure I did not miss something.

VinczeRobert:
> We have `ABCAdapter.load_traited_by_gid` with the same intention as `h5.load_from_gid`. We should keep only one version, pls check it out.
> Before closing this, could you analyze the usages of the `h5.h5_file_for_index` and `h5.load_from_index` and replace them with `h5.h5_file_for_gid` and `h5.load_from_gid` where better fitted?

I think I correctly replaced h5.h5_file_for_index everywhere it was possible and related to h5.load_from_index I used either h5.load_from_gid or load_traited_by_gid from ABCAdapter. The thing is that both of them use load_entity_by_gid, but there are two different load_entity_by_gid methods. 

As a last thing to improve, should we somehow merge the two methods into only one load_entity_by_gid? In that case, probably we could just keep one of load_from_gid and load_traited_by_gid as well.

PR #358, opened 2021-02-04
TVB-2859: Fix surface visualizer with region boundaries

adrian-dordea:
remove region boundry box in case of no region mapping present

### DIFF
--- a/framework_tvb/tvb/interfaces/web/templates/jinja2/visualizers/surface/surface_viewer_controls.html
+++ b/framework_tvb/tvb/interfaces/web/templates/jinja2/visualizers/surface/surface_viewer_controls.html
@@ -2,11 +2,7 @@
     <dl>
         {% include 'flow/color_scheme.html' %}
         {% with %}
-            {% if boundaryURL %}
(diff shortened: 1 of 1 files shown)


framework_tvb/tvb/interfaces/web/templates/jinja2/visualizers/surface/surface_viewer_controls.html
```diff
@@ -2,8 +2,12 @@
     <dl>
         {% include 'flow/color_scheme.html' %}
         {% with %}
-            {% set urlRegionBoundaries=True %}
+            {% if boundaryURL %}
+                {% set urlRegionBoundaries=True %}
+            {% else %}
+                {% set urlRegionBoundaries=False %}
+            {% endif %}
```
bvalean, line 9: In this case you can set directly the urlRegionBoundaries=boundaryURL because it will be checked in the brain_toggle_controls.html file

bvalean (changes requested):

popaula937 (approved):

PR #359, opened 2021-02-05
TVB-2442: Replace SQL-Migrate with Alembic

VinczeRobert:


VinczeRobert:
I wrote a few things about alembic here: https://req.thevirtualbrain.org/browse/TVB-2442. Also, we should decide if we want that template file for migration scripts and wether we want to support offline migrations or not.

popaula937 (commented):
Looks good, pls add the alembic dependency in the Dockerfiles and mac_conda_setup.sh, too. Basically, in every environment config file where sqlalchemy-migrate is currently added.

About the automatically generated files, I would vote to keep only the ones we need for the automatic update. I don't think offline migrations are necessary at this point.

popaula937:
Related to this, it would be of great help for us to have an automated test on Jenkins for the migration.

popaula937 (approved):

PR #362, opened 2021-02-09
TVB-2876: Fix integrator choice in PhasePlane

VinczeRobert:
I propose this solution which would work for any select fields that could have '/' in their name. When ui_values is provided, the integrator names that appear in the UI are NOT used to get the VMs from choices, instead we use another dict which does not have names containing '/'.


popaula937 (approved):
I am approving this fix but I believe it reveals a need to further review the behavior of the `SelectField` in a separate task [TVB-2879](https://req.thevirtualbrain.org/browse/TVB-2879).
Pls rebase this.

PR #363, opened 2021-02-10
HBP-134 Avoid wrong adapter dependencies

bvalean:


popaula937 (approved):
I think it looks good for now

PR #364, opened 2021-02-11
TVB-2875: Define a Jenkins job to test migration from 1.5.8

VinczeRobert:


tvb_build/docker/Dockerfile-migrate
```diff
@@ -0,0 +1,4 @@
+FROM tvb-run
```
bvalean, line 1: Do you want to use the **thevirtualbrain/tvb-run** image?

VinczeRobert: Yes

bvalean: Then you have to use full image name thevirtualbrain/tvb-run

tvb_build/Jenkinsfile-Migrate
```diff
@@ -0,0 +1,39 @@
...
+                }
+            }
+            steps {
+                sh '''#!/bin/bash
+                    source activate tvb-run
+                    cd tvb_build
+                    bash install_full_tvb.sh
+                    cd ..
+                    python -m tvb.config.init.migrated_files_counter
+                    xCopy /home/tvb_user/TVB_STORAGE/logs . /E/I
```
popaula937, line 22: maybe use `cp -R` command for the copy step

tvb_build/docker/Dockerfile-migrate
popaula937: The image is currently built on Robert's local machine, not on Jenkins. As a next step, we will try to build it on Jenkins and use the existing thevirtualbrain/tvb-run, but then we will need to keep the generated project in a central place.

popaula937 (approved):

PR #369, opened 2021-02-23
TVB-2881: Solved issue which appears when removing a PSE simulation.

VinczeRobert:
I removed that recursive call because there is no need for it. The way our code is built, one operation which is part of a group is enough to delete all datatypes and operations from the respective datatype groups and operation groups.

### DIFF
--- a/framework_tvb/tvb/core/services/operation_service.py
+++ b/framework_tvb/tvb/core/services/operation_service.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 #
-# TheVirtualBrain-Framework Package. This package holds all Data Management, and 
+# TheVirtualBrain-Framework Package. This package holds all Data Management, and
 # Web-UI helpful to run brain-simulations. To use it, you also need do download
 # TheVirtualBrain-Scientific Package (for simulators). See content of the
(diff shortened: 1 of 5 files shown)


framework_tvb/tvb/core/services/operation_service.py
```diff
@@ -540,17 +540,15 @@ def stop_operation(operation_id, is_group=False, remove_after_stop=False):
...
-            for operation in operations_in_group:
-                result = OperationService.stop_operation(operation.id, False, remove_after_stop) or result
-        else:
-            result = BackendClientFactory.stop_operation(operation_id)
-            if remove_after_stop:
-                burst_config = dao.get_burst_for_direct_operation_id(operation_id)
-                ProjectService().remove_operation(operation_id)
-                if burst_config is not None:
-                    result = dao.remove_entity(BurstConfiguration, burst_config.id) or result
+            operation_id = operations_in_group[0].id
```
popaula937, line 546: In `ProjectService` you can add a method: `get_first_operation_in_group(op_group)` and inside it do: `dao.get_operations_in_group(operation_group.id, only_first_operation=True)`.

popaula937 (approved):

popaula937 (commented):
What happens in case the PSE is still running and you want to cancel it?

PR #372, opened 2021-03-01
TVB-2664: Clean the fire operation flow.

VinczeRobert:


framework_tvb/tvb/tests/framework/core/services/operation_service_test.py
```diff
@@ -83,33 +85,40 @@ def _assert_stored_ddti(self, expected_cnt=1):
...
-        adapter_instance = ABCAdapter.build_adapter(algo)
-        view_model = TestModel()
-        data = {model_burst.RANGE_PARAMETER_1: 'param_5', 'param_5': [1, 2]}
-        # Create Group of operations
-        OperationService().fire_operation(adapter_instance, self.test_user, self.test_project.id,
-                                          view_model=view_model, **data)
+        dt_group = datatype_group_factory(project=self.test_project)
+        model = TestModel()
+        test_adapter_factory()
+        adapter = TestFactory.create_adapter("tvb.tests.framework.adapters.testadapter1", "TestAdapter1")
```
popaula937, line 98: Isn't `datatype_group_factory` generating view_models and using existing algorithms already?

popaula937 (commented):
Have you also tested the `fire_operation` method after these changes?
It is important for the command line interface - tvb/interfaces/command/lab.py

Please also verify that the related jupyter notebooks are running .

popaula937:
Should we also hide the "Analyzers" tab from the overlay opened for DatatypeGroups? Those are supposed to launch PSE for analyzers which we do not support now.

liadomide:
> Should we also hide the "Analyzers" tab from the overlay opened for DatatypeGroups? Those are supposed to launch PSE for analyzers which we do not support now.
> @liadomide what do you think?

yes, good point, we should hide analyzers tab for groups

popaula937:
Copy/Branch operations are broken, they end up referencing the same Burst as the original simulation.

popaula937:
@VinczeRobert pls always check the test results on Jenkins, one was still failing.

PR #373, opened 2021-03-02
TVB-2714 Fix windows docker tests

bvalean:


VinczeRobert:
The reason why those (22 I think) tests didn't work is the export method in simulation. When you call it from the tests, there is a buffered reader that remains opened and that's why the files in the EXPORT_TMP folder are "used by another process". When you call the export method through the UI, the same problem does not appear, you can delete the files after exporting has finished. In conclusion, I manually chose closed those readers and now every test is passing on my machine.

popaula937 (approved):
All tests are now passing on Windows, too. @bvalean will update the Jenkinsfile before merging, but the rest look good from my side.

PR #375, opened 2021-03-08
TVB-2901: Solve problems during sprint end testing

VinczeRobert:


framework_tvb/tvb/core/services/simulator_service.py
```diff
@@ -63,7 +64,7 @@ def __init__(self):
     @staticmethod
     def _reset_model(session_stored_simulator, form):
         session_stored_simulator.model = type(session_stored_simulator.model)()
-        vi_indexes = form.determine_indexes_for_chosen_vars_of_interest(session_stored_simulator)
+        vi_indexes = MonitorForm.determine_indexes_for_chosen_vars_of_interest(session_stored_simulator)
```
popaula937, line 67: The `form` argument does not seem to be necessary now

liadomide: True, but now we have an undesired import from adapters in core. Can't we do it differently?

VinczeRobert: I will try a different way then.

popaula937: Right, missed that. 
Looking better at these methods, they don't seem to belong under the `Monitor`.
How about moving them under `SimulatorAdapterModel` which is in `core`?

popaula937: Then you would have: session_stored_simulator.determine_indexes_for_chosen_vars_of_interest(..)

VinczeRobert: I didn't read your comment before doing in another way, I think this one is maybe better. I thought about putting it in the SimulatorAdapterModel class but I don't really see methods on the ViewModel classes so I thought not to do that.

popaula937: But this is still an import from adapters to the core and we are trying to avoid that

VinczeRobert: Oh yeah, that method is in adapters too... I will put it on the ViewModel then, don't have any other ideas.

popaula937 (approved):

PR #376, opened 2021-03-08
TVB-2685 Update licenses

bvalean:


popaula937 (approved):
The new licenses seem compatible. 
For sure the cfflib is not a direct dependency for TVB anymore, so it could be removed from packages_accepted and also from our Dockerfiles. Not sure whether it could be an indirect dependency though.

@bvalean please prepare the PR for merge. I believe there is a change to be done in a Dockerfile.

PR #378, opened 2021-03-09
TVB-2901: Assign another connectivity (if possible and if needed) at sim importing 

VinczeRobert:
This could be generalized for the other datatypes as well.


VinczeRobert:
We decided to solve this problem in a whole different way (see PR #379), so I will close this one.

PR #380, opened 2021-03-16
HBP-145 Use multiple encryption keys

bvalean:


popaula937 (commented):
Looks good, the only problem I see is related to extensibility. 
If we want to add more keys and we already have 21 projects: project 21 has been already encrypted with key1 at the moment we had only 20 keys. 
If we add another 10 keys, at decryption time, for project 21, key21 is chosen for decryption, and the decryption will not work.

popaula937 (approved):

PR #388, opened 2021-03-24
TVB-2911: Implement a BIDS importer

VinczeRobert:
So far I have extended the script written by Bogdan which was importing each connectivity. Now it also imports the FC.mat as a CorrelationCoefficients and HRF.csv file as a TS (with a good amound of hardcoding).

We need to store metadata in a JSON file and probably other files as well. For example in the HRF.csv file I suppose we only have the array_data matrix, but we also need a time matrix, so maybe another .csv file?


VinczeRobert:
I accidentally closed this PR while trying to move my commit from master to this branch, but it seems I cannot reopen it...

popaula937:
Taken care of by PR #412

PR #389, opened 2021-03-31
TVB 2892 Python codegen backend

maedoc:
This PR adds templates for generating Python code (plain NumPy & Numba) along with unit and integration tests.  Surface and monitor support are currently not present. 

Prior to considering merge, we should 
- [x] update Dockerfile for dependencies
- [x] passing GH CI
- [x] fix Marseille GPU runner for passing all tests
- [x] fix conflicts with current master
- [x] have pytest benchmark for Numba sim, generated & @i-Zaak version to avoid perf regressions


maedoc:
Setup local runner for special hardware testing (AVX512, CUDA) but currently without isolation.  I will stop the runner and set up with some isolation to avoid security issues.

maedoc:
I'm dropping the self-hosted CI for the moment, as it's trickier than I expected.

PR #393, opened 2021-04-03
Faster history updater

maedoc:
This reduces by 4x the time spent in `Simulator._loop_history_update` during a surface simulation, by caching useful data and applying an in-place parallelized update algorithm.

This is currently a draft since the code changes are not entirely well placed, the new algorithm assumes 1 mode and it also assumes region indices are contiguous in the regmap.


maedoc:
Marked in #555 as something requiring further work.  Closing PR for now.

PR #395, opened 2021-04-06
HPB-151 Do not create new threads if MAX_THREADS_NUMBER is achieved

bvalean:
This PR intends to optimize resources used when multiple operations are launched at the same time.


liadomide (commented):
By just reading the changes, this PS looks decent.

I would like to see some unit-tests written for this part, as it might easily spoil things:
- launch long and many operations, check that some enter the queue. Then cancel started ops, and check that others start to execute instead
- launch a PSE with op > MAX_THREAD_NUMBER and assert that all ops finish, but only maximum MAX_THREAD_NUMBER are at a certain moment in state started 
- other not so happy flows...

popaula937:
Could you separate the caching solution to another PR?

framework_tvb/tvb/core/entities/storage/session_maker.py
```diff
@@ -53,10 +54,13 @@
 
 def build_db_engine():
     if TvbProfile.current.db.SELECTED_DB == 'postgres':
-        ### Control the pool size for PostgreSQL, otherwise we might end with multiple
-        ### concurrent Python processes failing because of too many opened connections.
-        DB_ENGINE = create_engine(TvbProfile.current.db.DB_URL, pool_recycle=5, max_overflow=1,
-                                  pool_size=TvbProfile.current.db.MAX_CONNECTIONS)
+        if TvbProfile.current.db.BOUNCER_IN_FRONT:
```
liadomide, line 57: I think I understand this need, but this condition is not ideal, as it poses deep in our code a dependency towards deployment.

bvalean:
Required changes in #403

PR #397, opened 2021-04-08
Done profilling

Raghibshams456:
Profiling of benchmark.py...need some guidance what to do after this..
Screenshot from 2021-04-08 18-25-52
Screenshot from 2021-04-08 18-25-45

### DIFF
--- a/framework_tvb/tvb/interfaces/command/benchmark.py
+++ b/framework_tvb/tvb/interfaces/command/benchmark.py
@@ -34,6 +34,8 @@ Some standardized simulations are run and a report is generated in the console o
 """
 
 from os import path
+from numba import jit
+import memory_profiler
 
 import tvb_data
 from tvb.adapters.datatypes.db.connectivity import ConnectivityIndex
@@ -102,7 +104,7 @@ class Bench(object):
         self.sim_lengths = sim_lengths
         self.running_times = []
         self.coupling = coupling
-
+    @profile
     def run(self, project_id):
         for model in self.models:
             for conn in self.connectivities:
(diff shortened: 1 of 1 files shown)


maedoc:
I would suggest using cProfile instead of line profiler, then sort by time and compare between profiles.  There should be some clear function calls standing out in the command profile not present in the library profile.

Raghibshams456:
@maedoc please have a look...

maedoc:
I think it might make more sense to do the profiling within the simulator, because the benchmark you have here only measures the performance in command profile, not library profile.  Also I would guess that the simulation which is a submitted operation does not run in the same process as the script, so the profiler won't catch it.

Raghibshams456:
can you suggest on which part of code should I focus if I want to improve 

   1. Simulation wizard
   2. H5 files writing
    3..loading for visualizers from files
actually this is project of gsoc 2021 and I want to work on this..

maedoc:
Ah I see the goal is a bit wider.  If you want to profile across the web server, [Yappi](https://pypi.org/project/yappi/) might be a good choice.  But yes, for profiling the simulator itself, you need would want to use cProfile in the run method.

Raghibshams456:
@maedoc  can u guide me I writing proposal of that project "benchmarking and otimising tvb areas"..any hints
??

liadomide:
@Raghibshams456 I think this is a good start.
I suggest: 
- you create your own branch, maybe directly from tvb-root master branch (or rebased to it), but still containing the changes done to fix `benchmark.py` from the current branch
- for catching simulator execution stack (as pointed by Maedoc above), try to change `SimulatorAdapter.launch_mode` to `AdapterLaunchModeEnum.SYNC_SAME_MEM`
- then indeed, we will need to go towards profiling cherrypy server. Yappi would be a good choice, but also something like Selenium would be a good tool, from my point of view, to identify slow web pages. Here, it is up to you to propose a good solution, as part of your application

PR #398, opened 2021-04-08
TVB-2923 Optimise operations page

bvalean:


liadomide:
I like the principle of a cache, we have been discussing it also in the past for TVB, but we were hold back by the fact that no real benchmarks existed (to say exactly what we gain with what risks), and also not many large DBs were in use.

Anyway, I like the new cache service, but I am not convinced that the many `cache.clear_cache()` calls, spread  in multiple places now (mostly at the service layer, but not only) are reliable enough. Can we say these places are enough? Isn't a better place to put the clearing of the cache within DAO layer? I don't have an answer in mind, just a concern at this point.

bvalean:
Required changes in #403

PR #400, opened 2021-04-12
WIP TVB-2913: Moved storage related files to a new module

VinczeRobert:
We should review if I moved exactly what needed to be moved. I know I have to add a few more files including authors and a README and I also need to write a new file updates script to update the written_by attribute of the H5 files. I also need to deal with the tests.


VinczeRobert:
I am closing this PR because I kind of got it wrong and I a started a new one and I will open a new PR for the new one.

PR #403, opened 2021-04-29
HBP-157 Split submit simulation and running simulation in two different openshift services

bvalean:
This PR contains also changes from another PRs: https://github.com/the-virtual-brain/tvb-root/pull/395 and https://github.com/the-virtual-brain/tvb-root/pull/398.

TODO:
- [x] update openshift deploy yml file
- [x]  review/clean code


framework_tvb/tvb/core/entities/storage/session_maker.py
```diff
@@ -53,10 +54,13 @@
 
 def build_db_engine():
     if TvbProfile.current.db.SELECTED_DB == 'postgres':
-        ### Control the pool size for PostgreSQL, otherwise we might end with multiple
-        ### concurrent Python processes failing because of too many opened connections.
-        DB_ENGINE = create_engine(TvbProfile.current.db.DB_URL, pool_recycle=5, max_overflow=1,
-                                  pool_size=TvbProfile.current.db.MAX_CONNECTIONS)
+        if TvbProfile.current.db.MAX_CONNECTIONS == 0:
+            DB_ENGINE = create_engine(TvbProfile.current.db.DB_URL, poolclass=NullPool)
```
popaula937, line 58: Also here pls add a comment, just to state that in this case, we are using a PGbouncer

framework_tvb/tvb/core/services/backend_clients/standalone_client.py
```diff
@@ -122,6 +125,7 @@ def run(self):
         storage_interface.check_and_delete(project_folder)
 
         # Give back empty spot now that you finished your operation
+        # cache.clear_cache()
```
popaula937, line 128: If caching is moved to a separate branch, we should remove the commented code

popaula937 (approved):

VinczeRobert:
There are 2 tests on Jenkins that are failing. Also there is the test_launch_multiple_operations test that fails on Github. The latter seems to be because due to the optimizations made by Paula the 2 x 20 s sleep will make the operations be finished, so the last assert fails. If I put 2 x 10 s the test passes. Can you please fix them @bvalean ?

PR #408, opened 2021-05-18
TVB-2883 Show Branch button only if simulation is finished.

bvalean:


framework_tvb/tvb/interfaces/web/templates/jinja2/burst/burst_history.html
```diff
@@ -26,7 +26,7 @@ <h4>History</h4>
                             {% endif %}
                             <p><button class="action action-copy" onclick="copyBurst({{ burst.id }}, '{{ first_fragment_url }}')" title="Create a copy of the current simulation configuration.">
                                 Copy</button></p>
-                            {% if burst.range1 == None %}
+                            {% if burst.range1 == None and burst.status == 'finished' %}
```
popaula937, line 29: Maybe add a property on burst to check whether it is finished with success instead of keeping the hardcoded string in html

PR #410, opened 2021-05-19
TVB-2836 Use jQuery to show/hide burst extension in the burst history fragment

bvalean:


popaula937 (approved):
Remove the commented code before merge

PR #411, opened 2021-05-19
TVB-2909 Set fragment url also for the previous button

bvalean:


popaula937 (commented):
After testing these changes I think the _Cancel_ button works well, but I do not see improvements for the _Back_ button.  
Please check this again, if you look in the browser network, when clicking back you get the status code: _200 OK (from disk cache)_, maybe we could do something to force no caching.
Otherwise, we could add only the _Cancel_ buttons, but in that case, please clean the code.

framework_tvb/tvb/interfaces/web/templates/jinja2/simulator_fragment.html
```diff
@@ -49,7 +49,7 @@
     {% endif %}
     {% if renderer.include_setup_surface_model %}
             <button type="button" title="Set up model parameters for surface-based simulations" class="btn btn-primary" tabindex="1"
-               onclick="wizzard_submit(this.parentElement, configureModelParamsOnSurface);" id="configSurfaceModelParam" name="configSurfaceModelParam"
+               onclick="wizzard_submit(this.parentElement, configureModelParamsOnSurface, 'div-simulator-parameters', false);" id="configSurfaceModelParam" name="configSurfaceModelParam"
```
popaula937, line 52: Do we really need to call `wizzard_submit` before launching the _Surface model parameters_ page? We could call directly `configureModelParamsOnSurface` as we do for the other similar pages. We would then avoid going through the `set_model` endpoint that changes the `last_loaded_fragment`.

bvalean: Yes, because without the wizard submit we will always configure de default model in the configure model params on surface page. The set_model endpoint is needed anyway. It is called when the 'Previous' button is pressed.

framework_tvb/tvb/interfaces/web/controllers/simulator/simulator_controller.py
```diff
@@ -259,7 +272,11 @@ def set_model(self, **data):
         session_stored_simulator, is_simulation_copy, is_simulation_load, _ = self.context.get_common_params()
 
         if cherrypy.request.method == POST_REQUEST:
-            self.context.add_last_loaded_form_url_to_session(SimulatorWizzardURLs.SET_MODEL_PARAMS_URL)
+            set_next_wizard = True
+            if SimulatorController.KEY_KEEP_SAME_SIM_WIZARD in data:
+                set_next_wizard = False
+            if set_next_wizard:
```
popaula937, line 278: Condition could be based on:
keep_same_wizard = `SimulatorController.KEY_KEEP_SAME_SIM_WIZARD in data`
if keep_same_wizard is False: ...

popaula937 (approved):

PR #412, opened 2021-05-27
TVB-2911: Implement a BIDS importer

popaula937:
So far I have extended the script written by Bogdan which was importing each connectivity. Now it also imports the FC.mat as a CorrelationCoefficients and HRF.csv file as a TS (with a good amound of hardcoding).

We need to store metadata in a JSON file and probably other files as well. For example in the HRF.csv file I suppose we only have the array_data matrix, but we also need a time matrix, so maybe another .csv file?

### DIFF
--- a/framework_tvb/tvb/config/init/alembic/versions/32d4bf9f8cab_update_db_002.py
+++ b/framework_tvb/tvb/config/init/alembic/versions/32d4bf9f8cab_update_db_002.py
@@ -1,0 +1,35 @@
+"""
+Revision ID: 32d4bf9f8cab
+Create Date: 2021-06-03
+
+"""
+from alembic import op
+from sqlalchemy.engine.reflection import Inspector
+from sqlalchemy import Column, Boolean
+import uuid
+import json
+
+from sqlalchemy.sql.schema import UniqueConstraint
+from tvb.basic.profile import TvbProfile
+from tvb.core.neotraits.db import Base
+from tvb.basic.logger.builder import get_logger
+
+# revision identifiers, used by Alembic.
+revision = '32d4bf9f8cab'
+down_revision = 'ec2859bb9114'
+
+conn = op.get_bind()
+
+LOGGER = get_logger(__name__)
+
+
+def upgrade():
+    new_column = Column('has_valid_time_series', Boolean)
+    op.add_column('DataTypeMatrix', new_column)
+
+
+def downgrade():
+    """
+    Downgrade currently not supported
+    """
+    pass
--- a/framework_tvb/tvb/config/init/alembic/versions/32d4bf9f8cab_update_db_002.py
+++ b/framework_tvb/tvb/config/init/alembic/versions/32d4bf9f8cab_update_db_002.py
@@ -24,7 +24,7 @@ LOGGER = get_logger(__name__)
 
 
 def upgrade():
-    new_column = Column('has_valid_time_series', Boolean)
+    new_column = Column('has_valid_time_series', Boolean, default=True)
     op.add_column('DataTypeMatrix', new_column)
 
 
--- a/framework_tvb/tvb/interfaces/web/controllers/project/project_controller.py
+++ b/framework_tvb/tvb/interfaces/web/controllers/project/project_controller.py
@@ -48,7 +48,6 @@ from pathlib import Path
 import requests
 
 from tvb.adapters.exporters.export_manager import ExportManager
-from tvb.adapters.uploaders.tumor_dataset_importer import prepare_tumor_dataset_for_download, import_tumor_datatypes
 from tvb.config.init.introspector_registry import IntrospectionRegistry
 from tvb.core.entities.filters.factory import StaticFiltersFactory
 from tvb.core.entities.load import load_entity_by_gid
@@ -79,11 +78,6 @@ class ProjectController(BaseController):
     KEY_OPERATION_FILTERS = "operationfilters"
     NODE_OPERATION_TYPE = "operation"
     NODE_OPERATION_GROUP_TYPE = "operationGroup"
-
-    TUMOR_DATASET_URL = 'https://kg.ebrains.eu/proxy/export?container=https://object.cscs.ch/v1/AUTH_6ebec77683fb4' \
-                        '72f94d352be92b5a577/hbp-d000001_TVB_brain_tumor_pub'
-    MAXIMUM_DOWNLOAD_RETRIES = 3
-    SLEEP_TIME = 3
 
     def __init__(self):
         super(ProjectController, self).__init__()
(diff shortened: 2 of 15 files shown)


tvb_storage/tvb/storage/h5/file/files_helper.py
```diff
@@ -244,6 +244,12 @@ def get_allen_mouse_cache_folder(self, project_name):
         self.check_created(folder)
         return folder
 
+    def get_tumor_dataset_folder(self, project_name):
+        project_folder = self.get_project_folder(project_name)
```
popaula937, line 248: I suggest we download the dataset directly under the TVB folder, I think it should be available for import in any project once it is downloaded. What do you think?

popaula937 (commented):
The solution works nicely, but 2 points need a bit more work:
- it generates one operation for each datatype
- there is no feedback regarding the main operation (during download + unzip) until the first import operation starts

How about we define an Importer for the whole Tumor import? Its view_model could contain one field with the link towards ebrains. 
We could call the importers for the datatypes from here without launching an operation for each..or at least group the operations by subject?

tvb_storage/tvb/storage/h5/file/files_helper.py
VinczeRobert: I agree.

VinczeRobert:
> The solution works nicely, but 2 points need a bit more work:
> 
> * it generates one operation for each datatype
> * there is no feedback regarding the main operation (during download + unzip) until the first import operation starts
> 
> How about we define an Importer for the whole Tumor import? Its view_model could contain one field with the link towards ebrains.
> We could call the importers for the datatypes from here without launching an operation for each..or at least group the operations by subject?

We could define an Importer for the whole Tumor import, but then we would have a view model and a form class with nothing that we need. I understand that it would import everything as a single operation which in theory is correct cause it is one operation after all, but this still won't solve the feedback part, actually it will make it worse.

popaula937:
> > The solution works nicely, but 2 points need a bit more work:
> > 
> > * it generates one operation for each datatype
> > * there is no feedback regarding the main operation (during download + unzip) until the first import operation starts
> > 
> > How about we define an Importer for the whole Tumor import? Its view_model could contain one field with the link towards ebrains.
> > We could call the importers for the datatypes from here without launching an operation for each..or at least group the operations by subject?
> 
> We could define an Importer for the whole Tumor import, but then we would have a view model and a form class with nothing that we need. I understand that it would import everything as a single operation which in theory is correct cause it is one operation after all, but this still won't solve the feedback part, actually it will make it worse.

Why worse? The feedback to the user will be on the operations page, as they can see the import operations is still in progress, this being the usual feedback for any other operation.

framework_tvb/tvb/interfaces/web/controllers/project/project_controller.py
```diff
@@ -36,13 +36,17 @@
...
 import cherrypy
 import formencode
 import tvb.core.entities.model.model_operation as model
+import os
+import time
 from cherrypy.lib.static import serve_file
 from formencode import validators
 from simplejson import JSONEncoder
+from pathlib import Path
+import requests
```
popaula937, line 48: Are these imports used?

framework_tvb/tvb/interfaces/web/static/js/tumor_dataset_import.js
```diff
@@ -0,0 +1,42 @@
...
+function importTumorDataset(projectId){
+    displayMessage("Downloading Tumor Dataset from Ebrains has started." +
+        " This operation is going to take a while, please wait...")
+    doAjaxCall({
+        type: "POST",
+        /* TODO: Would be nice not to hardcode the algorithm id */
+        url: "/project/launchloader/" + projectId + "/29",
+        success: function(response){
+            if (JSON.parse(response).is_downloaded){
+                displayMessage("The Tumor Dataset has been succesfully imported!");
```
popaula937, line 31: Are these messages still working as expected?

framework_tvb/tvb/interfaces/web/templates/jinja2/project/structure.html
```diff
@@ -29,6 +31,12 @@ <h3>Data Structure</h3>
                                 title="Import ">Upload
                         </button>
                     </li>
+                    <li>
+                        <button tabindex='50' type="submit" class="action action-upload"
+                                onclick="importTumorDataset({{ project.id }})"
```
popaula937, line 36: From the server, we could send here the tumor importer algorithm ID in the same manner as we send the project ID. Or we could also send the whole link for the Ajax call (this template is rendered by `ProjectController.editstructure`)

framework_tvb/tvb/interfaces/web/static/js/tumor_dataset_import.js
```diff
@@ -0,0 +1,42 @@
...
+ * .. moduleauthor: Robert Vincze <robert.vincze@codemart.ro>
+ **/
+
+
+function importTumorDataset(projectId){
+    displayMessage("Downloading Tumor Dataset from Ebrains has started." +
+        " This operation is going to take a while, please wait...")
+    doAjaxCall({
+        type: "POST",
+        /* TODO: Would be nice not to hardcode the algorithm id */
```
popaula937, line 27: See comment in structure.html as a suggestion to avoid this.

framework_tvb/tvb/interfaces/web/controllers/project/project_controller.py
VinczeRobert: No, I removed them.

framework_tvb/tvb/interfaces/web/static/js/tumor_dataset_import.js
VinczeRobert: The error part should work, but the success part won't since launchloader doesn't return anything. I removed the success clause.

framework_tvb/tvb/interfaces/web/templates/jinja2/project/structure.html
VinczeRobert: Thanks for the idea, I added the algorithm id there.

VinczeRobert:
I remembered we also need a new db migration script because I added a new parameter in the DataTypeMatrix column.

framework_tvb/tvb/interfaces/web/controllers/project/project_controller.py
```diff
@@ -515,11 +516,14 @@ def editstructure(self, project_id=None, first_level=None, second_level=None,
         self._mark_selected(selected_project)
         data = self.project_service.get_filterable_meta()
         filters = StaticFiltersFactory.build_datatype_filters(selected=visibility_filter)
+        tumor_import_algorithm = dao.get_algorithm_by_module('tvb.adapters.uploaders.tumor_dataset_importer',
+                                                             'TumorDatasetImporter')
```
bvalean, line 520: You can use the TumorDatasetImporter.__module__ and TumorDatasetImporter.__class__ instead of hardcoded strings.

bvalean:
Please resolve conflicting files before we can merge it.

VinczeRobert:
I think it is okay to be merged now if the tests pass.

framework_tvb/tvb/config/init/alembic/versions/32d4bf9f8cab_update_db_002.py
```diff
@@ -0,0 +1,35 @@
...
+down_revision = 'ec2859bb9114'
+
+conn = op.get_bind()
+
+LOGGER = get_logger(__name__)
+
+
+def upgrade():
+    new_column = Column('has_valid_time_series', Boolean)
+    op.add_column('DataTypeMatrix', new_column)
```
popaula937, line 28: Shouldn't this get a default value?

VinczeRobert: OH yes, it should.

PR #414, opened 2021-05-27
TVB-2868: Update license text

VinczeRobert:
I have some statements/questions before we merge this.

1. I have added the license text to the empty __init__ files as well.
2. I have added the license to some library files made by Marmaduke and two test files written by Borana.
3. Some Javascript files have license from other places. Should we check if we need to update for those? Also, there is colorpicker.js, colorSchemeComponent.js, d3.v3.js, glUtils.js, sylvester.js (might be more I skipped) which I can't decide if are ours or from somewhere else but they don't have the license text.
4. Only the base html template has the license text, should we add it to the others as well?
5. Excepting the python, HTML, javascript and CSS files should we add the license text to other files as well (like configuration files, Docker files, shell scripts etc?).
6. I tried to add the license to the very few .xml and .xsd files that we have, but it doesn't work (a test failed because of that) and in PyCharm there is a warning saying "Xml declaration should precede all document content".
7. Probably it's not important but after the license text should we have 1 x # or 2 x #? In most places there are 2.
8. Anything else I have missed?


VinczeRobert:
Hmm, the failed test shows that I cannot add a comment like that in an xml file.

framework_tvb/tvb/adapters/portlets/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+#
+#   CITATION:
+# When using The Virtual Brain for scientific publications, please cite it as follows:
+#
+#   Paula Sanz Leon, Stuart A. Knock, M. Marmaduke Woodman, Lia Domide,
+#   Jochen Mersmann, Anthony R. McIntosh, Viktor Jirsa (2013)
+#       The Virtual Brain: a simulator of primate brain network dynamics.
+#   Frontiers in Neuroinformatics (7:10. doi: 10.3389/fninf.2013.00010)
+#
+#
```
liadomide, line 29: This entire 'portlets' module should be removed. I suggest a second small PR for cleanup

framework_tvb/tvb/config/init/alembic/versions/ec2859bb9114_update_db_001.py
```diff
@@ -1,10 +1,40 @@
...
+#   CITATION:
+# When using The Virtual Brain for scientific publications, please cite it as follows:
+#
+#   Paula Sanz Leon, Stuart A. Knock, M. Marmaduke Woodman, Lia Domide,
+#   Jochen Mersmann, Anthony R. McIntosh, Viktor Jirsa (2013)
+#       The Virtual Brain: a simulator of primate brain network dynamics.
+#   Frontiers in Neuroinformatics (7:10. doi: 10.3389/fninf.2013.00010)
+#
+#
+
```
liadomide, line 30: For autogenerated files we can skip the header, unless we have written/edited this file after auto-generation (as it seems to be the case here).

framework_tvb/tvb/core/portlets/__init__.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
```
liadomide, line 1: This module should also be marked deprecated or removed directly as part of small 'cleanup' PR

framework_tvb/tvb/interfaces/web/static/js/projectTree.js
```diff
@@ -4,7 +4,7 @@
  * TheVirtualBrain-Scientific Package (for simulators). See content of the
```
liadomide, line 4: I think that most of the JS files have a wrong header. 
It currently says "TheVierualBrain-Scientific package..." but it should be the text with "TheVirtualBrain-Framework Package. ....." (like in users_controller.py for example).
Could you please fix them?
All files under "framework_tvb" should have the header with the "...Framewrok Package ..."

framework_tvb/tvb/interfaces/web/static/style/base.css
```diff
@@ -4,7 +4,7 @@
  * TheVirtualBrain-Scientific Package (for simulators). See content of the
```
liadomide, line 4: CSS files seem to have the same problem (not new in this commit, but from before) like the JS files.
They are clearly "framework" related, not "simulator"

framework_tvb/tvb/tests/framework/adapters/analyzers/bct_test.py
```diff
@@ -6,7 +6,7 @@
 # TheVirtualBrain-Scientific Package (for simulators). See content of the
```
liadomide, line 6: Header under some tests in framework seems to also be wrong (mentioning "scientific ... simulator" instead of "...framework...". Could you please adjust these headers also?

scientific_library/tvb/analyzers/__init__.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
+#
+#
+# TheVirtualBrain-Framework Package. This package holds all Data Management, and
```
liadomide, line 4: Here the text should be different, as we are under scientific library

liadomide (changes requested):
The current changes are very good.
But I discovered an older problem (a bit also extended with this commit) of mixing the header for scientific package with the one for the framework. I suggest we fix those in the current PS also.

liadomide:
> 1. I have added the license text to the empty **init** files as well.

Not necessary, but also not wrong. Your call if to be kept or not.

> 2. I have added the license to some library files made by Marmaduke and two test files written by Borana.

Good

> 3. Some Javascript files have license from other places. Should we check if we need to update for those? Also, there is colorpicker.js, colorSchemeComponent.js, d3.v3.js, glUtils.js, sylvester.js (might be more I skipped) which I can't decide if are ours or from somewhere else but they don't have the license text.

`colorpicker.js, colorSchemeComponent.js` are ours, and should have our license text. 
The others mentions are copied, so we don't need to put our license.

> 4. Only the base html template has the license text, should we add it to the others as well?

ok as it is (only in base html templates), otherwise the merged resulting HTML responses would be unnecessarily large.

> 5. Excepting the python, HTML, javascript and CSS files should we add the license text to other files as well (like configuration files, Docker files, shell scripts etc?).

I think not. Do you have an argument otherwise?

> 6. I tried to add the license to the very few .xml and .xsd files that we have, but it doesn't work (a test failed because of that) and in PyCharm there is a warning saying "Xml declaration should precede all document content".

True. No need to add our license in such files.

> 7. Probably it's not important but after the license text should we have 1 x # or 2 x #? In most places there are 2.

Ideally it is the same. I am fine either way.

framework_tvb/tvb/core/portlets/__init__.py
VinczeRobert: Ok, I will remove the portlets module in a separate PR after this one.

framework_tvb/tvb/config/init/alembic/versions/ec2859bb9114_update_db_001.py
VinczeRobert: This is the only case I know of when the file is auto-generated but we write/edit it. The other alembic related file which was autogenerated is env.py but we didn't have to modify that, that's why it does not have the license text.

framework_tvb/tvb/interfaces/web/static/js/projectTree.js
VinczeRobert: In some places the whole first paragraph is wrong, for example in fcd_adapter. I think only the first paragraph is different for the library and framework tests.

liadomide: Correct, only the first paragraph is different between library and framework.

For tvb-build, tvb-documentation, tvb-rest, etc the header can be the same as the one in framework.

VinczeRobert:
I think it should be okay now. @liadomide  please have another look when you have time. I just have 3 more questions:
1. In the library there are some cuda related files and some c files, do those need the license header?
2. Why do we have the .__init__.py file in some html folders? 
3. For the files that are not ours, should we care about updating the license text if necessary?

framework_tvb/tvb/interfaces/web/static/js/projectTree.js
liadomide: You are right about

> I didn't see JS files with the wrong header, they actually started with "TheVirtualBrain-Framework Package.".. I completed the headers where the CITATION part was missing.

You are right about the header, is not switched in JS and CSS files. I didn't look carefull enough before.
I think that in the CSS and JS files, the citation was deliberately left out, to minimize the size. If it is not too much trouble I would in fact prefer to keep it out ...  :-|

liadomide:
> 1. In the library there are some cuda related files and some c files, do those need the license header?

Files written by INS Marseille (Marmaduke, Borana...) should be under the same license, thus the answer is yes. Just make sure the comment is not breaking the files.

> 2. Why do we have the .**init**.py file in some html folders?

I am not sure, but it is probably because we want to access those folders as Python packages. Either that, or by mistake 📦 

> 3. For the files that are not ours, should we care about updating the license text if necessary?

Do you mean files like D3, JQuery? Those should already have a license included in TVB_Distribution. 
If they came with a header, we don't need to change that, unless we change the file itself. 
I see you recently added a header to some flot&jquery files. Those changes should be ok, as long as you put the correct copyright.

VinczeRobert:
> Files written by INS Marseille (Marmaduke, Borana...) should be under the same license, thus the answer is yes. Just make sure the comment is not breaking the files.

Several files are commited by a user named DeLaVlag so I don't know who that belongs to. Also, for some files Pycharm already gives some errors  (for example tmpl8_CUDA.py) so It would be hard to know if the comment would be breaking the file.

> I am not sure, but it is probably because we want to access those folders as Python packages. Either that, or by mistake

They are needed... I tried deleting them and wrong things happen.

> Do you mean files like D3, JQuery? Those should already have a license included in TVB_Distribution.
> If they came with a header, we don't need to change that, unless we change the file itself.
> I see you recently added a header to some flot&jquery files. Those changes should be ok, as long as you put the correct copyright.

This part should be okay then.

framework_tvb/tvb/tests/framework/adapters/analyzers/bct_test.py
VinczeRobert: I fixed those.

framework_tvb/tvb/interfaces/web/static/style/base.css
VinczeRobert: They are okay now.

framework_tvb/tvb/interfaces/web/static/js/projectTree.js
VinczeRobert: Ok, but there were some js and css files were the citation was left there.  I am going to remove the citation from each js and css files then.

VinczeRobert:
There is a third citation for the contributors package, but those files are okay. Also, there a few files in the documentation module which have the library citation and I think it is okay because those files seem to be more library related. Anything else to do?

maedoc:
> user named DeLaVlag so I don't know who that belongs to. Also, for some files Pycharm already gives some errors (for example tmpl8_CUDA.py) so It would be hard to know if the comment would be breaking the file.

DeLaVlag is Michiel van der Vlag from Juelich who contributed on the templ8_CUDA.py which despite its extension is a Mako template for CUDA C++, so the license comment should use `//`.  He and Sandra Diaz (another author on those modules) agreed to contribute under the TVB license.

maedoc:
In retrospect, the Mako templates themselves should be under TVB license, however code that they generate should not.  I am not sure if there is an easy way to handle this apart from a comment after the license to that effect.

VinczeRobert:
@maedoc I added the license text to the .mako files and to tmpl8_CUDA and tmpl8_regTVB. Is there any other file in that area that I should add it to?

popaula937 (approved):
Looks good, I noticed a few citations left in JS files

framework_tvb/tvb/adapters/portlets/__init__.py
popaula937: Will be taken care of in task [TVB-2936](https://req.thevirtualbrain.org/browse/TVB-2936)

PR #416, opened 2021-06-03
TVB-2937. Speed-up the data structure page

popaula937:
The method `get_project_structure` is called when loading the **Data Structure** page, and a short profiling session shows that for around 200 datatypes, this method takes around 6.8 seconds to compute, out of which, around 5.2 seconds is spent calling the `dt_entity = dao.get_datatype_by_gid(dt.gid) ` inside the loop.
By avoiding that DAO call and the following check, the execution time goes under 1 second (0.8 seconds).

I am not sure what problem does this check for `dt_entity is None` solve, but from the commit message it seems to be rarely happening, so I am wondering whether we could find another solution that prevents the `get_project_structure` slow-down.
@liadomide do you remember more about this check?

### DIFF
--- a/framework_tvb/tvb/core/services/project_service.py
+++ b/framework_tvb/tvb/core/services/project_service.py
@@ -465,10 +465,6 @@ class ProjectService:
             data = {}
             is_group = False
             group_op = None
-            dt_entity = dao.get_datatype_by_gid(dt.gid)
-            if dt_entity is None:
-                self.logger.warning("Ignored entity (possibly removed DT class)" + str(dt))
-                continue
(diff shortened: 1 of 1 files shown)


liadomide (approved):
It looks to me like left-over code. I don't see a logical need for that part. 
Also, I am not sure about its previous usage/state, but I recommend merging this if all tests are ok.

popaula937:
> It looks to me like left-over code. I don't see a logical need for that part.
> Also, I am not sure about its previous usage/state, but I recommend merging this if all tests are ok.

We have a similar situation for the **Operations** page, so I will remove this code from both in this PR.

PR #418, opened 2021-06-04
TVB-2867: Fix models under tvb_contrib

VinczeRobert:
- [x] The only one I haven't been able to fully fix is the Brunel Wang one because I don't know what is the situation with the LookupTable class and its 2 subclasses.

Also, I fixed the obvious issues but I think someone else (who has worked on these) should also do a review.

- [x] PS: I accidentally did not make this branch from master so we should merge the PR for TVB-2911 first so I can do a rebase before merging this one.

### DIFF
--- a/tvb_contrib/tvb/contrib/simulator/models/epileptor.py
+++ b/tvb_contrib/tvb/contrib/simulator/models/epileptor.py
@@ -34,10 +34,12 @@ The Epileptor model
 """
 
 import numpy
+
 from tvb.simulator.common import psutil, get_logger
-LOG = get_logger(__name__)
 from tvb.basic.neotraits.api import NArray, Range, List, Final
 import tvb.simulator.models as models
+
+LOG = get_logger(__name__)
 
 
 class HMJEpileptor(models.Model):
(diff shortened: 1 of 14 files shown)


popaula937:
TVB-2911 is merged now, so please fix conflicts here

tvb_contrib/tvb/contrib/simulator/models/epileptor.py
```diff
@@ -154,28 +156,16 @@ class HMJEpileptor(models.Model):
...
-#        order=10)
-
-
     def __init__(self, **kwargs):
         """
         """
-
-        LOG.info("%s: init'ing..." % (str(self),))
-
         super(HMJEpileptor, self).__init__(**kwargs)
```
popaula937, line 162: Looks like we could take out the `__init__` from contrib models to make them consistent with the library models, the attributes set inside `__init__` could move directly on the class.

VinczeRobert: Is it okay now?

PR #419, opened 2021-06-09
NFS-18. Review Cortex and full_region_map usage

popaula937:
Starting from a review over the `PseudoCortex` class in NFS code, I am proposing the following changes to the TVB library:
- let the `Cortex` be responsible for computing the full region mapping, in case the subcortical regions are present in the `Connectivity`, but not in the `RegionMapping`, as it already references both of them
- this computation happens internally in the `Cortex` and the `Cortex.region_mapping` property always exposes the full region mapping (pls let me know in case there are places that would need only the cortical region mapping that I might have missed) 
- throughout the simulator, we only use the `Cortex.region_mapping` as a single point to access the region mapping (for NFS it would return the full region mapping of the `RegularSurface`)
- have a property on `Connectivity` to compute the subcortical indices

I added some tests, where I add 2 dummy subcortical regions to the `Connectivity` before calling the `Cortex.region_mapping` or launching simulations.

Let me know what are your thoughts on these, I would like to start bringing generic changes from the **spherical_harmonics** branch to **master**.

### DIFF
--- a/scientific_library/tvb/datatypes/connectivity.py
+++ b/scientific_library/tvb/datatypes/connectivity.py
@@ -36,12 +36,12 @@ The Connectivity datatype.
 
 """
 
+from copy import copy
+
 import numpy
 import scipy.stats
-from copy import copy
-from tvb.basic.readers import ZipReader, H5Reader, try_get_absolute_path
 from tvb.basic.neotraits.api import Attr, NArray, List, HasTraits, Int, narray_summary_info
-
+from tvb.basic.readers import ZipReader, H5Reader, try_get_absolute_path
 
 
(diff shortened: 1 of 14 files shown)


liadomide (approved):
This looks good to me. It has become critical to have this fix in master for simulations with subcortical regions.
Conflicts only will need to be solved first

liadomide (approved):

liadomide:
We marked this as a bug in Jira, as users complained about Bold ROi monitor not working when subcortical regions are present.
https://req.thevirtualbrain.org/browse/TVB-2970
This bug is fixed now with this commit PR

PR #423, opened 2021-06-16
TVB-2943: Fix deletion of PSE simulations.

VinczeRobert:


bvalean (approved):

liadomide (commented):
Is this code working in postgresql? I think there is a FK between operations and bursts tables, and removing the operation before the query for burst would either fail on PG, or delete the burst in cascade. Or am I missing something?

bvalean:
Yes. It is working also using postgresql. We have the on delete set to 'SET NULL' on fk_simulation FK on BurstConfig.

liadomide:
> Yes. It is working also using postgresql. We have the on delete set to 'SET NULL' on fk_simulation FK on BurstConfig.

Ok, so it won't throw exception, but what will the following query return? Isn't always empty result?
Dao.get-burst-for-operation-id

PR #424, opened 2021-06-17
TVB-2944 Add missing dependency in library setup.py

bvalean:


popaula937:
Pls check that tests are passing at a clean installation

PR #427, opened 2021-06-19
TVB-2929:Updated pyplot in place of pylab

Ishan2601:
Updated the code for plotters to use matplotlib.pyplot in place of pylab

### DIFF
--- a/scientific_library/tvb/simulator/plot/phase_plane_interactive.py
+++ b/scientific_library/tvb/simulator/plot/phase_plane_interactive.py
@@ -62,7 +62,7 @@ Example specifying a Model and stochastic sample trajectories::
 #      check for leaks or look into "forcing" cleanup...
 
 import numpy
-import pylab
+import matplotlib.pyplot as plt
 import colorsys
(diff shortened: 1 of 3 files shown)


i-Zaak:
Great, cool to have the the phase plane interactive updated. Would be also great if it worked with `%matplotlib inline` as the interactive plots in jupyterhub environments are fragile. 

I've stitched a bare-bones version for the EBRAINS tutorials, you can take a peak for inspiration: https://gitlab.ebrains.eu/fousekjan/tvb_tutorials/-/blob/mcgill_summer_school/phase_plane.py 

The main difference is, that the canvas is not clickable and only one trajectory can be plotted.

Ishan2601:
@i-Zaak  Thanks for sharing your bare-bones version.  I'll try to make it work with `%matplotlib inline`.

popaula937:
Thank you @i-Zaak for the suggestion, I am merging this version in the TVB-2929 branch and @Ishan2601 will soon provide a version that works with `%matplotlib inline`

PR #430, opened 2021-06-23
TVB-2942: Update DataExchange.rst

VinczeRobert:
This RST hadn't been updated in a while, and some changes happened over the time in TVB's IN/OUT files.
This is a small correction to the Data Exchange descriptor.

### DIFF
--- a/tvb_documentation/manuals/UserGuide/DataExchange.rst
+++ b/tvb_documentation/manuals/UserGuide/DataExchange.rst
@@ -22,7 +22,7 @@ at different levels and formats:
   surface, connectivity, time series) between two |TVB| instances or between |TVB|
   and an external application. As you may note later, depending on the targeted
   application, data can be exchanged in a custom |TVB| format or a commonly used
-  format used in the neuroscience community (e.g. CFF, GIFTI, NIFTI ...)
+  format used in the neuroscience community (e.g. H5, GIFTI, NIFTI ...)
 
 .. warning::
 
@@ -38,16 +38,16 @@ Before proceeding with more details about data exchange, it would be helpful to
 give you an idea how |TVB| stores its data. Basically there are two major
 storage areas:
 
-1. *Database* - where general information/metadata are stored and relation 
-   between stored elements (e.g. assignment of data to a project, data metadata
+1. *Database* - where general information/metadata and relations between
+   stored elements are stored (e.g. assignment of data to a project, data metadata
    - creation date, owner, etc...)
 
 #. *Disk* - here we store "the real" data in a HDF5 format 
    (http://www.hdfgroup.org/HDF5). This means that for each data type (e.g.
-   surface, connectivity, time series) we store on the disk, it is given a 
-   folder structure in an HDF5 / H5 file which contains all data (e.g. vertices,
-   triangles, etc ...). This format has the following advantages which made it an
-   optimal solution for our product:
+   surface, connectivity, time series) we store on the disk, we have a folder
+   containing a HDF5 / H5 file which contains all data (e.g. vertices,
+   triangles, etc.). This format has the following advantages which made it
+   an optimal solution for our product:
 
    - can store huge pieces of data and can access it very fast (even random
      access)
@@ -58,15 +58,15 @@ storage areas:
 
 |
 
-An important aspect of |TVB| storage is that each data/datatype has associated a
-GUID, which makes it unique on every system where that data exists.
+An important aspect of |TVB| storage is that each data/datatype has a GUID,
+which makes it unique on every system where that data exists.
 
 
 
 Exchange Projects
 -----------------
 
-|TVB| product can be installed both on a server, to be used concurrent by multiple
+|TVB| product can be installed both on a server, to be used concurrently by multiple
 users, but also as a standalone application on a desktop/laptop for personal use.
 Specifically for the second scenario, there was an important request to allow
 people to exchange data. So, |TVB| has a mechanism to export and import an entire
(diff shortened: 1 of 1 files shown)


tvb_documentation/manuals/UserGuide/DataExchange.rst
```diff
@@ -83,13 +83,13 @@ File Format
...
 Export results are a ZIP file (named: date + project name), containing in a
-folder structure, all the details about the project. More specifically, it
+folder structure all the details about the project. More specifically, it
 contains:
 
 - A root level XML file with details about the project itself
 - Folders for each operation performed as part of the project
-- Operation folder contains an XML file with details of the operation and 
-  a set of H5/HDF5 files for each data type generated during that operation.
+- Operation folder containts a set of H5/HDF5 files for each
```
popaula937, line 91: Pls mention the ViewModel instead of old XML with details of the operation

tvb_documentation/manuals/UserGuide/DataExchange.rst
```diff
@@ -122,13 +122,19 @@ Export Data
...
+Export in **TVB Format** (without links) results in a file with a format specific to |TVB|; it is not
 a standard format that can be used automatically by other software. This is 
 basically HDF5/H5 format
 [`http://www.hdfgroup.org/HDF5 <http://www.hdfgroup.org/HDF5>`_] which, for each
 data type, contains both data and metadata. These files can be easily opened in
 Python / Matlab / Java / C++ for additional processing.
 
+Export in **TVB Format with links** will export not only the chosen file, but all other files
+which are linked to this datatype as well. This second type of export will result in a ZIP file,
+which contains an HDF5/H5 file for each exported datatype.
```
popaula937, line 136: Pls add as example the export of RegionMapping that results in a ZIP file with RM, Conn and Surface

tvb_documentation/manuals/UserGuide/DataExchange.rst
```diff
@@ -190,24 +194,6 @@ In correlation with export operations, |TVB| interface allows import of data in
...
-   must be imported and then process the file accordingly. More specifically, it
-   tries to find an XML file, within the ZIP file, describing the operation(s)
-   that generated the data types and the list of HDF5 files for each datatype.
-
-#. If user uploads a simple HDF5/H5 file, the system assumes that a simple data
-   type is imported and tries to process the file accordingly. Basically it
-   reads the metadata stored in the root node group and determines the data type
-   (e.g. connectivity, time series ...). Based on the detected type of data, the
-   rest of the details are filled and the object is stored in the database.
-
```
popaula937, line 207: Why take this info out? We should add the case of importing the ZIP from Export Datatype with links

tvb_documentation/manuals/UserGuide/DataExchange.rst
```diff
@@ -149,14 +155,12 @@ File Format
 As a result of a Simulation or Analyze function, |TVB| can generate either a data
 type or a group of data types. Each of such structures can be exported as follows:
 
-1. if a simple data type is exported, the result is an HDF5 file which has a root
-   node datatype metadata and leaves the real data.
-#. if a data type group is exported, the result is a ZIP file containing:
+1. if a simple data type is exported, the result is a single HDF5 file.
```
popaula937, line 158: add also case when datatype is exported with links => ZIP

tvb_documentation/manuals/UserGuide/DataExchange.rst
```diff
@@ -194,6 +197,23 @@ In correlation with export operations, |TVB| interface allows import of data in
 |TVB| data type. Depending on the uploaded file format, imported data can be as
 follows:
 
+File Format
+~~~~~~~~~~~
+
+1. If user uploads a ZIP file, the system automatically assumes a datatype group
+   must be imported and then process the file accordingly. More specifically, it
+   tries to find an XML file, within the ZIP file, describing the operation(s)
+   that generated the data types and the list of HDF5 files for each datatype.
```
popaula937: This paragraph should be corrected to describe the current flow, where we have H5 view model instead of XML and when importing a ZIP tvb has to determine whether it contains a DT group or a DT with links

tvb_documentation/manuals/UserGuide/DataExchange.rst
```diff
@@ -275,49 +281,50 @@ ZIP file should include files with the following naming schema and format:
    - contains a matrix of weights
    - any value greater than zero is considered as a connection. You should not have negative values in your weights file.
 
-#. If any file name contains "centres" it will be considered as the container
+2. If any file name contains "centres" it will be considered as the container
```
liadomide, line 284: Why hard-code the numbers? 
When left with # placeholder, these will be automatically numbered when the HTML is generated.

VinczeRobert: When I was updating the GUI tutorials, the numberings were mixed (in some places the numbers were hardcoded, in other places there was the # placeholder), so that is why I thought to keep it consistent everywhere.

liadomide: I understand, but then the right consistency will be not to hard-code numbers, and use placeholders in every place possible.

tvb_documentation/manuals/UserGuide/DataExchange.rst
VinczeRobert: I think it actually looks if there are directories inside the zip file and if there are, it takes it as a DatatypeGroup import, otherwise it is an import with links.

PR #432, opened 2021-06-24
TVB-2886: Put placeholders instead of hardcoded numbers

VinczeRobert:


popaula937:
Just looked over the final generated documentation and these placeholders do not work correctly, please fix them. Attached the below pics as examples, but the numbers are wrong everywhere.
Screen Shot 2021-06-25 at 9 09 24 AM
Screen Shot 2021-06-25 at 9 12 42 AM

VinczeRobert:
Looks like the placeholders work as expected only if there is nothing between the numbered lines (no figures, no titles, nothing). I will keep the numbers.

PR #434, opened 2021-06-24
TVB-2939: Hide TumorDatasetImporter from the list of uploaders

VinczeRobert:


framework_tvb/tvb/core/services/algorithm_service.py
```diff
@@ -220,7 +220,10 @@ def get_upload_algorithms():
         """
         categories = dao.get_uploader_categories()
         categories_ids = [categ.id for categ in categories]
-        return dao.get_adapters_from_categories(categories_ids)
+        algorithm_list = dao.get_adapters_from_categories(categories_ids)
+
+        # We don't want TumorDatasetImporter to be shown among the other importers
+        return list(filter(lambda x: x.classname != 'TumorDatasetImporter', algorithm_list))
```
liadomide, line 226: This is not nice, to hard-code a hide with a specific name.
Why isn't this TumorData Algorithm a creator ? Then it won't appear in the web GUI, and can still be found in DB as an Algorithm to be launched from ProjectController.

VinczeRobert: We always discussed about as an importer because it doesn't create anything new, but if it is okay to make it a creator, I will do that.

liadomide: indeed, from certain definition/language perspectives, this is an "importer", but when  that is becoming a problem in the code, it is ok to review the category.
to support the "creators" group, one can say this algorithm is "creating tumor tvb datatypes, from online bids data"

liadomide (approved):
the code changes look good to me now.
i only wonder why in unit-test it had to be changed from 20 to 18 upload algos count?

VinczeRobert:
> the code changes look good to me now.
> i only wonder why in unit-test it had to be changed from 20 to 18 upload algos count?

No, unfortunately it is not okay. Because I made the class a creator and ZipConnectivityImporter is used inside it, it will treat it as a creator, not as an importer (that is why the numbers drops from 20 to 18).

liadomide:
> Actually first is it put as an uploader algorithm in the database (uploaders are put before creators) and then when it gets to the creators, it will update it as being a creator.

This looks like a bug in the introspection. Don't we look at the superclass also (to inherit abccreator or abcuploader)?

VinczeRobert:
> > Actually first is it put as an uploader algorithm in the database (uploaders are put before creators) and then when it gets to the creators, it will update it as being a creator.
> 
> This looks like a bug in the introspection. Don't we look at the superclass also (to inherit abccreator or abcuploader)?

This happens in import_adapters from introspector_registry.  Basically everything that is defined or imported into the module is taken and then there are 3 conditions: it must be a class, it must not be an abstract class and it most inherit from ABCAdapter. I did a change now adding a 4th condition: to be in the current module.

VinczeRobert:
@liadomide  can you have a look at my latest commit, please?

There are 4 conditions now, but after the first 2 most of the evaluations will be false, so I think it is okay.

popaula937 (approved):

PR #437, opened 2021-06-25
TVB-2929: Refactored Power Spectra Interactive to use ipywidgets

Ishan2601:
Updated the Power Spectra Interactive Tool.


popaula937:
The power_spectra_interactive looks good from my side. Could you add it to a notebook as well? Maybe the same notebooks where timeseries_interactive is used?

PR #438, opened 2021-06-28
TVB-2947: Create a new fragment for variables of interest to solve branch problem

VinczeRobert:


VinczeRobert:
Unfortunately this is not the only issue with branching. I found that you can't increase the sampling period for an already chosen monitor. You can choose new monitors when branching with non-default sampling period but you can't choose a bigger one for an already chosen monitor...

popaula937:
I think this can be closed if we merge PR #440

VinczeRobert:
> I think this can be closed if we merge PR #440

Okay, I will close it if we are satisfied with the solution provided in PR #440.

PR #439, opened 2021-06-28
TVB-2924: Change some libraries to support Python 3.8 environments

VinczeRobert:
Now simulator_controller tests and flask  should work in Python 3.8 envs as well. I added a more detailed comment on the Jira task.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,7 +22,7 @@ jobs:
         run: echo "PATH=$HOME/.local/bin:$PATH" >> $GITHUB_ENV
 
       - name: create a requirements file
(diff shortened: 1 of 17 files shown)


popaula937:
Have you tested the changes with python 3.7 as well?

VinczeRobert:
> Have you tested the changes with python 3.7 as well?

I did, I can run all the tests. Now for some reason I cannot run the rest related tests in a python 3.8 environment because it says I am outside of the application context.

Maybe someone else could test this as well with a new environment.

popaula937:
I also tested these, the issue you mention happens in the python 3.8 environment with either of flask-restplus or flask-restx libraries. Python 3.7 works well with both.
I suggest we merge these updates so we avoid the deprecated flask-restplus (plus the need to keep werkzeug to an older version). Please add an issue in Jira describing this python 3.8 error.
Before merging, please also take out the fixed version for werkzeug, so we would use the latest version now that we have flask-restx.

popaula937:
Actually, I read a bit about this issue with mocking flask in python 3.8 and found a workaround for our tests to pass:
https://www.gitmemory.com/issue/pallets/flask/3637/655372502

popaula937:
Tests pass now for both 3.7 and 3.8 envs, please check if everything passes also in Jenkins and merge this afterwards.

popaula937 (approved):

PR #440, opened 2021-07-01
Solve simulation branching issues.

VinczeRobert:
This PR proposes a different solution to TVB-2947 and serves to solve the integration step size and monitor sampling periods as well using the same idea. 

So far I have only done it for the integration step size and it works, so I could repeat the process for the other two issues as well.

The idea was suggested by Paula, inspired by the way we disable certain options for the Spatial Average Monitor. 
Here we have one more problem though: if we go Next from a fragment that has a disabled fragment, the code will try to put None on a simulator param which will raise a TraitValueError. This problem can be solved by putting the code in try-except and then putting the read-only param on the form and then retry the method that raised the exception.

### DIFF
--- a/framework_tvb/tvb/adapters/simulator/integrator_forms.py
+++ b/framework_tvb/tvb/adapters/simulator/integrator_forms.py
@@ -65,16 +65,24 @@ class IntegratorForm(Form):
     def get_subform_key(self):
         return SubformsEnum.INTEGRATOR.name
 
-    def __init__(self):
+    def __init__(self, is_dt_disabled=False):
         super(IntegratorForm, self).__init__()
+        self.is_dt_disabled = is_dt_disabled
         self.dt = FloatField(IntegratorViewModel.dt)
+
+    def fill_from_trait(self, trait):
+        # type: (IntegratorViewModel) -> None
+        super(IntegratorForm, self).fill_from_trait(trait)
+
+        if self.is_dt_disabled:
+            self.dt.disabled = True
 
 
 class IntegratorStochasticForm(IntegratorForm):
     template = 'form_fields/select_field.html'
 
-    def __init__(self):
-        super(IntegratorStochasticForm, self).__init__()
+    def __init__(self, is_dt_disabled=False):
+        super(IntegratorStochasticForm, self).__init__(is_dt_disabled)
         self.noise_choices = get_ui_name_to_noise_dict()
         default_noise = list(self.noise_choices.values())[0]
 
--- a/framework_tvb/tvb/adapters/simulator/model_forms.py
+++ b/framework_tvb/tvb/adapters/simulator/model_forms.py
@@ -31,6 +31,7 @@
 from tvb.adapters.simulator.form_with_ranges import FormWithRanges
 from tvb.core.neotraits.forms import Form, ArrayField, MultiSelectField
 from tvb.simulator.models import ModelsEnum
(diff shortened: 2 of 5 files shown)


framework_tvb/tvb/interfaces/web/controllers/simulator/simulator_controller.py
```diff
@@ -341,12 +342,18 @@ def set_integrator(self, **data):
 
     @expose_fragment('simulator_fragment')
     def set_integrator_params(self, **data):
-        session_stored_simulator, is_simulation_copy, is_simulation_load, _ = self.context.get_common_params()
+        session_stored_simulator, is_simulation_copy, is_simulation_load, is_branch = self.context.get_common_params()
 
         if cherrypy.request.method == POST_REQUEST:
-            form = get_form_for_integrator(type(session_stored_simulator.integrator))()
+            form = get_form_for_integrator(type(session_stored_simulator.integrator))(is_branch)
```
popaula937, line 348: How about instead of the try-except block, if is_branch is True, we add the dt with session_stored_simulator.integrator.dt value in the data dict before calling fill_from_post?

VinczeRobert: Yeah, that should work too.

framework_tvb/tvb/adapters/simulator/integrator_forms.py
```diff
@@ -65,16 +65,24 @@ class IntegratorForm(Form):
     def get_subform_key(self):
         return SubformsEnum.INTEGRATOR.name
 
-    def __init__(self):
+    def __init__(self, is_dt_disabled=True):
```
popaula937, line 68: disable_dt=False by default

framework_tvb/tvb/adapters/simulator/model_forms.py
```diff
@@ -107,11 +108,26 @@ class StateVariableRangesForm(Form):
...
+class ModelForm(FormWithRanges):
 
-class Generic2dOscillatorModelForm(FormWithRanges):
+    def __init__(self, model, are_voi_disabled=False):
+        super(ModelForm, self).__init__()
 
-    def __init__(self):
-        super(Generic2dOscillatorModelForm, self).__init__()
+        self.variables_of_interest = MultiSelectField(model.get_class().variables_of_interest)
+        self.are_voi_disabled = are_voi_disabled
```
popaula937, line 117: It does not seem necessary to move the definition of `self.variables_of_interest` on `ModelForm`. You can keep here only the flag and the definition of `fill_from_trait` method.

VinczeRobert: But doesn't it look better this way? We don't define variables_of_interest in every subclass + in the other fragments where they are more subclasses (integrators and monitors) the common param comes first (integration step_size and monitor sampling period) and only then come the specific params depending on the chosen type.

popaula937: But still, you need to send the model as a parameter from each subclass in order to define the variables_of_interest in a commonplace.
I would not change the order of the parameters if we can avoid it.

VinczeRobert: Ok. Pycharm will give a warning, but it works like you say too.

popaula937 (approved):

PR #441, opened 2021-07-01
TVB-2929: Refactored Time Series Interactive Tool using ipywidgets

Ishan2601:
Updated the Time Series Interactive Tool to use ipywidgets.


popaula937:
For the TimeseriesInteractive plotter, it would be nice to have also the possibility to select/unselect all regions at once

popaula937:
Also, could you also update the jupyter notebooks under tvb_documentation/demos that use this plotter to work correctly? Probably they only need the `%matplotlib widget` to be added.

popaula937:
Could we have a subtitle for the region selection section? Eg: _Space labels selection_
Also, could the _select/unselect all_ buttons be displayed somewhere besides this subtitle instead of the labels list?

popaula937:
The timeseries_interactive is used also in simulate_region_stimulus.ipynb, please update that one as well

PR #442, opened 2021-07-02
TVB-2949: Ensure tests are copied to the final tvb-storage package

VinczeRobert:
I don't know if these changes are correct or if they are enough to solve this problem, but so far its the best I could come up with.


VinczeRobert:
I ran python setup.py sdist bdist_wheel locally and the tests were copied to so I guess I managed to solve the problem.

popaula937 (approved):

PR #445, opened 2021-07-08
TVB-2933: Move more functionality from exporters to tvb-storage.

VinczeRobert:


framework_tvb/tvb/adapters/exporters/tvb_linked_export.py
```diff
@@ -74,24 +74,25 @@ def export_data_with_references(self, export_data_zip_path, data_export_folder):
 
         return None, export_data_zip_path, True
```
popaula937, line 76: Could the export method from here compute the filename first and then just make one call to the storage giving the list of datatypes to be exported and the filename for the ZIP?
If the separate methods to copy the dts and then archive them can be put into a single method, could that be reused in other cases? Eg: I see write_zip_folder_with_links and export_datatypes methods in the StorageInterface which seem to be peforming a similar action.

framework_tvb/tvb/adapters/exporters/export_manager.py
```diff
@@ -58,8 +58,6 @@ class ExportManager(object):
     all_exporters = {}  # Dictionary containing all available exporters
     export_folder = None
```
popaula937, line 59: This should have been moved to StorageInterface, and all usages throughout ABCExporter, ExportManager and actual exporters could be removed.

framework_tvb/tvb/adapters/exporters/tvb_linked_export.py
```diff
@@ -55,13 +55,13 @@ def get_supported_types(self):
...
 
-    def export(self, data, data_export_folder, project):
+    def export(self, data, export_folder, project):
         """
         Exports data type:
         1. If data is a normal data type, simply exports storage file (HDF format)
         2. If data is a DataTypeGroup creates a zip with all files for all data types
         """
-        self.copy_dt_to_export_folder(data, data_export_folder)
+        data_export_folder = self.copy_dt_to_export_folder(data, export_folder)
```
popaula937, line 64: as discussed, the multiple calls to storage_interface methods could have been merged into a single export_datatypes() method. This results in the possibility to exclude from the storage API some methods like write_zip_folder.
I am pushing these changes soon.

tvb_storage/tvb/storage/storage_interface.py
```diff
@@ -395,17 +394,62 @@ def export_project(self, project, folders_to_exclude, export_folder, linked_path
...
+        data_export_folder = self.__build_data_export_folder(data, data_export_folder)
+        for dt_path in dt_path_list:
+            file_destination = os.path.join(data_export_folder, os.path.basename(dt_path))
+            if not os.path.exists(file_destination):
+                self.copy_file(dt_path, file_destination)
+            self.get_storage_manager(file_destination).remove_metadata('parent_burst', check_existence=True)
+
+        return data_export_folder
+
+    def copy_dt_to_export_folder(self, data, data_path, export_folder):
```
popaula937, line 447: I guess these last 2 methods could be merged into one as well.

tvb_storage/tvb/storage/storage_interface.py
```diff
@@ -156,31 +157,17 @@ def compute_recursive_h5_disk_usage(start_path='.'):
...
-        for data_type in all_datatypes:
-            operation_folder = self.get_project_folder(project_name, str(data_type.fk_from_operation))
-            operation_folders.append(operation_folder)
-        self.tvb_zip = TvbZip(zip_full_path, "w")
-        self.tvb_zip.write_zip_folders(operation_folders, exclude)
-        self.tvb_zip.close()
-
     @staticmethod
     def zip_folder(result_name, folder_root):
         FilesHelper.zip_folder(result_name, folder_root)
```
popaula937, line 173: This is called but does not exist on FilesHelper

popaula937 (commented):
Please also review the methods to remove files like: remove_operation_data, remove_files, remove_datatype_file, remove_image_metadata. Why do we need both remove_project and remove_project_structure methods?

tvb_storage/tvb/storage/storage_interface.py
```diff
@@ -388,24 +373,107 @@ def build_data_export_folder(self, data, export_folder):
...
+        if len(dt_path_list) == 1:
+            return file_destination
+
+        export_data_zip_path = os.path.join(os.path.dirname(export_folder), download_file_name)
+        self.write_zip_folder(export_data_zip_path, export_folder)
+        return export_data_zip_path
+
+    def export_datatypes_structure(self, all_datatypes, data, download_file_name, project_name):
+        """
+        This method is used to export a list of datatypes as a ZIP file. It is only used during normal tvb exporting.
```
popaula937, line 461: This method is used to export a list of datatypes as a ZIP file, while preserving the folder structure (eg: operation folders). It is only used during normal tvb exporting for datatype groups.

VinczeRobert:
> Please also review the methods to remove files like: remove_operation_data, remove_files, remove_datatype_file, remove_image_metadata. Why do we need both remove_project and remove_project_structure methods?

The remove_project_structure method is almost useless. It is used inside of the remove_project method and in a lot of tests. It is supossed to delete the storage part of a project, but:
1) In the tests by the time it is called in the teardown methods, everything is already deleted.
2) When it is called from remove_project the only things that it still has is the folder with the images and the xml file, so those could be deleted separately.

The remove_operation_data method is supposed to delete the an operation folder with all its contents. However, when this method is called (either through deleting a project or an operation) only the view model h5 file is still there, the datatype h5 file is deleted before.

The remove_files method is used for example to delete the files that make up the connectivity (both zip and csv) from the operation folder (like areas, centres etc.). Looks like it is used in operation_service too but I don't know what to do from the UI to get it called from there.

The remove_datatype_file is redundant and should be deleted. It justs delete a h5 file and used only when removing a datatype. If we don't call it, then the remove_operation_data which is called later will be remove both h5 files anyway.

The remove_image_metadata deletes the .xml file associated with each image. The image itself is deleted before calling this method. I suggest deleting both in this method.

I will make a task for refactoring this methods and I will copy this comment there.

PR #447, opened 2021-07-16
TVB-2929: Added Epileptor Model Plot

Ishan2601:
Added Epileptor Model Plot with widgets.
Usage:
`%matplotlib widget`
`from tvb.simulator.plot.epileptor_plotter import EpileptorModelPlot`
`ep = EpileptorModelPlot(conn, coupling, integrator, monitors, sim_length)`
`ep.show()`


popaula937:
Could you please update the notebook to use this code, as well?

popaula937 (commented):
The plotter class looks good, but I think the notebook needs some changes. Now, by default, you display the same thing in the plotter for each cell.
Could you, for each cell, select by default the burster class mentioned in the corresponding description and plot that?

PR #450, opened 2021-07-21
TVB-2930 + TVB-2931 : Solved project deletion issues

VinczeRobert:
I solved both problems because they were related.

Project deletion with links until now only worked if we linked a datatype that is not linked to some other datatypes. If we had, for example, a region mapping then it didn't copy the connectivity and the surface to the new project, thus we couldn't view the region mapping.

However, the bigger problem was that copying linked datatype groups didn't work at all. I managed to solve this as well. Now it doesn't matter if we link the datatype group for TSIndex or DMIndex, in both cases both of them will be copied to the new project.

If we have a datatype that is linked to more projects then it will be moved to the first project returned by the database and it will have a link pointing to the other projects.

I have tested several cases both in Sqlite and Postgres and didn't find any issues.
Maybe the project deletion process is too complicated and we talked with Paula about reviewing generally the remove methods. I intend to do that soon but for now, I think this is okay to be merged.

### DIFF
--- a/framework_tvb/tvb/core/services/project_service.py
+++ b/framework_tvb/tvb/core/services/project_service.py
@@ -681,7 +681,8 @@ class ProjectService:
 
         # Datatype Groups were already handled when the first DatatypeMeasureIndex has been found
         if dao.is_datatype_group(datatype_gid):
-            return
+            is_datatype_group = True
+            datatype_group = datatype
         # Found the first DatatypeMeasureIndex from a group
         elif datatype.fk_datatype_group is not None:
             is_datatype_group = True
@@ -704,8 +705,12 @@ class ProjectService:
                 links.extend(dao.get_links_for_datatype(dt_group_for_metric_op_group.id))
 
                 if len(links) > 0:
-                    # We want to get the links for the TSIndex directly, instead of DatatypeMeasureIndex
-                    ts = h5.load_entity_by_gid(datatype.fk_source_gid)
+                    # We want to get the links for the first TSIndex directly
+                    # This code works for all cases
+                    dt_group = dao.get_datatypegroup_by_op_group_id(burst.fk_operation_group)
+                    datatypes = dao.get_datatype_in_group(dt_group.id)
+                    ts = datatypes[0]
+
                     new_dt_links = self._add_links_for_linked_datatypes(ts, links[0].fk_to_project, links[0].id,
                                                                         existing_dt_links)
 
--- a/framework_tvb/tvb/core/services/project_service.py
+++ b/framework_tvb/tvb/core/services/project_service.py
@@ -301,20 +301,12 @@ class ProjectService:
 
             # Delete datatypes one by one in the reversed order of their creation date
             project_datatypes.sort(key=lambda dt: dt.create_date, reverse=True)
+            links = []
             for one_data in project_datatypes:
-                self.remove_datatype(project_id, one_data.gid, True)
-
-            # Delete any links for datatypes that point to the project they are in because they are redundant.
-            # This can happen when a datatype is linked with more datatypes (e.g. a Connectivity linked to
-            # a Region Mapping and a Time Series
-            projects = dao.get_all_projects()
-            for project in projects:
-                datatypes = dao.get_datatypes_in_project(project.id)
-                for dt in datatypes:
-                    links = dao.get_links_for_datatype(dt.id)
-                    for link in links:
-                        if link.fk_to_project == project.id:
-                            dao.remove_entity(Links, link.id)
+                new_links = self.remove_datatype(project_id, one_data.gid, True, links)
+                if new_links is not None:
+                    # Keep track of links so we don't create the same link more than once
+                    links.extend(new_links)
 
             self.storage_interface.remove_project(project2delete)
             dao.delete_project(project_id)
@@ -532,7 +524,7 @@ class ProjectService:
             return meta_atts, states, None
 
     @staticmethod
-    def _add_links_for_linked_datatypes(datatype, fk_to_project, link_to_delete):
+    def _add_links_for_linked_datatypes(datatype, fk_to_project, link_to_delete, existing_dt_links):
         # If we found a datatype that has links, we need to link those as well to the linked project
         # so they can be also copied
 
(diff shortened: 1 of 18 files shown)


framework_tvb/tvb/core/services/project_service.py
```diff
@@ -295,16 +298,23 @@ def remove_project(self, project_id):
...
-                dao.remove_entity(Links, one_link.id)
-            project_bursts = dao.get_bursts_for_project(project_id)
-            for burst in project_bursts:
-                dao.remove_entity(burst.__class__, burst.id)
+            # Delete any links for datatypes that point to the project they are in because they are redundant.
+            # This can happen when a datatype is linked with more datatypes (e.g. a Connectivity linked to
+            # a Region Mapping and a Time Series
+            projects = dao.get_all_projects()
+            for project in projects:
+                datatypes = dao.get_datatypes_in_project(project.id)
```
liadomide, line 312: Waw. These queries are extremely costly. 
Why not to keep (maybe fix) the original intended query ?

framework_tvb/tvb/core/services/project_service.py
```diff
@@ -36,10 +36,13 @@
 """
 
 import formencode
+import os
 
+from tvb.adapters.datatypes.db.time_series import TimeSeriesIndex
```
liadomide, line 41: This dependency is not good here (from 'core' towards 'adapters')

framework_tvb/tvb/core/services/project_service.py
VinczeRobert: Because that query was deleting the links from the project to be deleted and after I refactored the code by the time it would reach that line, those links are already deleted. My queries are intending to remove redundant links: links which point to the same project that the datatype is in. I will think of doing it without these 3 fors...

framework_tvb/tvb/core/services/project_service.py
VinczeRobert: True, I can check some attribute here that time series have and datatype measures don't, like 'data_ndim'.

VinczeRobert:
I removed those three fors now by keeping in a list the additional links that need to be created and this way we don't add the same link twice. 

Also, I don't understand why when we link a datatype group we are making a links towards each data of the group. I changed it to add the link only for the group and I don't see any errors, but if there is a reason, please enligthen me.

The only problem with this solution is that there are two cases when redundant links (links that point towards the same project that the datatype is in) will remain in the database, but they will not cause any errors in the functionality:
1. if the user links both the ts group and the datatype measure group, the link for the ts group will remain
2. e.g. if the user links a region mapping and the connectivity that was used for the region mapping: a link towards the conn will remain.
These 2 problems can be solved by additional conditions but that would just further complicate the solution. Truth is, its pretty hard to copy all kinds of linked datatypes when a project is deleted.

framework_tvb/tvb/core/services/project_service.py
```diff
@@ -521,65 +523,123 @@ def get_datatype_details(datatype_gid):
...
+        h5_file = h5.h5_file_for_index(datatype)
+        h5.gather_all_references_by_index(h5_file, linked_datatype_paths)
+
+        for h5_path in linked_datatype_paths:
+            if h5_path in existing_dt_links:
+                continue
+
+            dt = h5.load(h5_path)
+            dt_index = h5.load_entity_by_gid(dt.gid)
+            new_link = Links(dt_index.id, fk_to_project)
```
popaula937, line 541: Could we avoid loading the `dt` and `dt_index` here, and find the gid directly from the H5 file with the `H5File.get_metadata_param` method which is static?

framework_tvb/tvb/core/services/project_service.py
```diff
@@ -521,65 +523,123 @@ def get_datatype_details(datatype_gid):
...
-                    dao.remove_entity(Links, links[0].id)
+                op = dao.get_operation_by_id(datatype.fk_from_operation)
+                # Instead of deleting, we copy the datatype to the linked project
+                # We also clone the operation
+                new_operation = self.__copy_linked_datatype_before_delete(op, datatype, project,
+                                                                          links[0].fk_to_project)
+
+                # If there is a  datatype group and operation group and they were not moved yet to the linked project,
+                # then do it
+                if datatype.fk_datatype_group is not None:
```
popaula937, line 565: isn't this called for every dt in a group? Could we avoid that?

framework_tvb/tvb/core/services/project_service.py
```diff
@@ -599,7 +659,7 @@ def remove_operation(self, operation_id):
         else:
             self.logger.warning("Attempt to delete operation with id=%s which no longer exists." % operation_id)
 
-    def remove_datatype(self, project_id, datatype_gid, skip_validation=False):
+    def remove_datatype(self, project_id, datatype_gid, skip_validation=False, existing_dt_links=[]):
```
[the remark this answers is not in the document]
VinczeRobert: I saw this in other places this as well but this one can be easily changed to None so it is okay.

framework_tvb/tvb/core/services/project_service.py
```diff
@@ -610,11 +670,20 @@ def remove_datatype(self, project_id, datatype_gid, skip_validation=False):
             self.logger.warning("Attempt to delete DT[%s] which no longer exists." % datatype_gid)
             return
 
+        if datatype.parent_operation.fk_launched_in != int(project_id):
```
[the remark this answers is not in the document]
VinczeRobert: When a project is deleted which has a datatype group being linked to another project. When the first member of the group will be removed, all the related files (all TSs and Datatypemeasures) are moved to the new project thus the return statement.

framework_tvb/tvb/core/services/project_service.py
```diff
@@ -521,65 +523,123 @@ def get_datatype_details(datatype_gid):
             # We ignore exception here (it was logged above, and we want to return no details).
             return meta_atts, states, None
 
-    def _remove_project_node_files(self, project_id, gid, skip_validation=False):
+    @staticmethod
+    def _add_links_for_linked_datatypes(datatype, fk_to_project, link_to_delete, existing_dt_links):
```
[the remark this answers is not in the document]
VinczeRobert: Of course.

framework_tvb/tvb/core/services/project_service.py
VinczeRobert: But I need the id, not the gid.

framework_tvb/tvb/core/services/project_service.py
VinczeRobert: No it is not. The condition you asked before at line 673 makes the flow not reach this method for every datatype in the group.

PR #454, opened 2021-07-23
TVB-2941: Show message when older projects are imported

VinczeRobert:
…imported and delete related code.

### DIFF
--- a/framework_tvb/tvb/core/services/import_service.py
+++ b/framework_tvb/tvb/core/services/import_service.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 #
-# TheVirtualBrain-Framework Package. This package holds all Data Management, and 
+# TheVirtualBrain-Framework Package. This package holds all Data Management, and
 # Web-UI helpful to run brain-simulations. To use it, you also need do download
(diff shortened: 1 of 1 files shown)


VinczeRobert:
If there is going to be a demand to support importing older projects, here's an ideea I have briefly discussed with Lia and that could work: launch a separate tvb import operation for each datatype of the old project.

framework_tvb/tvb/core/services/import_service.py
```diff
@@ -180,17 +175,24 @@ def import_project_structure(self, uploaded, user_id):
...
-                project_roots.append(root)
+                temp_project_path = root
+                break
 
-        for temp_project_path in project_roots:
+        if temp_project_path is not None:
             update_manager = ProjectUpdateManager(temp_project_path)
+
+            if update_manager.checked_version < 3:
+                raise ImportException('Importing projects with older versions is unsupported! Please import the '
```
popaula937, line 192: Inform which is the minimum version supported: "Importing projects with versions older than 3 is not supported in TVB 2...launch the current version of TVB in order to upgrade this project!" ?

popaula937 (approved):

PR #456, opened 2021-08-10
TVB-2889: Suppress numpy warnings

VinczeRobert:
The warnings specified in the task come from numpy operations. Most of these warnings come from the test_simulator_region, but I have suppressed all of them. Numpy has 4 types of warnings: overflow, underflow, invalid, divide. We can choose to suppress all of them or only some types as written here: https://numpy.org/doc/stable/reference/generated/numpy.errstate.html

With Lia we thought that there is a problem with the numba logging in the library_logger_test.conf file. I tried to run only one test by trying to edit it in several ways but it did not make a difference. Maybe the numba logging doesn't do anything and is irrelevant. And one important thing: when we run only one test, the profile is set AFTER the logging configurations, so that means that the library_logger.conf file will be used and not the testing one.

### DIFF
--- a/scientific_library/tvb/tests/library/datatypes/sensors_test.py
+++ b/scientific_library/tvb/tests/library/datatypes/sensors_test.py
@@ -69,7 +69,9 @@ class TestSensors(BaseTestCase):
         dummy_surf = SkinAir()
         dummy_surf.vertices = numpy.array(list(range(30))).reshape(10, 3).astype('f')
         dummy_surf.triangles = numpy.array(list(range(9))).reshape(3, 3)
-        dummy_surf.configure()
+
+        with numpy.errstate(all='ignore'):
+            dummy_surf.configure()
         try:
             dt.sensors_to_surface(dummy_surf)
             self.fail("Should have failed for this simple surface!")
--- a/scientific_library/tvb/tests/library/datatypes/sensors_test.py
+++ b/scientific_library/tvb/tests/library/datatypes/sensors_test.py
@@ -69,12 +69,13 @@ class TestSensors(BaseTestCase):
         dummy_surf = SkinAir()
         dummy_surf.vertices = numpy.array(list(range(30))).reshape(10, 3).astype('f')
         dummy_surf.triangles = numpy.array(list(range(9))).reshape(3, 3)
+        dummy_surf.triangle_normals = numpy.array(list(range(9))).reshape(3, 3)
+        dummy_surf.vertex_normals = numpy.array(list(range(30))).reshape(10, 3).astype('f')
+        dummy_surf.configure()
 
-        with numpy.errstate(all='ignore'):
-            dummy_surf.configure()
         try:
             dt.sensors_to_surface(dummy_surf)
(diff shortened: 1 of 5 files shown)


scientific_library/tvb/tests/library/datatypes/sensors_test.py
```diff
@@ -69,7 +69,9 @@ def test_sensors(self):
         dummy_surf = SkinAir()
         dummy_surf.vertices = numpy.array(list(range(30))).reshape(10, 3).astype('f')
         dummy_surf.triangles = numpy.array(list(range(9))).reshape(3, 3)
-        dummy_surf.configure()
+
+        with numpy.errstate(all='ignore'):
+            dummy_surf.configure()
```
liadomide, line 74: Can't we avoid the errors in `dummy_surf.configure()`, if we define the vertices/triangles in a smarter way ?

scientific_library/tvb/tests/library/datatypes/surfaces_test.py
```diff
@@ -117,7 +120,9 @@ def test_cortical_topology_isolated_vertex(self):
         dt = surfaces.Surface()
         dt.vertices = numpy.array([[0, 0, 0], [1, 0, 0], [0, 1, 0], [0, 0, 1], [0, 0, 2]]).astype(numpy.float64)
         dt.triangles = numpy.array([[0, 2, 1], [0, 1, 3], [0, 3, 2], [1, 2, 3]])
-        dt.configure()
+
+        with numpy.errstate(all='ignore'):
+            dt.configure()
```
liadomide, line 125: Here, also, I see two dummy surfaces define above. Would there be a simple way to fix them (e.g. make sure it is a cube) ?

liadomide (approved):
I think this is a decent solution. 
Only one small suggestion from me for the dummy surface, in case it is reasonable to find a different solution there.

scientific_library/tvb/tests/library/datatypes/sensors_test.py
VinczeRobert: How? I tried a few different options for the array but line 350 will give the warnings because tri_norm will have 0. values.

liadomide: The fix depends on the actual operation giving error. 
As you say that `tri_norm` are 0, probably they should not be. It should be easy to fill triangle normals with something meaningful (a vector perpendicular on the triangle pointing towards the outside of the object).

popaula937: Avoided the computations generating warnings by defining dummy normals for triangles and vertices. @VinczeRobert please double-check this is reducing the numpy warnings.

popaula937 (approved):

scientific_library/tvb/tests/library/datatypes/sensors_test.py
VinczeRobert: I checked it again. There are no more numpy warnings when running the library tests.

PR #457, opened 2021-08-20
Update psi plot

JulieCB:
Hi,
Following a discussion on the TVB Google group, I have found some errors in the code for plotting the power spectra:

- the sampling period of the input data was not converted in seconds
- some floats in slicing were not converted in integers

Best,
Julie


liadomide (approved):

scientific_library/tvb/simulator/plot/power_spectra_interactive.py
```diff
@@ -156,7 +157,7 @@ def configure(self):
         LOG.debug("time_series shape: %s" % str(self.time_series.data.shape))
         #TODO: if isinstance(self.time_series, TimeSeriesSurface) and self.first_n == -1: #LOG.error, return.
         self.data = self.time_series.data[:, :, :self.first_n, :]
-        self.period = self.time_series.sample_period
+        self.period = self.time_series.sample_period/1000  #JC: convert sampling period of time_series.data in second
```
liadomide, line 160: Is this generic enough ?
Could we maybe wrap this line with a more detailed condition in which we consider `time_series.sample_period_unit` property?
Or use `1/self.time_series.sample_rate` ?

JulieCB: Hi Lia,
Yes,  we could directly use `self.period = 1/self.time_series.sample_rate`.
I was also thinking that we could specify the unit of the converted data: `self.period_unit='s'`

scientific_library/tvb/simulator/plot/power_spectra_interactive.py
```diff
@@ -35,9 +35,9 @@
 ::
 
     #Load the demo data
-    import numpy
-    data = numpy.load("demos/demo_data_region_16s_2048Hz.npy")
```
JulieCB, line 39: does this file still available?

liadomide: i think such files can be produced by some notebooks under "demo" folder

popaula937: Hi Julie, regarding that particular file, it is generated by a function called `generate_region_demo_data()` that is present in the notebook: tvb-root/tvb_documentation/demos/sandbox/pca_analyse_view_region.ipynb

JulieCB: Hi Paula,
Thank you!

Should we not make this dataset available in the sandbox folder for the user, as it appears in several python files for usage case? Same for the file generate_region_demo_data.py (line 487) for the `LOG.error`?

scientific_library/tvb/simulator/plot/power_spectra_interactive.py
```diff
@@ -156,7 +158,11 @@ def configure(self):
         LOG.debug("time_series shape: %s" % str(self.time_series.data.shape))
         #TODO: if isinstance(self.time_series, TimeSeriesSurface) and self.first_n == -1: #LOG.error, return.
         self.data = self.time_series.data[:, :, :self.first_n, :]
-        self.period = self.time_series.sample_period
+        if self.time_series.sample_period_unit == "s":
+            self.period = self.time_series.sample_period
+        else 
+            self.period = 1 / self.time_series.sample_rate  #JC: convert sampling period of time_series.data in second
```
popaula937, line 164: I think the value of the `sample_rate` property is always converted to seconds inside the `TimeSeries` class, so we could avoid the if-else and keep only line 164. But please correct me if I am wrong.

JulieCB: yes, you are right!
I didn't see the last line code for the sample rate in seconds inside the `TimeSeries`

scientific_library/tvb/simulator/plot/power_spectra_interactive.py
popaula937: I suppose the generate_region_demo_data.py used to generate the same dataset (demo_data_region_16s_2048Hz.npy). I would suggest keeping only the function and run it automatically when the dataset is not found. 
Would it be better to make the `generate_region_demo_data` function available through the `tvb.simulator.lab `interface? @liadomide

popaula937 (approved):

scientific_library/tvb/simulator/plot/power_spectra_interactive.py
popaula937: @JulieCB  I think this PR can be merged, and I can take care of making the dataset or the function available afterwards if that is ok from your side.

PR #458, opened 2021-08-20
TVB-2963 pulse train fix

i-Zaak:
The evaluation of PulseTrain equation returned wrong results if the provided `var` vector doesn't start at 0. 

e.g.

```python
pt = equations.PulseTrain( parameters={ 'onset': 40, 'T': 15.0 , 'tau': 10.0, 'amp': 1.} )
t = np.linspace(5,100,1000)
plt.plot(t, pt.evaluate(t))
plt.axvline(x=40, c='r', ls=':')
plt.axvline(x=50, c='r', ls=':')
```

image

### DIFF
--- a/scientific_library/tvb/datatypes/equations.py
+++ b/scientific_library/tvb/datatypes/equations.py
@@ -301,20 +301,14 @@ class PulseTrain(TemporalApplicableEquation):
 
     * :math:`\\tau` :  pulse width or pulse duration
     * :math:`T`     :  pulse repetition period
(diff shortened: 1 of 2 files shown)


i-Zaak:
Btw, do we have a simple way of rendering the docstrings with math?

liadomide (approved):

liadomide:
For rendering docstrings with math, Maybe you can test some stuff in editors like: https://www.tuhh.de/MathJax/test/sample-dynamic.html, or just try TVB web GUI.

image

PR #461, opened 2021-08-24
TVB-2755 + TVB-2879: Add EnumAttr to TVB and eliminate the usage of dicts in SelectField

VinczeRobert:
The two tasks are strongly related so I solved them at the same time. However, pretty much 90% of the changes are related to EnumAttr.

The task(s) are almost done. There should be only a few tests still failing, I have to fix the contrib tests by adding a separate ModelsEnum there so models that don't appear in the UI can be used there and an extensive UI testing will be necessary.

There are a lot of changes so please ask me if I need to explain why I did certain changes or if you have suggestions.


framework_tvb/tvb/adapters/simulator/simulator_fragments.py
```diff
@@ -138,41 +138,36 @@ def prepare_stimulus_fragment(simulator, rendering_rules, is_surface_simulation,
...
 
     def fill_from_trait(self, trait):
         # type: (SimulatorAdapterModel) -> None
-        self.model.data = trait.model.__class__
+        self.model.data = type(trait.model)
 
     def fill_trait(self, datatype):
-        if type(datatype.model) != self.model.value:
-            datatype.model = self.model.value()
+        if type(datatype.model) != type(self.model.value.instance):
```
popaula937, line 152: Why not avoid creating an instance and use .value instead? I think I saw this contruction in more places.

popaula937 (commented):
There is a regression in the phase plane regarding integrators containing '/' in their names (task https://req.thevirtualbrain.org/browse/TVB-2876) that you solved in the past. Please take a look at that.

framework_tvb/tvb/adapters/simulator/simulator_fragments.py
VinczeRobert: Yeah, you are right, but I think this is the only place where I didn't do like you say.

popaula937 (approved):
Looks good, but few unused imports left and we could live with only 2 Enum types, as discussed.

PR #464, opened 2021-08-30
Fix phase plane noise slider config (TVB-3026)

JulieCB:
Hi,

Fixation of a bug for noise usage.

### DIFF
--- a/scientific_library/tvb/simulator/plot/phase_plane_interactive.py
+++ b/scientific_library/tvb/simulator/plot/phase_plane_interactive.py
@@ -139,7 +139,7 @@ class PhasePlaneInteractive(HasTraits):
         # figure
         self.ipp_fig = None
 
-        # p hase-plane
+        # phase-plane
         self.pp_ax = None
         self.X = None
         self.Y = None
@@ -401,8 +401,7 @@ class PhasePlaneInteractive(HasTraits):
         pos_shp = [0.825, 0.1, 0.125, 0.025]
         sax = self.ipp_fig.add_axes(pos_shp, facecolor=AXCOLOUR)
 
-        self.noise_slider = widgets.Slider(sax, "Log Noise", -9.0, 1.0,
-                                           valinit=self.integrator.noise.nsig)
+        self.noise_slider = widgets.Slider(sax, "Log Noise", numpy.array([-9.0,]), numpy.array([1.0,]), valinit=self.integrator.noise.nsig)
         self.noise_slider.on_changed(self.update_noise)
 
     def add_reset_param_button(self):


JulieCB:
Hi Paula,

I have a bug when using stochastic scheme (Euler or Heun). I am using a jupyter notebook to visualize the results.
The error I get it's this one:
`AttributeError: 'float' object has no attribute 'rint'` indicating line 404 `self.noise_slider = widgets.Slider(sax, "Log Noise", -9.0, 1.0, valinit=self.integrator.noise.nsig)`

Hope that helps you!

I also have a short question: why using "Log Noise"?

maedoc:
Log noise is appropriate because the noise scaling cannot be negative, and the interesting tuning is on a logarithmic scale which allows changing order of magnitude more easily than linear.

VinczeRobert:
I tried to reproduce the error but I wasn't able to. It seems like this add_noise_slider method is called from the show method of the PhasePlanInteractive class. I ran the cells from tutorial_s3_exploring_a_model notebook cause I saw that the show method is called there, but everything ran smoothly.

Could you maybe copy-paste here that code that you ran when encountering the problem?

JulieCB:
The problem appears when using small noise values below 1.


---
Best regards,

Julie Courtiol





Le jeu. 2 sept. 2021 à 15:23, Vincze Robert ***@***.***> a
écrit :

maedoc:
@JulieCB reply by email doesn't post the attachment, please open GitHub and post it

JulieCB:
Using tutorial_s3_exploring_a_model.ipynb:

in the Stochastic Integration section, if we modify the noise strength and set it to a value below 1 like here: 0.01

nsigma = np.array([0.01,])           
hiss = noise.Additive(nsig=nsigma)   
heunstoint = integrators.HeunStochastic(dt=dt, noise=hiss)  

you should get an error in the function add_noise_slider: 'float' object has no attribute 'rint'

maedoc:
Ok, this allows the UI to show up, but the noise slider doesn't work, and the same error message shows up in the console when trying to change the noise level. 

Additionally, the PPI doesn't update trajectories when moving the slider.

maedoc:
@JulieCB any updates?

liadomide:
I think the better fix here is:
```
        self.noise_slider = widgets.Slider(sax, "Log Noise", -9.0, 1.0,
                                           valinit=self.integrator.noise.nsig[0])
```
because params `valmin, valmax, valinit` all expect single float values, and the previous `valinit=self.integrator.noise.nsig` seems to be the culprit.

Would you agree that i close this PR, make this change in a tvb-root branch, then we merge ?

PR #465, opened 2021-09-01
Fix small error in simulator.py

VinczeRobert:
I found this error while doing UI testing. There is no n_intvar parameter, there is only nintvar and that's on the model classes, not the simulator.

Also, there is another small issue I found and discussed with Paula: the equation for local connectivity is unrequired in the code when in the UI it definetely should be required. Are there some examples in the library where you can use LCs without an equation? If not, I shall fix that as well in this PR.


popaula937:
Could you do this fix on a branch with code only from master, so we can easily merge it?

popaula937 (approved):

liadomide:
Regarding the LC equation: it s correct to have it required in TVB web GUI on page "LC creator", as we can only create now a LS based on an equation. But in the library, one could simply import a LS sparse matrix, of the right shape, without giving us a tvb equation. I think this is the case with the default LC, which does not have an equation.

PR #467, opened 2021-09-06
TVB-2964: Move generate_region_demo_data from notebook to library and adapt usages

VinczeRobert:


scientific_library/tvb/simulator/plot/power_spectra_interactive.py
```diff
@@ -484,11 +470,13 @@ def plot_spectra(self):
...
     LOG.info("Testing %s module..." % __file__)
+    file_path = os.path.join(os.getcwd(), "demo_data_region_16s_2048Hz.npy")
     try:
-        data = numpy.load("../demos/demo_data_region_16s_2048Hz.npy")
+        data = numpy.load(file_path)
     except IOError:
-        LOG.error("Can't load demo data. Run demos/generate_region_demo_data.py")
-        raise
+        LOG.error("Can't load demo data. It will be created now running the generate_region_demo_data() function.")
+        generate_region_demo_data()
```
popaula937, line 478: Give the `file_path` as an argument here for consistency

liadomide (approved):
The location under *plot.utils.py seems a bit odd at first, but after seeing its current usages is a decent position, and we should consider changing it only if usages will change

PR #468, opened 2021-09-10
Mergeable Unified RateML

maedoc:
This is the unified-rateml branch with some fixes for merging


maedoc:
@DeLaVlag no conflicts now, rateml tests pass locally and we'll just wait for the regular library tests to pass, then merge

maedoc:
Maybe we should mark the whole set of rateml tests as conditional on pycuda being available?  It's a little tricky to track down all the dependencies.  wdyt @DeLaVlag ?

DeLaVlag:
I agree. It doesnt make sense to run them if you don't have pycuda.

PR #469, opened 2021-09-13
TVB-2968: Add Cortex.Coupling_Strength as a range parameter

VinczeRobert:
Just adding FormWithRanges as a superclass to SurfaceRMFragment was not enough.

We have only one surface range parameter but I added in the same way as the other parameter for consistency.

Also, I removed the constructor arguments for SimulatorRangeParameters because they are always None.

### DIFF
--- a/framework_tvb/tvb/interfaces/web/controllers/simulator/simulator_controller.py
+++ b/framework_tvb/tvb/interfaces/web/controllers/simulator/simulator_controller.py
@@ -223,6 +223,7 @@ class SimulatorController(BurstBaseController):
             form = SimulatorSurfaceFragment()
             form.fill_from_post(data)
             self.simulator_service.reset_at_surface_change(is_simulation_copy, form, session_stored_simulator)
+            self.range_parameters.surface_parameters = None
             form.fill_trait(session_stored_simulator)
 
             if session_stored_simulator.surface:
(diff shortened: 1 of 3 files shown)


framework_tvb/tvb/interfaces/web/controllers/simulator/simulator_controller.py
```diff
@@ -238,13 +237,14 @@ def set_surface(self, **data):
...
         session_stored_simulator, is_simulation_copy, is_simulation_load, _ = self.context.get_common_params()
+        rm_fragment = SimulatorRMFragment()
 
         if cherrypy.request.method == POST_REQUEST:
             self.context.add_last_loaded_form_url_to_session(SimulatorWizzardURLs.SET_STIMULUS_URL)
-            rm_fragment = SimulatorRMFragment()
             rm_fragment.fill_from_post(data)
             rm_fragment.fill_trait(session_stored_simulator.surface)
 
+        self.range_parameters.surface_parameters = rm_fragment.get_range_parameters()
```
popaula937, line 247: Isn't the coupling_strength added to the PSE params list even if no surface has been selected for the simulation?

VinczeRobert: It won't be added, because this setting is in the set_cortex method and that method is not called when a surface is not provided.

popaula937: After talking with Robert, we concluded that we need to handle more carefully the PSE params which come from optional fragments, otherwise they will be left in the PSE params list even if no surface/noise is chosen. He will add a fix soon.

VinczeRobert: It should be okay now.

popaula937 (approved):

PR #471, opened 2021-09-16
HBP-146: Handle export feature when security is active

VinczeRobert:
I managed to make it work for the simplest case: when a single datatype is exported in TVB format.
I wrote 2 scripts:
- one for generating a pair of public and private keys
- one for decrypting data

I will turn these into two Jupyter notebooks and the user will have to specify the paths to his data, just as it happens with encrypt_data.ipynb. 

Also, I moved encryption related methods from ABCUploader to StorageInterface because now they have to be used in two places (not just the uploader) and also because it seems more fitting to be there. 

I don't know how to make the HTML template look as it should: with only one Choose file button formatted in a consistent way with similar buttons in TVB.


framework_tvb/tvb/adapters/exporters/tvb_export.py
```diff
@@ -64,11 +64,15 @@ def export(self, data, project):
             _, op_file_dict = self.prepare_datatypes_for_export(data)
 
             # Create ZIP archive
-            zip_file = self.storage_interface.export_datatypes_structure(op_file_dict, data, download_file_name)
+            zip_file = self.storage_interface.export_datatypes_structure(op_file_dict, data, download_file_name,
+                                                                         public_key_path, password)
+            if password is not None:
+                download_file_name = self.storage_interface.add_encrypted_suffix(download_file_name)
```
popaula937, line 70: Could this suffix be added directly inside `export_datatypes_structure`?

framework_tvb/tvb/core/adapters/abcuploader.py
```diff
@@ -100,79 +92,19 @@ def _prelaunch(self, operation, view_model, available_disk_space=0):
         if view_model.encrypted_aes_key is not None:
             trait_upload_field_names = list(self.get_form_class().get_upload_information().keys())
 
+            if TvbProfile.current.UPLOAD_KEY_PATH is None or not os.path.exists(TvbProfile.current.UPLOAD_KEY_PATH):
+                raise LaunchException("We can not process Encrypted files at this moment, "
+                                      "due to missing PK for decryption! Please contact the administrator!")
```
popaula937, line 97: Maybe leave just: "TVB can not process Encrypted files at this moment! Please contact the administrator!"

framework_tvb/tvb/interfaces/web/controllers/project/project_controller.py
```diff
@@ -677,14 +677,15 @@ def updatemetadata(self, **data):
...
+    def downloaddatatype(self, data_gid, export_module, **data):
         """ Export the data to a default path of TVB_STORAGE/PROJECTS/project_name """
         current_prj = common.get_current_project()
         # Load data by GID
         entity = load_entity_by_gid(data_gid)
         # Do real export
         export_mng = ExportManager()
-        file_name, file_path, delete_file = export_mng.export_data(entity, export_module, current_prj)
+        file_name, file_path, delete_file = export_mng.export_data(entity, export_module, current_prj,
+                                                                   data['exporter_key'])
```
popaula937, line 688: Use `data.get('exporter_key')` to avoid inexistent key error and get None value instead

tvb_storage/tvb/storage/storage_interface.py
```diff
@@ -341,7 +346,66 @@ def move_datatype_with_sync(self, to_project, to_project_path, new_op_id, full_p
...
+
+    @staticmethod
+    def encrypt_password(public_key, symmetric_key):
+        return ImportExportEncryptionHandler.encrypt_password(public_key, symmetric_key)
+
+    def save_encrypted_password(self, encrypted_password, path_to_encrypted_password):
+        return self.import_export_encryption_handler.save_encrypted_password(encrypted_password,
+                                                                             path_to_encrypted_password)
+
+    def encrypt_and_save_password(self, public_key_path, password, path_to_encrypted_password):
```
popaula937, line 394: Method used only within StorageInterface, could we avoid exposing it and use directly and use `import_export_encryption_handler` in the methods caling it?

tvb_storage/tvb/storage/storage_interface.py
```diff
@@ -341,7 +346,66 @@ def move_datatype_with_sync(self, to_project, to_project_path, new_op_id, full_p
...
+        self.import_export_encryption_handler.generate_public_private_key_pair(key_path)
+
+    def add_encrypted_suffix(self, name):
+        return self.import_export_encryption_handler.add_encrypted_suffix(name)
+
+    def get_path_to_encrypt(self, input_path):
+        return self.import_export_encryption_handler.get_path_to_encrypt(input_path)
+
+    @staticmethod
+    def load_public_key(public_key_path):
```
popaula937, line 383: Method used only within StorageInterface, could we avoid exposing it?

tvb_storage/tvb/storage/storage_interface.py
```diff
@@ -341,7 +346,66 @@ def move_datatype_with_sync(self, to_project, to_project_path, new_op_id, full_p
...
+
+    def save_encrypted_password(self, encrypted_password, path_to_encrypted_password):
+        return self.import_export_encryption_handler.save_encrypted_password(encrypted_password,
+                                                                             path_to_encrypted_password)
+
+    def encrypt_and_save_password(self, public_key_path, password, path_to_encrypted_password):
+        self.import_export_encryption_handler.encrypt_and_save_password(public_key_path, password,
+                                                                        path_to_encrypted_password)
+
+    def __encrypt_data_at_export(self, file_path, password):
```
popaula937, line 398: same here

tvb_storage/tvb/storage/storage_interface.py
```diff
@@ -34,15 +34,20 @@
 
 .. moduleauthor:: Robert Vincze <robert.vincze@codemart.ro>
 """
+
 import os
+import shutil
 import uuid
+from cgi import FieldStorage
 from datetime import datetime
+from cherrypy._cpreqbody import Part
```
popaula937, line 43: Could we avoid the cherrypy dependency at storage level?

tvb_storage/tvb/storage/storage_interface.py
popaula937: Please review a bit all these new methods in StorageInterface to check which are not necessary to be exposed in the API

popaula937 (commented):
Encryption at export should be required when storage encryption is activated, so, in that case, we should not allow the export button to work without a password provided. And if storage encryption is not activated, we should not display the password upload field.

tvb_storage/tvb/storage/storage_interface.py
VinczeRobert: Yes, the method were cherrypy is used is called only once and we can keep that method outside of StorageInterface, for example in export_manager.

framework_tvb/tvb/adapters/simulator/simulator_fragments.py
```diff
@@ -264,6 +265,9 @@ def __init__(self, pse_param1, pse_param2):
         if pse_param2:
             self._add_pse_field(pse_param2, self.KEY_PARAM2)
 
+        self.max_pse_number = HiddenField(Int(default=TvbProfile.current.MAX_RANGE_NUMBER, required=False),
```
popaula937, line 268: Isn't this part of another PR?

VinczeRobert: A rebase and forced push solved the issue.

popaula937 (approved):

PR #474, opened 2021-09-22
Urateml fixes

DeLaVlag:
Fixes/updates for RateML:
-Both XML for python and cuda are similar. Last difference in state variables is now similar. 
-Users can enter a range to initialize the states randomly.
-The produced driver file named after the model generated
-Only lower cases of inf and pi need to be used. Will be processed accordingly in XML2model.py.
-The replacement of the powers is simplified. 
-Pytest is updated accordingly. 
-Model and XML files updated accordingly.


maedoc:
👆 looks good 👆 

@DeLaVlag you want to merge or more stuff coming?

DeLaVlag:
Nice! These are some necessary fixes/upgrades which make rml life more easy, so lets merge this one. Will work later on more tests on tvb vs cuda but this can be a separate PR.

PR #482, opened 2021-10-06
TVB-2979: Fix infinite numbers problem in matrix viewer

VinczeRobert:


framework_tvb/tvb/adapters/visualizers/matrix_viewer.py
```diff
@@ -69,7 +69,13 @@ def compute_raw_matrix_params(matrix):
         """
         Serializes matrix data, shape and stride metadata to json
         """
+
+        # replace infinite values with the biggest float number available in numpy
+        float_max = numpy.finfo(numpy.float64).max
+        matrix[matrix == numpy.inf] = float_max
```
liadomide, line 75: Maybe also -Inf ashould be handled ?

VinczeRobert: Riiight, I forgot about -Inf.

PR #485, opened 2021-10-07
Model specific stimulus

dionperd:
I added a simple mechanism for having model specific time dependent stimuli, i.e., stimuli that will be computed and applied to model.dfun, at each time step, by code specific to each TVB model.

I find the application of stimuli in the integrator. scheme as not flexible enough.

I only added model.stimulus (default = None) and a method update_stimulus (originally empty) in model base.py

as well as a call to model.update_stimulus(step, integrator.dt) inside simulator._loop_update_stimulus() method, in order that the stimulus is updated at each time step, knowing the time. 

I SUGGEST WE CANCEL THIS AND WE TURN INSTEAD TO THIS ONE:
[https://github.com/the-virtual-brain/tvb-root/pull/489]

### DIFF
--- a/scientific_library/tvb/simulator/models/base.py
+++ b/scientific_library/tvb/simulator/models/base.py
@@ -157,7 +157,7 @@ class Model(HasTraits):
         dt = integrator.dt
         return self.initial(dt, shape, rng)
 
-    def _update_stimulus(self, step, dt):
+    def update_stimulus(self, step, dt):
         """
         When needed, and if stimulus is not None,
         this should be a method for computing a model specific, time-dependent stimulus.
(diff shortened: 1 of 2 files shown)


maedoc (changes requested):
This allows models to be non-autonomous if they implement the interface, while avoiding the need to update the `dfun` signature of all the models, and thread `step` and `dt` through the relevant section of the call graph.  I'd prefer to rename to something more general since it could be used for other purposes, time dependent forcing functions.  But stimulus is ok.

Please add an example though, preferably a simple notebook to the examples folder.

dionperd:
Initially I went for a solution where stimulus=0.0 is the last kwarg in model.dfun(), which would allow most of the current models to ignore it.
However, then, there was no way to know if the user wants the stimulus to be applied in the dfun or outside it, i.e., if a simulator.stimulus were defined, it would be applied both inside the dfun, and in the integrator schema....
In my opinion the best solution would be indeed for stimulus=0.0 to be a kwarg in model.dfun, but there should be a flag also indicating if it has to be applied inside dfun or outside it, in the integrator.schema.

So, let me know what is better. 

I can always add a simple example then.

PR #487, opened 2021-10-08
TVB-2977: Rendering Simulator params with MathJax.

VinczeRobert:


framework_tvb/tvb/adapters/simulator/model_forms.py
```diff
@@ -141,7 +141,9 @@ def __init__(self, are_voi_disabled=False):
 
     @staticmethod
     def get_params_configurable_in_phase_plane():
-        return ['tau', 'a', 'b', 'c', 'I', 'd', 'e', 'f', 'g', 'alpha', 'beta', 'gamma']
+        return {"tau": r":math:`\tau`", "a": ":math:`a`", "b": ":math:`b`", "c": ":math:`c`", "I": ":math:`I`",
```
popaula937, line 144: Are the values of these dictionaries used? I see the mathjax labels are being taken directly from attributes and not from these dictionaries.

popaula937:
The bug related to adding focal points in SetupSurfaceParameters still persists. That one should be fixed before merging this PR. Left some screenshots on Jira, please check them out.

framework_tvb/tvb/adapters/simulator/model_forms.py
```diff
@@ -141,7 +157,12 @@ def __init__(self, are_voi_disabled=False):
 
     @staticmethod
     def get_params_configurable_in_phase_plane():
-        return ['tau', 'a', 'b', 'c', 'I', 'd', 'e', 'f', 'g', 'alpha', 'beta', 'gamma']
+        return ["tau", "a", "b", "c", "I", "d", "e", "f", "g", "alpha", "beta", "gamma"]
+
+    @staticmethod
+    def get_mathjax_rendered_params():
+        return [r":math:`\tau`", ":math:`a`", ":math:`b`", ":math:`c`", ":math:`I`", ":math:`d`", ":math:`e`",
+                ":math:`f`", ":math:`g`", r":math:`\alpha`", r":math:`\beta`", r":math:`\gamma`"]
```
liadomide, line 165: These 2 methods (`get_params_configurable_in_phase_plane` and  `get_mathjax_rendered_params `) seem a bit difficult to maintain in synch. What if someone adds one param in one list and not in the other? Or what of the math text changes in the model, but not here?

Couldn't we better change what the first/existing method `get_param_configurable_in_phase_plane` returns?

For example, instead of returning a list of strings to return a list of `Attr` objects `[ModelsEnum.GENERIC_2D_OSCILLATOR.value.tau, ModelsEnum.GENERIC_2D_OSCILLATOR.value.a, ....]` ?
Other alternatives could also be imagined to solve part of the above issues...

VinczeRobert: Ok, I'll do that then.

VinczeRobert: In that case I will make get_params_of_configurable_in_phase_plane not a static method and will just return the needed parameters using self and separate them into lists after.

VinczeRobert: By the way: EpileptorRestingStateModelForm has a returned params in that method called gamma_rs, but the class doesn't have an attribute with this name. I am going to eliminate it from the method.

liadomide (approved):

PR #492, opened 2021-10-14
WIP HBP-147: Lay down the ground for TVB REST encryption.

VinczeRobert:
So far I made encryption/decryption work only for the fire_simulation script.
I hardcoded paths for the public key and private key. I think the best way would be to keep the private key the same way as we keep the one used for decrypting at uploading. The only difference is that here encryption would be done with the private key by TVB and the decryption with a downloadable public key by the user.

It seems like only the retrieve_datatype method does actually download H5 file(s). Should encrypted downloading be obligatory or optional?


popaula937:
The code flow looks good, I would suggest moving the mechanism from fire_simulation on a lower-level to hide the calls that decrypt data from the user. Also, I think decryption should happen only if the encryption is active in the server, we should speak about options to get that info. And regarding the passwords, maybe we could keep them as attributes on the TVBClient  class, to let the user easily set the paths towards them?

PR #505, opened 2021-11-02
WIP TVB-2615: Add Subcortical Surfaces in TVB Visualizations

VinczeRobert:


liadomide:
This attempt has too much JS logic, we will go towards a more back-end oriented solution, for memory and caching purposes, as surfaces can get really big, thus tedious to be handles in the pure JS code we have now in TVB.

PR #508, opened 2021-11-04
HBP-165 Common landing page

bvalean:


liadomide:
How do you propose to package this?
Do we need it in a docker image for OpenShift to run it ?
would it be too heavy if it is another entry point from the main tvb docker image (to avoid maintaining a separate one) ?

bvalean:
We will override the container's startup command in the openshift deployment.

PR #509, opened 2021-11-04
Added nonzero of weights to the determination of min delay for cosimulator

dionperd:
This is a minor modification to the cosimulator.py to determine the minimum delay of the connectome, and therefore the ideal synchronization time for cosimulation, by discarding all connections that have a zero weight, as well as a zero idelay, instead of the current version that considers only the idelays.

The reason is that a zero weight represents a missing connection more robustly than a zero idelay.

For instance, in some cases, and for my own reasons, I initialize zero delays to be equal to 0.1 (=integrator.dt; this might be stupid, but there are many users out there that might do something similar, given that they can still use the weights to determine a missing connection).

Still, the corresponding weights = 0.0, if this is a missing connection.

In this scenario, the current if statement:

        if self.synchronization_n_step > numpy.min(self.connectivity.idelays[numpy.nonzero(self.connectivity.idelays)]):
            raise ValueError('the synchronization time is too long')

in the cosimulator's configure_cosimulation() will give an error, because it will receive the 0.1 as the minimum delay and this is actually equal to the integrators.dt.

Instead: numpy.nonzero(self.connectivity.weights * self.connectivity.idelays) will be correct.

### DIFF
--- a/tvb_contrib/tvb/contrib/cosimulation/cosimulator.py
+++ b/tvb_contrib/tvb/contrib/cosimulation/cosimulator.py
@@ -97,7 +97,9 @@ class CoSimulator(Simulator):
         self.synchronization_n_step = iround(self.synchronization_time / self.integrator.dt)
         # Check if the synchronization time is smaller than the delay of the connectivity
         # the condition is probably not correct. It will change with usage.
-        if self.synchronization_n_step > numpy.min(self.connectivity.idelays[numpy.nonzero(self.connectivity.idelays)]):
+        if self.synchronization_n_step > numpy.min(self.connectivity.idelays[
+                                                       numpy.nonzero(self.connectivity.weights *
+                                                                     self.connectivity.idelays)]):
             raise ValueError('the synchronization time is too long')
 
         # Check if the couplings variables are in the cosimulation variables of interest
--- a/tvb_contrib/tvb/contrib/cosimulation/cosimulator.py
+++ b/tvb_contrib/tvb/contrib/cosimulation/cosimulator.py
@@ -95,11 +95,10 @@ class CoSimulator(Simulator):
         self.synchronization_time = numpy.maximum(self.synchronization_time, self.integrator.dt)
         # Compute the number of synchronization time steps:
         self.synchronization_n_step = iround(self.synchronization_time / self.integrator.dt)
-        # Check if the synchronization time is smaller than the delay of the connectivity
-        # the condition is probably not correct. It will change with usage.
-        if self.synchronization_n_step > numpy.min(self.connectivity.idelays[
-                                                       numpy.nonzero(self.connectivity.weights *
-                                                                     self.connectivity.idelays)]):
+        # Check if the synchronization time is smaller than the delay of the connectivity:
+        mask =  numpy.logical_and(self.connectivity.weights != 0,
+                                  self.connectivity.idelays != 0 )
+        if self.synchronization_n_step > self.connectivity.idelays[mask].min():
             raise ValueError('the synchronization time is too long')
 
         # Check if the couplings variables are in the cosimulation variables of interest
--- a/tvb_contrib/tvb/contrib/cosimulation/cosimulator.py
+++ b/tvb_contrib/tvb/contrib/cosimulation/cosimulator.py
@@ -97,7 +97,7 @@ class CoSimulator(Simulator):
         self.synchronization_n_step = iround(self.synchronization_time / self.integrator.dt)
         # Check if the synchronization time is smaller than the minimum delay of the connectivity:
         mask =  numpy.logical_and(self.connectivity.weights != 0,
-                                  self.connectivity.idelays != 0 )
+                                  self.connectivity.idelays != 0)
         if self.synchronization_n_step > self.connectivity.idelays[mask].min():
             raise ValueError('the synchronization time is too long')
 
--- a/tvb_contrib/tvb/contrib/cosimulation/cosimulator.py
+++ b/tvb_contrib/tvb/contrib/cosimulation/cosimulator.py
@@ -99,10 +99,10 @@ class CoSimulator(Simulator):
         if numpy.any(existing_connections.size):
             min_idelay = self.connectivity.idelays[existing_connections].min()
             if self.synchronization_n_step > min_idelay:
-                raise ValueError('The synchronization time is longer than '
+                raise ValueError('The synchronization time %g is longer than '
                                  'the minimum delay time %g '
                                  'of all existing connections (i.e., of nonzero weight)!'
-                                 % min_idelay * self.integrator.dt)
+                                 % (self.synchronization_time, min_idelay * self.integrator.dt))
 
     def _configure_cosimulation(self):
         """This method will
--- a/tvb_contrib/tvb/contrib/cosimulation/cosimulator.py
+++ b/tvb_contrib/tvb/contrib/cosimulation/cosimulator.py
@@ -95,7 +95,7 @@ class CoSimulator(Simulator):
         self.synchronization_time = numpy.maximum(self.synchronization_time, self.integrator.dt)
         # Compute the number of synchronization time steps:
         self.synchronization_n_step = iround(self.synchronization_time / self.integrator.dt)
-        # Check if the synchronization time is smaller than the delay of the connectivity:
+        # Check if the synchronization time is smaller than the minimum delay of the connectivity:
         mask =  numpy.logical_and(self.connectivity.weights != 0,
                                   self.connectivity.idelays != 0 )
         if self.synchronization_n_step > self.connectivity.idelays[mask].min():
--- a/tvb_contrib/tvb/contrib/cosimulation/cosimulator.py
+++ b/tvb_contrib/tvb/contrib/cosimulation/cosimulator.py
@@ -83,6 +83,27 @@ class CoSimulator(Simulator):
     _cosim_monitors_noncoupling_indices = []
     _cosim_monitors_coupling_indices = []
 
+    def _configure_synchronization_time(self):
+        """This method will set the synchronization time and number of steps,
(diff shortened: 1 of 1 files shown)


lionelkusch:
For my part, I prefer to have a modification of the function set_idelays in connectivity.py than to modify this condition.
The idelays should be zero when there are no connections.
What do you think?

dionperd:
> For my part, I prefer to have a modification of the function set_idelays in connectivity.py than to modify this condition. The idelays should be zero when there are no connections. What do you think?

I think that there is no reason for us to assume what the user wants for his (i)delays.

For us, both should be true:
a. the connection to exist -> weight !- 0.0
AND
b. idelay > 0.0

So, let's check for both and not worry about what the crazy user (e.g. me) does with his delays!

maedoc (changes requested):
I disagree with both of you in terms of the discussion so far: per original TVB semantics, delays are never non-zero except for nodes at identical points in space.  idelays are a derived integer array constant inversely proportional to conduction speed, and they should not be zero either with the same exception.  How you initialize with respect to the cosim is up to you, but please do not reinterpret these arrays otherwise. 

Concerning this change in particular, please write this instead:
```python
if min_nsteps > idelays[weights!=0].min():
```
(with variable names suitably adapted), multiplying weights and idelays is an abuse of notation.

lionelkusch:
I check the initial default connectome in TVB which should respect the TVB semantics.
For the connections without track-length have idelays equal zeros because the track-length is zeros. I assume that this respect the TVB semantics and the meaning of the idelays. If this is not standard, what is the standard values for the track-length when there doesn't exist.
Moreover your condition doesn't work because the delay between the node and itself is zero because it's identical to the same position in space. So the idelays[weights!=0].min() will always be zeros if there are recurrent connections, this doesn't allow some optimizations.

maedoc:
Of course you should ignore the diagonal, if those optimizations are important.  The default connectome doesn't have tract length information and uses distance only, so in some sense is an exception: recurrent connections in dMRI tractography are non-zero in both weight and tract length.

To reiterate, the logic behind the PR makes sense to me: if a weight is zero, you don't need to consider the corresponding delay since it doesn't contribute to the numerical solution of the equations.  I simply asked to implement this a little more readably, e.g.
```python
        mask = self.connectivity.weights != 0
        numpy.fill_diagonal(mask, False)  # ignore diagonal
        off_diag_nnz_idelays = self.connectivity.idelays[mask]
        if self.synchronization_n_step > off_diag_nnz_idelays.min():
            ...
```

dionperd:
So, I have improved it:

# Check if the synchronization time is smaller than the minimum delay of the connectivity:
        mask =  numpy.logical_and(self.connectivity.weights != 0,
                                  self.connectivity.idelays != 0 )
        if self.synchronization_n_step > self.connectivity.idelays[mask].min():
            raise ValueError('the synchronization time is too long')

In my opinion:
1. We should make it more robust to TVB semantics' errors. Therefore, better to check for zeros in both weights and idelays.
2. We should allow for nonzero weights and lengths for region self-connections. There can be cases where a node is modeled both as TVB and as a spiking network, e.g., in a scenario where the spiking network doesn't substitute the TVB mean field node, and the coupling is one directional, i.e., from the "whole" TVB mean field region node TO a specific neural patch of it modeled as a spiking network.
Therefore, I prefer if we don't exclude the diagonal from the mask, and require from the user to zero the diagonal of weights to prevent self-connections.

maedoc:
Won't this miss the case of a non-zero weight and a zero idelay which violates some synchronization condition?

dionperd:
> Won't this miss the case of a non-zero weight and a zero idelay which violates some synchronization condition?

Nope!

The delay will be considered as a candidate for min, which determines the synchronization time, only if both the weight AND the delay are not 0.

If any of them are zero, it will be excluded from the mask, and therefore, from being considered for minimum delay.

The point is that we exclude zero delays, and from nonzero delays, we also exclude those that correspond to zero weights.

maedoc:
Why would you exclude zero delays? Doesn't this affect synchronization?

dionperd:
> Why would you exclude zero delays? Doesn't this affect synchronization?

Aha, sorry, I didn't realize you were not aware of that.

Well, we cannot synchronize TVB<->NEST with zero delays.
I think this is a limitation, i.e., for a minimum delay of 1 TVB integration time step.
I don't remember exactly why.

This is also why in tvb-multiscale I set artificial tract_lengths such -and given a connectivity.speed - as to have min(delays) = integrator.dt, and I use weights to remove connections.
This is also why the previous code was failing for me, because my min delays were 0.1, but those delays corresponded to zero weights, and the way the IF statement was, not taking into consideration the weights, I was getting a min delay of 0.1.

maedoc:
> Well, we cannot synchronize TVB<->NEST with zero delays.

I'm aware of that, hence my question, which you answered by saying, basically, that you are careful.  Good for you! :1st_place_medal: This is however an error check in core code, and it should cover inputs from other users who may be less careful, or simply datasets that slip through quality control.  One could still put in a idelay==0 with non zero weight and then get the wrong answer, with you changes, yes?  For instance, a user who wishes to remove delays temporarily idiomatically sets speed to np.inf in TVB.  What happens then?

dionperd:
> > Well, we cannot synchronize TVB<->NEST with zero delays.
> 
> I'm aware of that, hence my question, which you answered by saying, basically, that you are careful. Good for you! 🥇 This is however an error check in core code, and it should cover inputs from other users who may be less careful, or simply datasets that slip through quality control. One could still put in a idelay==0 with non zero weight and then get the wrong answer, with you changes, yes? For instance, a user who wishes to remove delays temporarily idiomatically sets speed to np.inf in TVB. What happens then?

OK. I understand that you think that the code should throw an error for the obvious erroneous case where someone provides delays = 0.0 for valid connections, i.e., with weights != 0.0, instead of those connections being ignored for the synchronization_time check. 

This will force the user to define missing connections via weights = 0.0, even if he forgets to set delays = 0.0.

In the case you mention with speed = inf, there is going to be an error the way the code stands now, since the mask will be empty with all delays = 0.0.

OK. So, lets agree then that the mask will look only at weights, without idelays, and it is the responsibility of the user to not forget any 0 delays without 0 corresponding weights. I am correcting it now.

dionperd:
So, in this new solution

- first of all I made a separate method for configuration of synchronization time, so that we could also write some test.

- I am considering as existing connections (mask) only those with nonzero weights.

- If there are any existing connections,
- I make sure that synchronization_time > min_delay, with no other exclusion of delays, i.e., possible delay = 0.0 without weight = 0.0, will give an error for the CoSimulator, unlike the Simulator, since synchronization_time will be > 0.0, and minimally, equal to integrator.dt...
However, this error will happen only if the user sets the other inputs of CoSimulator, vois etc, and still keeps some delays equal to 0.0, i.e., if he configures CoSimulator for co-simulation, and at the same time, forgets to correct his delays.

tvb_contrib/tvb/contrib/cosimulation/cosimulator.py
```diff
@@ -83,6 +83,27 @@ class CoSimulator(Simulator):
...
+           than the minimum delay of all existing connections.
+           Existing connections are considered those with nonzero weights.
+        """
+        # The synchronization time should be at least equal to integrator.dt:
+        self.synchronization_time = numpy.maximum(self.synchronization_time, self.integrator.dt)
+        # Compute the number of synchronization time steps:
+        self.synchronization_n_step = iround(self.synchronization_time / self.integrator.dt)
+        # Check if the synchronization time is smaller than the minimum delay of the connectivity:
+        existing_connections = self.connectivity.weights != 0
+        if numpy.any(existing_connections.size):
```
maedoc, line 99: This will only be false if the weights matrix is empty. Perhaps you meant w/o `.size`?

maedoc: and you don't really need a nested if here.  This `any(.)` will always return true for any realistic weights matrix, just collapse into a single if-raise.

tvb_contrib/tvb/contrib/cosimulation/cosimulator.py
```diff
@@ -91,14 +112,8 @@ def _configure_cosimulation(self):
...
-        # the synchronization time should be at least equal to integrator.dt:
-        self.synchronization_time = numpy.maximum(self.synchronization_time, self.integrator.dt)
-        # Compute the number of synchronization time steps:
-        self.synchronization_n_step = iround(self.synchronization_time / self.integrator.dt)
-        # Check if the synchronization time is smaller than the delay of the connectivity
-        # the condition is probably not correct. It will change with usage.
-        if self.synchronization_n_step > numpy.min(self.connectivity.idelays[numpy.nonzero(self.connectivity.idelays)]):
-            raise ValueError('the synchronization time is too long')
+        # Configure the synchronizatin time and number of steps
+        self._configure_synchronization_time()
```
maedoc, line 116: this is good.  reducing the length of complicated configuration functions by piecing them into smaller, well named methods bits is good.

maedoc (approved):

maedoc:
I would suggest merging once Lionel approves.

lionelkusch:
I agree with the new conditions.
I will approve it after the modification of the line 99 following the comments of Marmaduke.

dionperd:
> I will approve it after the modification of the line 99 following the comments of Marmaduke.

I corrected line 99.

tvb_contrib/tvb/contrib/cosimulation/cosimulator.py
dionperd: > and you don't really need a nested if here. This `any(.)` will always return true for any realistic weights matrix, just collapse into a single if-raise.

I would suggest to keep it in cases someone wants to perform a simulation without weights for whatever reason.

lionelkusch (approved):
Thank you for the modification.

PR #510, opened 2021-11-04
Bug fixing + improvements on Surface model page.

VinczeRobert:
Applying the html() method in JS is like calling empty() + append(). I thought of generalizing the way the rendering is done with mathjax. Also, there was a bug left on the model surface page after merging TVB-2977.

### DIFF
--- a/framework_tvb/tvb/interfaces/web/static/js/genericTVB.js
+++ b/framework_tvb/tvb/interfaces/web/static/js/genericTVB.js
@@ -134,7 +134,10 @@ function fireOnClick(redirectElem) {
 }
 
 // ---------- Function for rendering HTML elements with Mathjax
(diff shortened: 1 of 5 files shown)


framework_tvb/tvb/interfaces/web/static/js/genericTVB.js
```diff
@@ -133,6 +133,11 @@ function fireOnClick(redirectElem) {
     }
 }
 
+// ---------- Function for rendering HTML elements with Mathjax
+function renderWithMathjax(element, elementToAppend){
+    element.append(elementToAppend);
```
liadomide, line 138: Maybe Add a flag in this function's signature append=true/false and thus avoid the .empty() calls before calling this function?

PR #512, opened 2021-11-09
TVB-2990 Review usage of default cortex instance in tutorials

liadomide:
The module level variable `cortex` in `utils.py` is not recommended, 
thus I replaced it with a parameter to be received by the public function `multiview ` when it is called


liadomide:
@lionelkusch  I will merge this, as other people start to stumble over the wrong usage of the default Cortex.
We can continue to discuss on possible API improvements in Jira.

PR #514, opened 2021-11-10
Transposing Allen connectome  for TVB connectivity

dionperd:
I modified the allen_creator.py adaptor.

However, I cannot regenerate the connectivity.h5 file that is used in the notebooks.

Please help!

### DIFF
--- a/framework_tvb/tvb/adapters/creators/allen_creator.py
+++ b/framework_tvb/tvb/adapters/creators/allen_creator.py
@@ -490,7 +490,7 @@ def construct_structural_conn(projmaps, order, key_ord):
     sc_down = np.concatenate((second_quarter, first_quarter), axis=1)
     structural_conn = np.concatenate((structural_conn, sc_down), axis=0)
     structural_conn = structural_conn / (np.amax(structural_conn))  # normalize the matrix
-    return structural_conn
+    return structural_conn.T  # transpose because TVB convention requires SC[target, source]!
 
 
 # the method returns the centres of the brain areas in the selected parcellation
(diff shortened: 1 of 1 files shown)


maedoc:
Unfortunately, it is unlikely that you would be able to reproduce the results.  Allen frequently updates their ontology and data, and terms of use do not allow us to have a versioned cache, if I recall correctly.  This means that the matrix that's created will vary in time.

dionperd:
> Unfortunately, it is unlikely that you would be able to reproduce the results. Allen frequently updates their ontology and data, and terms of use do not allow us to have a versioned cache, if I recall correctly. This means that the matrix that's created will vary in time.

I am not trying to reproduce results but to reproduce the connectivity that ships with TVB (mouse/allen_2mm/Connectivity.h5) and is loaded by the notebook simulate_for_mouse.ipynb in documentations/demos.
I cannot reproduce it because for some reason, when I list_projects() in interacting_with_Allen.ipynb, I don't find any and the notebook doesn't run...

maedoc:
> reproduce the connectivity that ships with TVB (mouse/allen_2mm/Connectivity.h5)

That's what I meant by reproduce.  Maybe their ontology has changed and now the creator doesn't work.  However, unlike the lack of transpose that you identified, this is not a bug with TVB: the underlying data source is changing over time, and the license for the data does not allow us to cache the source data, only our derived artifacts.  Furthermore, no one has had time to maintain this over the years.

Let's keep this PR simple: we have the transpose now, and we're done.

maedoc (approved):

dionperd:
That's fine! The important thing is that whoever regenerated an Allen connectome will have correctly transposed now.
If someone can also regenerate the connectivity.h5 that ships with TVB even better!
I would prefer if I am not that person! :)

PR #517, opened 2021-11-22
TVB-2994: Stop forms from being resubmitted at page refresh.

VinczeRobert:
This solution should work for any adapter that uses full_adapter_interface.html (AllenCreator, the new Pipeline Creator and maybe others too). Any idea how to generalize this? I tried adding this to genericTVB.js but it didn't work.

### DIFF
--- a/framework_tvb/tvb/interfaces/web/templates/jinja2/flow/full_adapter_interface.html
+++ b/framework_tvb/tvb/interfaces/web/templates/jinja2/flow/full_adapter_interface.html
@@ -42,3 +42,10 @@
         {% endif %}
     </form>
 </div>
+
+<!--! This if statement makes sure that a form won't be resubmitted after refreshing a page -->
+<script>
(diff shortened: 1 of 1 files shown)


VinczeRobert:
Looks like this replaceState method is not going to work in Safari. 

This issue is a more general problem and it can be solved with the Post-Redirect-Get pattern: https://stackoverflow.com/questions/10827242/understanding-the-post-redirect-get-pattern (the first two comments explain it pretty well).

I just don't know if we want to provide some new success page after posting. 

Another possible solution is to just not do anything about it. If the user refreshes the page, he is going to be asked:
"The page that you're looking for used information that you entered. Returning to that page might cause any action that you took to be repeated. Do you want to continue?" and he can just choose No.

Or we can choose to not submit forms, but use onClick events everywhere and have some new ajax calls.

I vote for not doing anything about it. Any thoughts?

PR #518, opened 2021-11-26
TVB-2900: BIDS Derivatives Importer

VinczeRobert:
We got one BIDS derivatives dataset from Charite and we are building here a new Importer into TVB.
The user needs to ZIP its BIDS folder and then upload it into TVB as a single file


framework_tvb/tvb/adapters/uploaders/bids_importer.py
```diff
@@ -0,0 +1,202 @@
...
+    SPATIAL_TOKEN = "spatial"
+    TS_TOKEN = "ts"
+
+    def get_form_class(self):
+        return BIDSImporterForm
+
+    def get_output(self):
+        return [VolumeIndex, StructuralMRIIndex, TimeSeriesVolumeIndex, RegionVolumeMappingIndex]
+
+    def launch(self, view_model):
```
liadomide, line 96: This method had become extremely long. Can it be divided into smaller (maybe private) methods ?
e.g. _build_connectivity(...)

liadomide (approved):
I think this is a good start. In case we don't get clarifications on the NII files soon (2 weeks) I suggest we merge as it is and open another PR for the volumes.

framework_tvb/tvb/adapters/uploaders/bids_importer.py
```diff
@@ -0,0 +1,288 @@
...
+
+            ts_folder = os.path.join(subject_folder, self.TS_TOKEN)
+            if os.path.exists(ts_folder):
+                ts_dict = self.__build_time_series(ts_folder, connectivity)
+
+            spatial_folder = os.path.join(subject_folder, self.SPATIAL_TOKEN)
+            if os.path.exists(spatial_folder):
+                self.__build_functional_connectivity(spatial_folder, ts_dict)
+
+    def __build_connectivity(self, net_folder):
```
SantamRC, line 147: The methods, build_connectivity, build_surface_build_time_series, build_functional_connectivity can be put into their own python files having relevant file names which define their functionalities.

framework_tvb/tvb/adapters/uploaders/bids_importer.py
```diff
@@ -0,0 +1,288 @@
...
+        surface = CorticalSurface()
+
+        # The vertex values can be too small to be seen with the viewer, so we multiply them with orders of 10
+        vertices_mean = numpy.mean(vertices)
+        if vertices_mean < 0.1:
+            vertices = vertices * 1000
+        elif vertices_mean < 1:
+            vertices = vertices * 100
+        elif vertices_mean < 10:
+            vertices = vertices * 10
```
liadomide, line 212: This scaling mechanism could go into the `CorticalSurface` class

liadomide (approved):

liadomide:
Sonar reports few code smells. I think all those should be fixed

tvb_framework/tvb/tests/framework/adapters/uploaders/bids_zip_importer_test.py
```diff
@@ -0,0 +1,118 @@
...
+
+try:
+    import tvb_data.bids
+
+    BIDS_DATA_FOUND = True
+except ImportError:
+    BIDS_DATA_FOUND = False
+
+
+@pytest.mark.skipif(not BIDS_DATA_FOUND, reason="Older or incomplete tvb_data")
```
liadomide, line 50: @popaula937 @davidbacter01 Do you think this is reasonable for the time being, as we can not have a buds demo dataset published yet on Zenodo ?

popaula937: I agree with this change. Until the dataset is published, we should run this test locally whenever changes are made to the BIDS importer.

tvb_framework/tvb/adapters/uploaders/bids_importer.py
```diff
@@ -0,0 +1,279 @@
...
+        return BIDSImporterModel
+
+    @staticmethod
+    def get_upload_information():
+        return {
+            'uploaded': '.zip'
+        }
+
+
+class BIDSImporter(ABCUploader):
```
popaula937, line 80: Rename to BIDSDerivativesImporter?

liadomide: I consider this change, but I am not decided if it is a good idea or not.... 
We would then need to change also `BIDSImporterModel` and `BIDSImporterForm` into the more explicit names. Wouldn't these become too long ?

tvb_framework/tvb/tests/framework/adapters/uploaders/bids_zip_importer_test.py
liadomide: Also, in the current form of the code, even after merge, the new BIDS derivatives Importer is not visible in TVB web GUI, until Charite gives us an OK on it. The code being merged into master is not much different than having it in a slowly deprecating branch, but at least we don't need to rebase and keep the branch in synch.

tvb_framework/tvb/adapters/uploaders/bids_importer.py
popaula937: On the other hand, TVB cannot read the official BIDS which describes empirical data...
Would it be enough to add a description near the upload field? (to explain that it expects derivatives)

tvb_framework/tvb/tests/framework/adapters/uploaders/bids_zip_importer_test.py
popaula937: Fully agree to merge it even if is not visible, would be easier to maintain.

tvb_framework/tvb/adapters/uploaders/bids_importer.py
liadomide: Is this similar to what you have in mind ?
image

popaula937: Yes, this is informative enough from my POV

liadomide (approved):

PR #520, opened 2021-12-06
removed mpl_toolkits - not a standalone package

i-Zaak:


liadomide:
Correct, `mpl_toolkits` is not right to be put there.
We should replace it with `matplotlib`.
It is under EXTRA, so it won't block normal installation, and should be treated as an optional dependency,

i-Zaak:
The matplotlib is already in library requirements, so deletion in the extras should be enough...

liadomide (approved):

PR #521, opened 2021-12-06
BOLD analyzer fixes

i-Zaak:


i-Zaak:
This is to make the normalization of inputs conditional as it is causing numerical issues with certain models. I have also removed the `dt` attribute, as it was not used consistently (no input interpolation), and instead this is now derived from the sampling rate of the  input data.

I see the tests have passed, but I assume the code in `fmri_balloon_adapter.py` should be adapted to reflect the changes. Any volunteers?

liadomide (approved):

maedoc:
This can happen when merge conflicts are resolved incorrectly.  The easiest thing would be to remerge I think

liadomide:
I rebased my local master branch on the remote (partially vanished) master and pushed again. It should be fine now.

Not sure what happened previously. @maedoc  is right in terms of theoretically possible scenario ..... I will look again into the rules on master and think if we can avoid it in the future.

PR #522, opened 2021-12-07
Generalize refreshing subforms.

VinczeRobert:


VinczeRobert:
I couldn't find an easy and totally clean solution for this.

There is a little logistical issue. The surface model param page does not have an associated View Model but it has a SurfaceContextModelParameters object which at least from this point of view acts as a view model.

PR #537, opened 2022-03-18
User tab from footer not active on user page

Spectre-ak:
User tab is not selected when the user page is opened.
user_tab_not_enabled

I found that the `section_name`  inside [footer.html](https://github.com/the-virtual-brain/tvb-root/blob/master/framework_tvb/tvb/interfaces/web/templates/jinja2/footer.html) is empty for user's page.
So added 
```
{% if section_name is not defined %}
        {% set section_name = 'user' %}
{% endif %}
```
user_tab_enabled


liadomide:
Good catch. I reported this as a bug in TVB Jira. 
As solution I propose the alternative in #538

PR #540, opened 2022-03-31
WIP: Parameters api

i-Zaak:
Missing:

- [ ] a demo
- [x] check if some parts in the showcase can be refactored here 
- [ ] polish (e.g. docstrings, tests)


maedoc:
@i-Zaak thanks I will look at the `deepcopy` issue since that's an annoying thing to have crop up.

i-Zaak:
The showcase has implementation of a UNICORE runner, which is IMHO better to be left outside the generic API presented here. However I've also subclassed the joblib exec there to make use of the Numba backend. This we should have available as config of the JoblibExec here.

maedoc:
Unicore is a good point, since it's being solved in the framework regardless, we should try to make reuse of it.

maedoc:
wdyt about generalizing `values: List[List[Any]]`, e.g. 

```python
class Value:
    # implement iter protocol

class SimSeq:
    ...
    values: List[Value]
```
Ideally I'd want to cover use cases of

- Cartesian product grid search e.g. `values=itertools.product(*[np.r_[:1.0:10j] for _ in range(3)])`
- Precomputed arrays of values e.g. `values=saltelli.sample(problem, 1024)` (from [SAlib](https://salib.readthedocs.io/en/latest/basics.html#generate-samples))
- SBI obv
- Proper distribution support or at least normal, log normal

For the last case, we'd have parallel sampling of the distributions as in Monte-Carlo, and "single-threaded" MCMC, where the `SimSeq` is realized sample by sample from a small number of chains.  I'm not sure if these are similar enough to warrant a common implementation or not.

maedoc:
I would also be ok if we just merge this to have the basic parts in place, and leave generalization to a later issue.  ok @i-Zaak ?

maedoc:
can we merge something now to avoid getting stale? or go back to drawing board and redo?

liadomide:
I think we can close this, as a more advanced API can be found here:
https://github.com/the-virtual-brain/tvb-inversion/blob/main/tvb_inversion/parameters.py
Do you agree @i-Zaak

PR #541, opened 2022-03-31
Fix warnings in library

maedoc:
A bunch of warnings crop up when running pytest on the library.  These include

- deprecation warnings of use of NumPy aliases like `bool`, `int`, `float` and `str`, which I replaced with concrete data types
- test model classes starting with `Test` are seen by pytest as unit tests but warns that it can't initialize them because they implement `__init__`, so I renamed them to `ModelTestBlaBla`
- one numerical test had no `atol`, fixed generously at 1e-3


liadomide (approved):

liadomide:
Could we merge this into main ?
It looks good to have it as soon as possible integrated

PR #551, opened 2022-04-19
Add few readers from byte streams in tvb_library

popaula937:
We want to expose in the main DataTypes API also the possibility to be built from a file stream, similarly with building a DT from a file path (connectivity.zip).
This need arise in tvb-widgets module, when building tvb DataTypes from files stored on EBRAIN Drive, but it is generic enough to worth a place in the DataTypes API, and allows us not to duplicate the building part of a DT someplace else

### DIFF
--- a/scientific_library/tvb/datatypes/connectivity.py
+++ b/scientific_library/tvb/datatypes/connectivity.py
@@ -37,6 +37,7 @@ The Connectivity datatype.
 """
 
 from copy import copy
+from io import BytesIO
 
 import numpy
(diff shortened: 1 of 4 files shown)


liadomide (approved):
All changes look good.
I would like to see the tests running before we merge, but for some reason we have problems in framework, thus I will attempt a quick fix for the configuration

liadomide:
@maedoc-amu Would you agree with this slight change in DataTypes API?

liadomide (approved):

PR #564, opened 2022-05-16
Rateml ebrains fix

DeLaVlag:
small fix for rateml usage on ebrains for args parser
and additionally the hpc version of the zerlaut model


DeLaVlag:
Hi, which bug needs to be fixed here?

DeLaVlag:
I don't see that much confliction. Is there something I need to fix?

maedoc:
Yes, you have to resolve the conflict or the branch can't be merged.  If the conflict is simple, then you are lucky.

liadomide:
> Hi, which bug needs to be fixed here?
Hi. 
The tag `bug` on this PR was added because I understand this code fixes a problem/bug in the previous versions, not for being buggy this PR itself.

But a merge from here was indeed problematic, as things changed in `master` since this PR was opened. Mainly we renamed `scientific_library` folder into `tvb_library`. Usually a simple rebase with master in existing older PRs should solve conflicts, except when new files were added in the old folder, then those files need to be moved to the new location in a new commit. I just did these steps here, and the PR should be in good technical state (hopefully all tests pass). Remains up to @maedoc to judge the logic of these changes in detail.

maedoc:
Is there any demo of using the Zerlaut model?  A notebook or similar?  It's a lot of code to add without any tests etc.

DeLaVlag:
A generic notebook demo how to run the Zerlaut on HPC through pyunicore is added. Also the Zerlaut model has been added to the tests.

DeLaVlag:
Still in progress... I tried to fix the tests but now I get lots of errors related to the runner. Also the lib tests didn't get picked up.

maedoc:
The lib runner has a network problem, I'm working on it.  You could ignore it for the moment.

maedoc:
@DeLaVlag there's a test which isn't passing, can you take a look? does it pass on your system?

maedoc:
sorry for the force push, but it dropped all the deleted files from teh bct which is a git fart not related to this pr.

maedoc (approved):
I think this PR is , but the overall workflow of rateml as currently implemented needs improvement, ideally adhering to the simulator backend approach we are now using else where.  In particular: generating and git-committing nearly identical pieces of code every iteration isn't great, and the drivers themselves are almost identical too.  We can address this in the next PR.

PR #565, opened 2022-05-16
TVB-3009 Rename scientific_library into tvb_library and framework_tvb into tvb_framework

liadomide:
Rename the folders with the source code, to resemble the other folders and the released packags


liadomide:
Due to the manner they were written, RateML tests will fail in this branch, until we merge this PR into master.

`schema_file = urlopen("https://raw.githubusercontent.com/the-virtual-brain/tvb-root/master/tvb_library/tvb/rateML/rML_v0.xsd")` - path does not exist, yet.

@DeLaVlag @maedoc  Not critical, but maybe we can brainstorm for something better here ? 
Otherwise, I would merge this PR soon...

maedoc:
It would preferable to have the schema file read locally (+ with versioned release via GH asset or tvb.org/bin folder if required for other use cases).  Some systems won't have internet access (the French HBP HPC resources for instance) which would exclude rateML from functioning, which doesn't make much sense.  @DeLaVlag can we use a local open for this?  e.g. in `tvb/rateML/XML2model.py`, placing at the top
```python
xsd_fname = os.path.join(
    os.path.abspath(os.path.dirname(__file__),
    "rML_v0.xsd"))

with open(xsd_fname, 'r') as fd:
    xsd_content = fd.read()
```
and then all uses of `urlopen('....rML_v0.xsd')` are replaced with
```python
from tvb.rateML.XML2model import xsd_content as rml_xsd
```

tvb_library/tvb/rateML/XML2model.py
```diff
@@ -172,18 +172,13 @@ def set_template(self, name):
...
 
-        # Global XSD file location
-        schema_file = urlopen(
-            "https://raw.githubusercontent.com/the-virtual-brain/tvb-root/master/scientific_library/tvb/rateML/rML_v0.xsd")
-        xmlschema = etree.XMLSchema(etree.parse(schema_file))
+        xsd_fname = os.path.join(os.path.abspath(os.path.dirname(__file__)), "rML_v0.xsd")
+        xmlschema = etree.XMLSchema(etree.parse(xsd_fname))
         xmlschema.assertValid(etree.parse(self.xml_location))
-        logger.info("True validation of {0} against {1}".format(self.xml_location, schema_file.geturl()))
+        logger.info("True validation of {0} against {1}".format(self.xml_location, xsd_fname))
```
liadomide, line 181: Above I changed the code similarly to what @maedoc  suggested.

PR #568, opened 2022-05-20
Fix TVB-3015

maedoc:
This address the library issue in TVB-3015.


liadomide (approved):

maedoc:
Coverage result seems wrong ?
image

liadomide:
Indeed. Integration with sonar is still not complete. 
Tests with coverage are executed, but the source code files are not correctly matched by Sonar with the coverage.xml resulting files. Something still on the TODO list :-/

PR #573, opened 2022-06-27
WIP: AdEx model update

i-Zaak:
The purpose of this pull request is to consolidate the various implementations of the AdEx model, create relevant demos, and change the approach taken to derived parameters for this model.


i-Zaak:
For start, I've updated the first order Zerlaut model according to the EBRAINS gitlab repository which is linked from the paper:
- https://gitlab.ebrains.eu/kancourt/tvb-adex-showcase3-git/-/blob/master/tvb_model_reference/src/Zerlaut.py#L10

In particular, this meant splitting the `p_connect` parameter into `p_connect_e` and `p_connect_i`. In addition the `tvb-adex-showcase3` variants have additional `noise` state variable - this is something we need to discuss, as it would be cleaner to have the (colored) noise handled by the `Noise` module. 

@rcofre will lead this from here.

maedoc:
isn't the OU something you could do with the colored noise support?

i-Zaak:
Well, not really unfortunately. The issue is that the noise drive enters same as the other external inputs (e.g. connectivity coupling), and is then dragged through the transfer functions:

https://github.com/the-virtual-brain/tvb-root/blob/6d075f377d2411743534e27bfccb47fc754968e0/tvb_library/tvb/simulator/models/zerlaut.py#L714)

i-Zaak:
One thing I've noticed is that the first order model will need a more careful review (at least the dfun). To be done together with the examples reproducing plots from papers
.

rcofre:
Thanks Jan for this first update.

Best,

R.

i-Zaak:
I've consolidated the notebooks and reviewed the changes to the model. If there are no objections, I'll merge early next week. Further potential improvements and fixes to be done in separate pull requests.

liadomide (approved):

PR #578, opened 2022-07-19
TVB-2878 Implement Deepcopy

liadomide:
- [x] all **checks** should pass
- [x] **benchmark** our implementation against the default deepcopy implementation
- [x] answer the question: **what can we give more** than the default deepcopy implementation ?
- [x] validate/document what happens with **random stream generator** on deepcopy

### DIFF
--- a/tvb_library/tvb/basic/neotraits/_core.py
+++ b/tvb_library/tvb/basic/neotraits/_core.py
@@ -217,7 +217,7 @@ class HasTraits(object):
         provides a unique id for example for a model configuration
         """  # these strings are interpreted as docstrings by many tools, not by python though
 
-        self.title = '{} gid: {}'.format(self.__class__.__name__, self.gid)
+        self.set_title()
         """ a generic name that the user can set to easily recognize the instance """
 
         for k, v in kwargs.items():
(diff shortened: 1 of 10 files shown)


tvb_library/tvb/basic/neotraits/_core.py
```diff
@@ -304,3 +304,19 @@ def summary_info(self):
             except TraitError:
                 ret[aname] = 'unavailable'
         return ret
+
```
liadomide, line 307:
``` 
class HasTraits...

def duplicate(self):
        duplicate = copy.deepcopy(self)
        duplicate.gid = uuid.uuid4()
        duplicate.configure()
        return duplicate
```
Could this work for handling the gid on the root, but not the rest?

tvb_library/tvb/basic/neotraits/_core.py
```diff
@@ -304,3 +304,19 @@ def summary_info(self):
...
+    def __deepcopy__(self, memodict={}):
+        """
+        Method for deep copying a HasTraits object correctly, event when its attributes are of type HasTraits as well.
+        """
+        cls = type(self)
+        copied = cls()
+        for k in cls.declarative_attrs:
+            attr = getattr(self, k)
+            if isinstance(attr, HasTraits) or isinstance(attr, numpy.random.RandomState):
```
liadomide, line 316: This can be shortened: `isinstance(x, (A, B, ...))`

PR #580, opened 2022-07-21
TVB-3023 Mutable defaults

liadomide:


popaula937 (approved):
The BalloonModel also has an integrator attribute, we should change the default there as well.

romina1601:
The Projection monitor has an 'obsnoise' attribute with a default that should be changed.

romina1601 (approved):

PR #582, opened 2022-07-25
TVB-2951. Extract H5 API to serialize datatypes 

popaula937:


tvb_framework/tvb/core/neocom/_h5loader.py
```diff
@@ -268,8 +268,8 @@ def load_ht_function(sub_gid, traited_attr):
 
 class ViewModelLoader(DirLoader):
```
liadomide, line 269: Do I understand correctly that `ViewModelLoader` as well as the new `DtLoader` are somehow similar with the previous `TVBLoader`, the main difference being that the later uses the relational DB for some steps, while the first 2 are solely relying on the disk structure ?
Maybe this aspect can be documented on the classes a bit better.

liadomide (approved):

tvb_framework/tvb/core/neocom/_h5loader.py
popaula937: Yes, that is one difference. I will document it, thanks for the suggestion.
TVBLoader is also based on the TVB storage folder structure. The other 2 write in a directory provided as input.

liadomide: Thanks for the details. In the future perhaps we can get rid of the prev TVBLoader all together, as part of our tvb-storage refactor

popaula937: Please check whether the docs are informative enough

liadomide: very informative now. Thank you!

liadomide (approved):

PR #590, opened 2022-09-06
TVB-3030. Update notebook documenting the usage of the siibra API

romina1601:


popaula937:
The API examples look good to me, but the first cells holding functions can be removed, as they are already part of the TVB API.

liadomide (approved):

PR #591, opened 2022-09-08
TVB-3025 bctpy

liadomide:
Replace matlab BCT with bctpy python library.
Add dependency towards `bctpy`
Remove revious matlab execution related code


liadomide:
We should wait for #590 to be merged, then rebase, as some numpy and setuptools updates are producing failures here.

tvb_framework/tvb/adapters/analyzers/bct_adapters.py
```diff
@@ -130,34 +108,25 @@ def get_required_disk_size(self, view_model):
...
+        value.data_value = str(float(result_value))
         value.data_type = 'float'
         value.data_name = title
         return self.store_complete(value)
 
-    def build_int_value_wrapper(self, result, key, title=""):
+    def build_int_value_wrapper(self, result_int, title=""):
         value = ValueWrapper()
-        value.data_value = str(int(result[key]))
+        value.data_value = str(int(result_int))
```
davidbacter01, line 129: if the result is expected to be int is this conversion to int still required?

liadomide: The result is int-like, but not always int, thus the conversion is required.

tvb_framework/tvb/adapters/analyzers/bct_degree_adapters.py
```diff
@@ -121,81 +117,78 @@ class Strength(Degree):
...
-        value1 = self.build_float_value_wrapper(result, 'kden', title="Density")
-        value2 = self.build_int_value_wrapper(result, 'N', title="Number of vertices")
-        value3 = self.build_int_value_wrapper(result, 'K', title="Number of edges")
         return [value1, value2, value3]
 
 
 class DensityUndirected(DensityDirected):
     """
     """
     _ui_name = "Density Unirected: Undirected (weighted/binary) connection matrix"
```
davidbacter01, line 183: I think there is a small typing error here. Shouldn't it be "Density Undirected:..."?

davidbacter01 (approved):

tvb_framework/tvb/adapters/analyzers/bct_adapters.py
davidbacter01: That makes sense. Thanks for the clarification!

davidbacter01 (approved):

PR #593, opened 2022-10-04
Update jansen_rit.py

JulieCB:
Update error in description of parameter nu_max.


maedoc:
@dionperd could you have a look plz?

PR #594, opened 2022-10-09
TVB-2946 Adjust sample rate in wavelet to be in kHz 

liadomide:
In the past TimeSeries.sample_rate property was always returning 1/sample_period (which in most of the cases - e.g. the default simulations was returning kHz).

Changes in 2020 - 2021 into TimeSeries.sample_rate implementation improved that, to take into account the Ts.sample_period_unit, but also made the result Hz.

To reproduce results like here: https://docs.thevirtualbrain.org/tutorials/tutorial_1_BuildingYourOwnBrainNetworkModel.html#tutorial-1-buildingyourownbrainnetworkmodel
As reported here https://groups.google.com/g/tvb-users/c/lLT9j7i-ehI/m/KANAqlEKAQAJ?utm_medium=email&utm_source=footer, we should use in the Wavelet analyzer kHz.

@i-Zaak @maedoc  please have a look and confirm this is the right way to fix this bug

### DIFF
--- a/tvb_library/tvb/analyzers/wavelet.py
+++ b/tvb_library/tvb/analyzers/wavelet.py
@@ -114,7 +114,8 @@ def compute_continuous_wavelet_transform(time_series, frequencies, sample_period
     log.debug("freqs")
     log.debug(narray_describe(freqs))
 
-    sample_rate = time_series.sample_rate
+    # We need this to be kHz (see TVB-2946)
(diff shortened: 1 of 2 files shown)


maedoc:
It looks like the right start, I think the use of `sample_period` needs to be adjusted as well, since it's still in ms and used to size the result array and skip time points.

PR #599, opened 2022-10-20
run_sim entry-point; single license in codegen

i-Zaak:
TODO before merge:
- [x] tests
- [x] (bonus) demo notebook


tvb_library/tvb/simulator/backend/templates/np-dfuns.py.mako
```diff
@@ -1,32 +1,32 @@
...
-#   CITATION:
-# When using The Virtual Brain for scientific publications, please cite it as follows:
-#
-#   Paula Sanz Leon, Stuart A. Knock, M. Marmaduke Woodman, Lia Domide,
-#   Jochen Mersmann, Anthony R. McIntosh, Viktor Jirsa (2013)
-#       The Virtual Brain: a simulator of primate brain network dynamics.
-#   Frontiers in Neuroinformatics (7:10. doi: 10.3389/fninf.2013.00010)
-#
-#
+## -*- coding: utf-8 -*-
```
maedoc, line 1: good catch

maedoc: there are other files with only single `#` can you double check at least the makos related to this PR?

tvb_library/tvb/tests/library/simulator/backend/npbackend_test.py
```diff
@@ -53,40 +53,29 @@
...
     def _test_mvar(self, integrator):
         pass # TODO
 
-    def _test_integrator(self, Integrator, delays=False):
+    def _test_integrator(self, Integrator, delays=False, nsig=False):
         dt = 0.01
         if issubclass(Integrator, IntegratorStochastic):
-            integrator = Integrator(dt=dt, noise=Additive(nsig=np.r_[dt]))
+            if nsig:
+                integrator = Integrator(dt=dt, noise=Additive(nsig=np.r_[dt,dt*2]))
```
maedoc, line 76: why is there a dt*2 here? is this a MPR-ism?

maedoc (approved):

maedoc:
@i-Zaak have look at comments see what you think otherwise we can just merge.   thanks for the demo notebook!

tvb_library/tvb/tests/library/simulator/backend/npbackend_test.py
i-Zaak: Yup - we want the noise to $V$ to be approx twice the noise to $r$. This has actually uncovered the bug in the backed, as it was not handling the extended nsig correctly (the shape after `Simulator.configure` is `(2,1,1), but the mode dimension is dropped elsewhere).

maedoc: ok makes sense.

PR #602, opened 2022-10-25
HBP-189 Fix tests in tvb-contrib related to cosimulation

dionperd:
I made corrections to CoSimulator tests in contrib, which were failing because the default connectivity had 0 delays without 0 weights for self connections.

I added a function to correct tract_lengths given a choice for synchronization time and connectivity speed, such as that all connections have a delay at leats equal to the synchronization time.

There were still some tests failing in the cases of bad cosimulation input or of lack of cosimulation input, when testing differences for every single time point. Actually, they were failing only for the first two time points.

I am not sure why. Therefore, I corrected them for the moment by testing for total sum of the differences across the whole simulation. @lionelkusch could probably help with this. They are marked tih TODOs.


liadomide (approved):

romina1601 (approved):

lionelkusch:
From the discussion with Denis, we define the bets way to solve the problems.

PR #604, opened 2022-10-26
HBP-199. Move/add tests from tvb-multiscale to tvb-contrib

romina1601:
This PR also includes [HBP-200](https://req.thevirtualbrain.org/browse/HBP-200), which relates to running the tvb-contrib tests inside a Github Actions workflow.

EDITED: This PR was waiting for #602 otherwise builds will fail with the recent inclusion of tvb-contrib tests in the Github Action

### DIFF
--- a/tvb_contrib/tvb/contrib/tests/model/model_test.py
+++ b/tvb_contrib/tvb/contrib/tests/model/model_test.py
@@ -47,7 +47,7 @@ from tvb.simulator.plot.phase_plane_interactive import PhasePlaneInteractive
 LOG = get_logger(__name__)
 
 
-@pytest.mark.skipif(True, reason="Because it opens a window for manual inspection")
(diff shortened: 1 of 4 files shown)


tvb_contrib/tvb/contrib/tests/model/model_test.py
```diff
@@ -41,12 +41,14 @@
 from tvb.contrib.simulator.models.morris_lecar import MorrisLecar
 from tvb.contrib.simulator.models.wong_wang import WongWang
 from tvb.simulator.integrators import HeunDeterministic
-from tvb.tests.framework.core.base_testcase import BaseTestCase
+# from tvb.tests.framework.core.base_testcase import BaseTestCase
```
liadomide: This commented line could be removed directly.
We should not import from tvb-framework in tvb-contriv

liadomide (approved):

PR #608, opened 2022-10-27
TVB-3036. Review tests.

popaula937:


tvb_framework/tvb/tests/framework/interfaces/rest/bids_monitor_test.py
```diff
@@ -0,0 +1,201 @@
+import shutil
```
liadomide, line 1: Ideally this file should include the copyright header

liadomide (approved):

PR #611, opened 2022-11-03
HBP-189 Fixes for tvb-multiscale in tvb-contrib

romina1601:
This PR contains fixes for bugs in tvb-contrib, which were found when testing tvb-multiscale.


liadomide (approved):

dionperd:
I can see that it is more consistent with TVB traits to configure outside the __init__.
However, I would have to change all my examples now if we do so.
I can do it if I have too.

However, what happens when I want to create a new time series with some get_subspace / item call?

Will I have to configure always the new instance?

dionperd (approved):

romina1601 (commented):
The reason for removing the `configure()` call from the `__init__` method is that it leads to errors. For example, when trying to load a TS from a h5 file, the constructor for the TS gets called before setting a connectivity on it. Calling configure() inside the constructor, but before setting a connectivity for the TS yields an error. (the constructor is called in [_h5loader.py](https://github.com/the-virtual-brain/tvb-root/blob/90eacd7386b2c8c4455ee73028b15a5e43c4e946/tvb_framework/tvb/core/neocom/_h5loader.py#L125))

PR #614, opened 2022-11-04
HBP-203. Add option to disable imports on projects and restrict operation size

popaula937:
Also includes a DB migration script.

Before merge, we should decide:
- what should be the max_operation_size allowed on Default_Project
- messages displayed, DB columns naming informative enough
- default value for max_operation_size to remain null


tvb_framework/tvb/interfaces/web/templates/jinja2/project/upload_data_overlay.html
```diff
@@ -8,9 +8,17 @@
             <fieldset class="toolbar-inline">
                 <ul>
                     <li>
-                        <button type="submit" class="action action-upload" onClick="showOverlayProgress()">Upload
+                        <button type="submit" class="action action-upload disabled" {{ 'disabled' if disable_imports }}
```
liadomide, line 11: `class ="disabled"` is added by intention here, or some left-over ?

tvb_framework/tvb/interfaces/web/templates/jinja2/project/editone.html
```diff
@@ -82,7 +82,28 @@ <h4>
...
+                    </dd>
+
+                    <dt>
+                        <label for="max_operation_size">Max allowed size for one operation (KB)</label>
+                    </dt>
+
+                    <dd>
+                        <p class="field-data"><input tabindex='5' id="max_operation_size" type="number" name="max_operation_size"
+                                                     value="{{ data.max_operation_size }}" {{ 'disabled' if not editUsersEnabled }}/>
+                        </p>
```
liadomide, line 106: I believe after these, somewhere on the page `tabindex='4'` exists already, and it should be change into `tabindex='6'`

And I believe that `max_operation_size ` could be a validated field (e.g. to be a positive number), thus an error message would be possible to appear nearby.

tvb_framework/tvb/interfaces/web/controllers/project/project_controller.py
```diff
@@ -733,3 +735,5 @@ class EditForm(formencode.Schema):
     administrator = validators.UnicodeString(not_empty=False)
     project_id = validators.UnicodeString(not_empty=False)
     visited_pages = validators.UnicodeString(not_empty=False)
+    disable_imports = validators.Bool()
+    max_operation_size = validators.UnicodeString(not_empty=False)
```
liadomide, line 739: Maybe here we can use a number validator class for `max_operation_size` (e.g. `validators.Int(...`) ?

tvb_framework/tvb/interfaces/web/controllers/base_controller.py
```diff
@@ -60,6 +60,7 @@ class BaseController(object):
     """
     This class contains the methods served at the root of the Web site.
     """
+    MAX_SIZE_ERROR_MSG = "Max operation size has been exceeded"
```
liadomide, line 63: `MAX_SIZE_ERROR_MSG = "Max operation size has been exceeded. The current project admin can change this limit in Project - Basic Properties"` ?

tvb_framework/tvb/interfaces/web/controllers/simulator/simulator_controller.py
```diff
@@ -607,8 +607,15 @@ def launch_pse(self, **data):
         burst_config = self.context.burst_config
         burst_config.start_time = datetime.now()
         burst_config.range1 = range_param1.to_json()
+        len_range = len(range_param1.get_range_values())
         if range_param2:
             burst_config.range2 = range_param2.to_json()
+            len_range += len(range_param2.get_range_values())
```
liadomide, line 613: Shouldn't we multiply `len_range *= len(range_param2.get_range_values())` ?

tvb_framework/tvb/core/services/project_service.py
```diff
@@ -109,7 +109,8 @@ def store_project(self, current_user, is_create, selected_id, **data):
         if started_operations > 0:
             raise ProjectServiceException("A project can not be renamed while operations are still running!")
         if is_create:
-            current_proj = Project(new_name, current_user.id, data["description"])
+            current_proj = Project(new_name, current_user.id, data["max_operation_size"], data["description"],
+                                   data["disable_imports"])
```
liadomide, line 113: Maybe this part could be re-written into something like bellow (as I see some tests failing from this area)
```
            max_size = data["name_size"] if "max_size" in data else None
            disable_imports = data["disable_imports"] if "disable_imports" in data else False
            current_proj = Project(new_name, current_user.id, max_size, data["description"], disable_imports)
```

liadomide (changes requested):
It looks decent to me.
Only polishing things seem to be left: I added some comments bellow

tvb_framework/tvb/core/services/project_service.py
popaula937: Having the validators implemented should ensure that these keys are always present in the data dict, even if they have None or False values.
In this case, shouldn't we adjust the tests to prepare the data dict in the same manner?

liadomide (approved):

PR #615, opened 2022-11-04
TVB-3042 TVB-3043 Fixes for SiibraCreator token handling

liadomide:
- [x] don't show in an operation overlay the actual value in case of UserSessionStrField
- [x] store ebrains_token as a user-session field, instead of UserPreference 
- [x] test with keycloak HBP

### DIFF
--- a/tvb_framework/tvb/adapters/creators/siibra_creator.py
+++ b/tvb_framework/tvb/adapters/creators/siibra_creator.py
@@ -36,20 +36,17 @@ the EBRAINS Knowledge Graph using siibra
 """
 
 import os
-import siibra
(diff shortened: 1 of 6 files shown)


romina1601 (approved):

popaula937 (approved):
The changes look good. 
But testing the siibra creator locally with the default setup, gives me an error:
`'NoneType' object has no attribute 'items'`
and after debugging, it seems to be caused by the tracts being empty. The weights are downloaded correctly.

liadomide:
> testing the siibra creator locally

Yep, SiibraCreator is having new problems (I suspect a new release/update in KG): https://req.thevirtualbrain.org/browse/TVB-3044

PR #616, opened 2022-11-05
HBP-205 Replace usage of canvas.set_window_title 

liadomide:
Use `canvas.manager.set_window_title` instead 
the previous form was deprecated and is now removed in Matplotlib 3.6


dionperd (approved):
I trust you that this is the solution to the bug problem!

liadomide:
> I trust you that this is the solution to the bug problem!

It does. I just managed to test in https://tvb-multiscale-dev.apps.hbp.eu/ with this branch.

PR #625, opened 2022-11-10
TVB-3049: Add constraints for non admin project members

davidbacter01:
- members can no longer edit project name of which they are not admin
- members can no longer delete projects of which they are not admin
- members have option to leave a project


liadomide (approved):
I found and fixed a problem on edit description is only member.
The submit expect to find the project name also.

PR #629, opened 2022-11-14
TVB-3053. Better Error handling

romina1601:
Offer the users of TVB (especially those using the web app) more suggestive error messages when operations related to siibra do not work properly.

### DIFF
--- a/tvb_framework/tvb/adapters/creators/siibra_creator.py
+++ b/tvb_framework/tvb/adapters/creators/siibra_creator.py
@@ -36,6 +36,7 @@ the EBRAINS Knowledge Graph using siibra
 """
 
 import os
+from siibra.retrieval.requests import SiibraHttpRequestError
 from tvb.adapters.creators import siibra_base
 from tvb.adapters.datatypes.db.connectivity import ConnectivityIndex
 from tvb.adapters.datatypes.db.graph import ConnectivityMeasureIndex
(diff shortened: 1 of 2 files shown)


tvb_framework/tvb/adapters/creators/siibra_creator.py
```diff
@@ -175,8 +176,11 @@ def launch(self, view_model):
         # list of all resulting indices for connectivities and possibly connectivity measures
         results = []
 
-        conn_dict, conn_measures_dict = siibra_base.get_connectivities_from_kg(atlas, parcellation, subject_ids,
-                                                                               compute_fc)
+        try:
+            conn_dict, conn_measures_dict = siibra_base.get_connectivities_from_kg(atlas, parcellation, subject_ids,
+                                                                                    compute_fc)
+        except SiibraHttpRequestError:  # this type of Error is raised in case of invalid token, but it might change?
```
liadomide, line 182: I saw an HTTP 403 code in the actual Siibra exception in case of the invalid token.
Can you check that too ?
If that is present also (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) then it is clear the token issue, otherwise, raise a generic message, something like:"We could not complete the operation. Please check the logs and contact the development team from TVB, siibra, EBRAINS KG"

liadomide (changes requested):
Looks good. I suggest one some change on the token check, to be sure it is it.

tvb_framework/tvb/adapters/creators/siibra_creator.py
romina1601: Thanks for the suggestion, will implement it.
I personally got only error 401, but I see in their code that 403 is also possible, so I will include both of them.

liadomide (approved):

PR #631, opened 2022-11-17
TVB-3029: Use CLIENT_SECRET in Siibra tests & License adjustments

liadomide:


tvb_framework/tvb/tests/framework/adapters/creators/siibra_creator_test.py
```diff
@@ -46,7 +48,14 @@ def transactional_setup_method(self):
 
     def test_happy_flow_launch(self, operation_factory):
         view_model = SiibraModel()
-        view_model.ebrains_token = os.environ[CLB_AUTH_TOKEN_KEY]
+        if CLB_AUTH_TOKEN_KEY in os.environ:
+            view_model.ebrains_token = os.environ[CLB_AUTH_TOKEN_KEY]
+        else:
+            # This path might be too white-box, as we are replicating siibra mechanism of retrieving
```
romina1601, line 54: That's true, but I don't see at the moment a better way to retrieve the kg_token, so I think this is the best we can do right now

romina1601 (approved):
Everything looks good. I also tested these changes locally and there were no errors. From my side, this can be merged. 💯

PR #633, opened 2022-12-07
Add a tvb_library Pipfile

maedoc:
This adds Pipfile for the tvb_library to improve the library setup process for development purposes with pipenv, addressing the issue of divergent requirements.txt, Dockerfiles, and CI build scripts (or maybe I'm not up to date on the latest?).  I don't think it's smart to merge changes to every CI script at once, so I just want to get the Pipfile in there first, start using it a bit, and later change the Dockerfile/CI scripts.

### DIFF
--- a/tvb_library/Pipfile
+++ b/tvb_library/Pipfile
@@ -7,10 +7,11 @@ name = "pypi"
 numba = "*"
 pytest = "*"
 matplotlib = "*"
-tvb-data = "*"
 scipy = "*"
 mako = "*"
 autopep8 = "*"
+tvb-gdist = "*"
+pylems = "*"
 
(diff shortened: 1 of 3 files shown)


tvb_library/Pipfile
```diff
@@ -0,0 +1,25 @@
...
+
+[dev-packages]
+pytest = "*"
+pytest-cov = "*"
+pytest-xdist = "*"
+pytest-benchmark = "*"
+pytest-mock = "*"
+
+[requires]
+python_version = "3.9"
```
liadomide, line 25: If I am not mistaken EBRAINS lab is Python 3.8, maybe test against that with priority?

PR #637, opened 2023-01-11
TVB-3062 EpileptorT TVB-3063 KuramotoT

liadomide:
**EpileptorT**

As reported here https://req.thevirtualbrain.org/browse/TVB-3062 `EpileptorT` model is not working to be imported in Python, due to a problem on attribute `variables_of_interest ` (`default` vs `choices` had invalid values given as defaults).

**KuramotoT**

As reported here https://req.thevirtualbrain.org/browse/TVB-3063 `KuramotoT` models is not having a valid `state_variable_range` interval (only given `low` and not `high`).

This wrong `state_variable_range` interval repeats also for `EpileptorT` generated class. 

**Questions:**

Maybe there issues are pointing towards something that needs to be changed in the generation mechanism  ?

Also, is it ok that we have these 2 models submitted on GitHub next to the classic TVB models? shouldn't we leave the user generate these at runtime, if desired ?


romina1601 (approved):

DeLaVlag:
Agreed to remove these models and leave it to the user to generate these models at runtime. If all ratemls fields are properly filled, the error should not occur after conversion. This is an issue

PR #639, opened 2023-01-11
WID-14 Improve the HTML representation on a Traited obj in Jupyter lab

rarescodemart:
I converted the description for each model from RST format to HTML and I managed to render the mathematical equations (which are present in it) using MathJax.


liadomide:
The result would look as in the attached image
Screenshot 2023-01-11 at 21 59 27

liadomide (approved):

i-Zaak:
> The result would look as in the attached image Screenshot 2023-01-11 at 21 59 27

This looks great! However is there a reason to duplicate the class docstrings in the `dfun`?

liadomide:
> This looks great! However is there a reason to duplicate the class docstrings in the `dfun`?

`dfun.__doc__` is expected to have only the equations (plus minimum text in some cases). The Class docstring is more generic and the eq there help for the overall description of the Model.

It is a decision we made a long time ago, to keep the eq in both places. 
Some places which have less space to display (here, the PhasePlane in tvb web GUI) are better if only the equations are given. The overall Sphinx generated documentation can benefit for the class extended documentation.

It might be that we need to revise this decision...
For the current task we only took the existent convention/state and worked on the display in jupyterlab ;-)

i-Zaak:
I see, wasn't aware of that - in that case it of course makes sense. I guess the class docstring then would be shown for the component instance representation, to render e.g. the `Coupling` equations. 

Alternatively one could keep the simulator representation as is (no equations) and show more details for the individual components separately.

i-Zaak:
And one more thing - do you want to do some more improvements for the HTML representations in this PR, or just focus on the equations?

i-Zaak:
>  Could we delay answering until our next call 17 Jan? 

Of course. In the meantime, these might be relevant in the long-term:
 - https://github.com/ipython/ipython/issues/3687
 - https://github.com/python-lsp/docstring-to-markdown/issues/25
 - https://github.com/jupyter/papyri

liadomide:
> Alternatively one could keep the simulator representation as is (no equations) and show more details for the individual components separately.

That might be a more generic strategy to follow (instead of the current code in this PR). 
Because on `Model` we have `dfun._doc__` (as used here), but `Coupling` has only an equation on the class doc (in this PR not shown)

tvb_library/tvb/basic/neotraits/info.py
```diff
@@ -33,6 +33,7 @@
 import numpy
 import typing
 
+from docrepr import sphinxify
```
liadomide, line 36: We should do a `try ... except ImportError` as before, to avoid breaking working envs

tvb_library/tvb/basic/neotraits/info.py
```diff
@@ -94,23 +95,35 @@ def narray_summary_info(ar, ar_name='', omit_shape=False):
...
 
     ret = {}
-    if not omit_shape:
-        ret.update({'shape': str(ar.shape), 'dtype': str(ar.dtype)})
 
-    if ar.size == 0:
-        ret['is empty'] = 'True'
+    if condensed_form:
+        condensed_desc = ""
+        if ar.shape == (1,):
```
liadomide, line 101: I think for condensed form also `is empty` or `has NaN` would be interesting to display.
What about `arr = None ` ?  NPE ?

tvb_library/tvb/basic/neotraits/info.py
```diff
@@ -156,3 +177,16 @@ def trait_object_repr_html(self):
...
     return '\n'.join(result)
+
+
+def prepare_html(doc):
+    # type: (str) -> str
+    """
+    Create html (that can be further enhanced by MathJax) from the description received as parameter
+    """
+    try:
+        html = sphinxify.sphinxify(doc, "sphinxify_dir", "html")
```
liadomide, line 188: If this remains this short maybe put the `sphinxify` call inside `trait_object_repr_html ` function directly and avoid having a supplementary function here?
It is not used in other places, right ?

tvb_library/tvb/basic/neotraits/info.py
liadomide, line 36: I look at the many dependencies of Sphinx https://github.com/sphinx-doc/sphinx/blob/master/pyproject.toml
Also at its size (6.7MB) on Pypi https://pypi.org/project/Sphinx/#files

Although license is BSD (3 clauses) for Sphinx, I would give a big vote to stay with `docutils` instead of `docrepr` as new dependency to add for tvb-library, for being much lighter.

Also, as we are packaging TVB for EBRAINS in Spack,  `docrepr` would not exist https://packages.spack.io/ and we would need to pack it, as well as maybe some of its dependencies, while `docutils` already exists.

rarescodemart: I dropped the docrepr commit

rarescodemart: I dropped the docrepr commit

tvb_library/tvb/basic/neotraits/info.py
liadomide: I think None case is not solved. Is it ?

tvb_library/tvb/basic/neotraits/_declarative_base.py
```diff
@@ -31,10 +31,17 @@
...
-from tvb.basic.neotraits.info import auto_docstring
+from tvb.basic.neotraits.info import auto_docstring, prepare_html
 from tvb.basic.logger.builder import get_logger
 
+try:
+    from IPython.display import display_html
+except ImportError:
+    def display_html(doc, raw):
+        return doc
+
```
liadomide, line 44: This has a nice outcome, but I am not convinced the ipython dependency is healthy, being here at the very core.
image
@maedoc @i-Zaak what do you think ?

liadomide:
With only one dependency in `tvb.basic.neotraits.info` towards `docutils`, we get the following on instance level:

For a model instance : 
image

A coupling instance:
image

And the simulator:
image

@i-Zaak would this outcome be ok ?

i-Zaak:
> @i-Zaak would this outcome be ok ?

These look great!

i-Zaak:
I guess the next step would be to fix the representation of the monitors instances and the connectivity, correct?

liadomide:
> I guess the next step would be to fix the representation of the monitors instances and the connectivity, correct?

image

image

i-Zaak:
image

And the connectivity breaks when instantiated like this:

image

liadomide:
> And the connectivity breaks when instantiated like this:

Thank you for testing. I fixed the HTML/summary_info result for "empty" Connectivity and Surface (that was similarly broken).

For the Bold monitor, the doc string on the class (having a table or papers) is clashing with our html table for the attributes. I leave that to @rarescodemart on Monday to find a good compromise.

liadomide (approved):

i-Zaak:
Hello, just had a quick look - looks great to me across the components. Only before it was also working in the VSCode (here on codespaces), whereas now the equations don't render anymore. I personally don't care that much, but it was a nice-have...

liadomide:
> before it was also working in the VSCode (here on codespaces), whereas now the equations don't render anymore

Could you put a before & after screenshot, so that we know which parts are affected?
Or, can you identify if it was the docstrings small fixes that produced this effect, or was it the HTML representation ?

liadomide:
I created https://req.thevirtualbrain.org/browse/TVB-3067 for the Profile issue you report. We should fix that.

Seeing the current state in VSCode on a model instance, would it be fair to assume that before you just had the table of traited attributes, but no equation string (as that was only recently added in this HTML representation ?

PR #642, opened 2023-01-27
WID-165: display the html representation for the classes

rarescodemart:
Displays a specific html representation for classes.

image


liadomide:
We separate here the Class HTML representation from #639.
We can have a nice outcome on the Class HTML representation, but I am not convinced the ipython dependency is healthy, being here at the very core (where the MetaClass for HasTraits is defined).
Also we would need to add it as dependency of tvb-library package (not done yet, until we decide to proceed with this direction).

Personally, I would vote against merging this, but i am open to arguments pro.

@i-Zaak What is your opinion ?

PR #647, opened 2023-02-22
TVB-2513 Avoid numpy deprecation warnings

liadomide:


popaula937 (approved):

popaula937:
I see in the frmk tests some similar warnings regarding np.bool and np.str, should we change those as well?

liadomide:
> I see in the frmk tests some similar warnings regarding np.bool and np.str, should we change those as well?

Yes, I think it is a good idea.

PR #656, opened 2023-03-02
HBP-218: add a field for method name in elasticsearch logs

rarescodemart:


liadomide (changes requested):
I think you should remove some unit-tests recently added, as the method tested there is now missing

liadomide (approved):

PR #659, opened 2023-03-06
HBP-184: add a backup cronjob in the openshift deployment configuration

rarescodemart:


tvb_build/hbp_openshift_deployer/tvb_deployer.yml
```diff
@@ -520,5 +520,49 @@ objects:
           targetPort: 5432
       selector:
         deploymentconfig: pg-bouncer
+  - kind: CronJob
```
liadomide, line 523: After internal discussions we decided to separate the CronJob in a separate .yml file

PR #660, opened 2023-03-06
HBP-184: Create an YML template file that generates a backup cronjob in OKD

rarescodemart:
- Have a separate YML file (than the one creating the TVB image initially) as we might want to edit this job
  - edit of the CronJob we found how to do only with the YML file, using the same name 
- We want to have a DB backup in EBRAINS OKD for the official TVB images
- maybe run it every weekend (to not propagate immediately a mistake) as the backup is not incremental


tvb_build/hbp_openshift_deployer/tvb_backup.yml
```diff
@@ -0,0 +1,84 @@
...
+      name: backup
+      annotations:
+        volume.beta.kubernetes.io/storage-class: managed-nfs-storage
+        volume.beta.kubernetes.io/storage-provisioner: fuseim.pri/ifs
+    spec:
+      accessModes:
+        - "ReadWriteOnce"
+      resources:
+        requests:
+          storage: "2Ti"
```
liadomide, line 39: Why this big ? the DB original volume is only 100GB

tvb_build/hbp_openshift_deployer/tvb_deployer.yml
```diff
@@ -292,6 +292,9 @@ objects:
                   cpu: '3'
                 requests:
                   cpu: '1'
+              env:
+                - name: APP_ENCRYPTION_HANDLER
+                  value: 'true'
```
liadomide, line 297: @popaula937 We recently found this flag from missing from the YML by reading through the code (on what is expected). Are you aware of any reason why this should not be added here ?

liadomide (approved):

tvb_build/hbp_openshift_deployer/tvb_deployer.yml
popaula937: No reason against it, most probably was forgotten while working on this setup.

tvb_build/hbp_openshift_deployer/tvb_backup.yml
```diff
@@ -0,0 +1,84 @@
...
+      jobTemplate:
+        spec:
+          template:
+            spec:
+              containers:
+                - name: tvb-postgres-backup
+                  image: docker-registry.ebrains.eu/okd_tools/postgresql_backups:1.1
+                  env:
+                    - name: DATABASE_HOST
+                      value: tvb-postgres
```
liadomide, line 69: Maybe document here that this should match in OKD the postgresql service name?

tvb_build/hbp_openshift_deployer/tvb_backup.yml
```diff
@@ -0,0 +1,84 @@
...
+      suspend: false
+      successfulJobsHistoryLimit: 7
+      failedJobsHistoryLimit: 7
+      jobTemplate:
+        spec:
+          template:
+            spec:
+              containers:
+                - name: tvb-postgres-backup
+                  image: docker-registry.ebrains.eu/okd_tools/postgresql_backups:1.1
```
liadomide, line 66: Here we should change after we have a new release with ubuntu:latest and a newer pg_dump version

liadomide (changes requested):

liadomide (approved):

PR #663, opened 2023-03-10
K Ion Exchange model

i-Zaak:


i-Zaak:
@liadomide any idea why the win framework tests fail? I'm not too worried as this PR is in the library though..

liadomide:
@i-Zaak i did not investigate in depth, but that multi process test has been "randomly" failing before. I suspect something related to the env initialization...
As the framework and library tests are passing in all other setups, I think it is ok to ignore that failing test.

PR #672, opened 2023-05-03
Corrected stimulus preparation for cosimulator.py

dionperd:
I moved the _prepare_stimulus() method call to the configuration of the CoSimulator, because during co-simulation the simulation_length is modified and equalized with the synchronization_time, leading to an erroneous reconfiguration of the stimulus in time.


liadomide (approved):
I think PR #673 should be merged before this one, rebase against master, and then contrib unit-tests should be passing

maedoc:
hi @dionperd, I'm a bit late for this one.  still relevant for merging? will you rebase as Lia suggested?

PR #675, opened 2023-05-12
[WIP] Framework to download entities from Zenodo

abhi-glitchhg:
Using [Pyzenodo3](https://github.com/space-physics/pyzenodo3) to download the repositories. 

Still work in progress. Could think of using `Pooch` to handle the downloading and remove the `utils.py`.


i-Zaak:
Oh that is super simple:

```python
import pooch

files = pooch.retrieve(
    url='doi:10.5281/zenodo.4263723/tvb_data.zip',
    known_hash=None,
    downloader=pooch.DOIDownloader(),
    processor=pooch.Unzip()
)
```

The download is cached, and you can also substitute the `None`  with actual hash which gets printed out on first download...

abhi-glitchhg:
I will improve the docs and tests soon. 

an example how user can download and fetch the data. 



```python 

from tvb.datasets import TVB_Data

tvb_data = TVB_Data(version = "2.0.3")
connectivity_66_path = tvb_data.fetch_data("connectivity_66.zip") 
print(connectivity_66_path, connectivity_66_path.is_file())

ql_bold_path = tvb_data.fetch_data("QL_BOLD_regiontimecourse.mat")
print(ql_bold_path, ql_bold_path.is_file())


```

output of the above script. 

```bash
file tvb_data.zip is downloaded at /home/abhijit/.cache/pooch/df068f14e199441033182c2ea36820e8-tvb_data.zip
/home/abhijit/oss/gsoc/tvb-root/tvb_library/tvb_data/connectivity/connectivity_66.zip True
/home/abhijit/oss/gsoc/tvb-root/tvb_library/tvb_data/berlinSubjects/QL_20120814/QL_BOLD_regiontimecourse.mat True

```

tvb_library/tvb/datasets/tvb_data.py
```diff
@@ -0,0 +1,122 @@
...
+    
+    def __init__(self, version= "2.7"):
+        """
+        Constructor for TVB_Data class 
+
+        parameters
+        -----------
+
+        version: str
+              - Version number of the dataset, Default value is 2.
```
abhi-glitchhg, line 23:
```suggestion
        version: str
              - Version number of the dataset, Default value is 2.7
```
a typo

tvb_library/tvb/datasets/zenodo.py
```diff
@@ -0,0 +1,111 @@
...
+
+        
+
+        for file in self.data["files"]:
+            url = file['links']['self']
+            known_hash = file['checksum']
+            file_name = file['key']
+            
+            file_path = pooch.retrieve(url= url, known_hash= known_hash, progressbar=True)
+
```
abhi-glitchhg, line 39: Due to this, the files are downloaded to `~/.cache/pooch`, I plan to change this to `~/.cache/tvb`.

tvb_library/tvb/datasets/zenodo.py
```diff
@@ -0,0 +1,111 @@
...
+            raise AttributeError("No files to download! Please check if the record id entered is correct! or the data is publically accessible")
+
+        
+
+        for file in self.data["files"]:
+            url = file['links']['self']
+            known_hash = file['checksum']
+            file_name = file['key']
+            
+            file_path = pooch.retrieve(url= url, known_hash= known_hash, progressbar=True)
```
abhi-glitchhg, line 38: as im using pooch.retrieve, the files are downloaded to `~/.cache/pooch`, I plan to change this to `~/.cache/tvb`.

tvb_library/tvb/datasets/base.py
```diff
@@ -0,0 +1,19 @@
+
+
+class ZenodoDataset:
+
+    def __init__(self, version):
+        self.version = version
```
liadomide, line 6:
```

class BaseDataset:

    def __init__(self, version, target_download=None):

        self.log = get_logger(self.__class__.__module__)
        self.cached_files = None
        self.version = version

        if target_download is None:
            target_download = TvbProfile.current.DATASETS_FOLDER
        if os.path.exists(target_download):
            self.log(f"Folder {target_download} already exists, it will not be created again.")
        else:
            self.log(f"Folder {target_download} does not exist, we will create it now ...")
            os.mkdir(target_download)
        self.target_download = target_download
```
And in `tvb.basic.config.profile_settings.py` BaseSettingsProfile have a new attribute initialised in init: `self.DATASETS_FOLDER = os.path.join(self.TVB_STORAGE, "DATASETS")`

abhi-glitchhg: this looks good!

tvb_library/tvb/datasets/tvb_data.py
```diff
@@ -0,0 +1,122 @@
...
+import json
+import pooch
+from pathlib import Path
+import logging
+from zipfile import ZipFile
+
+from .base import ZenodoDataset
+from .zenodo import Zenodo, Record, BASE_URL
+
+class TVB_Data(ZenodoDataset):
```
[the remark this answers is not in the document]
abhi-glitchhg: Sure!

tvb_library/tvb/datasets/base.py
abhi-glitchhg: i feel that the `target_download` is not required when we are instantiating TVBZenodoDataset class but when we are fetching the data.
```python
        if target_download is None:
            target_download = TvbProfile.current.DATASETS_FOLDER
        if os.path.exists(target_download):
            self.log(f"Folder {target_download} already exists, it will not be created again.")
        else:
            self.log(f"Folder {target_download} does not exist, we will create it now ...")
            os.mkdir(target_download)
         
```
So, above part looks good to me, but it should be placed in fetch_data method. 

```python 

          self.target_download = target_download  
```

Im bit against of linking the target_download variable to the class because this way we are fixing the directory where the data will be extracted. 

Lets say one wants to extract `connectivity_66.zip` to `dir1` and `connectivity_192.zip` to `dir2`, he will have to create two separate instances. Keeping it as a parameter to `fetch_data` method would be better in my opinion.   

```python 
#when target_download is used while instantiating class. 

connectivity_66 = TVBZenodoDataset(target_download = "dir1").fetch_data("connectivity_66.zip")
connectivity_192 = TVBZenodoDataset(target_download = "dir1").fetch_data("connectivity_192.zip")

#we needed two separate instances of the class 
```

vs 

```python
#when target_download is a parameter for fetch_data class

tvb_data = TVBZenodoDataset()
connectivity_66 = tvb_data.fetch_data("connectivity_66.zip", "dir1")
connectivity_192 =tvb_data.fetch_data("connectivity_192.zip", "dir2")

#only one instance is enough.
```

Let me know what you think @liadomide .

liadomide: In my opinion one single `target_download` per instance of `BaseDataset ` is enough.
I feel that changing the `target_download` is happening rarely.
If we keep it in the  `fetch_data` as param, then the user needs to pass that folder path multiple times (which might be annoying, if the same value).

tvb_library/tvb/datasets/zenodo.py
liadomide: Can we use here the target_download param that we discussed earlier ?

tvb_library/tvb/datasets/base.py
abhi-glitchhg: okay, sounds good!

PR #687, opened 2023-07-22
Fix the indentation mistakes for workflow files in TVB-1999-deo branch

abhi-glitchhg:
If this doesnt work, ill revert back to original settings. 

😢


liadomide:
You should do a rebase(or merge) against the current `master`.
These files look a bit different in there.
Then I suggest, just for the running of workflows create a separate PR directly against `master`

PR #689, opened 2023-08-22
TVB-3089 Compatibility with latest numexpr

liadomide:
`numexpr.evaluate` implementation slightly changed in the latest release 2.8.5.
Here https://github.com/pydata/numexpr/blob/master/numexpr/necompiler.py#L985 is no longer considering the `global_dict` values, just the `local_dict` (2nd positional arg).
Code from the prev release 2.8.4 is here: https://github.com/pydata/numexpr/blob/ec5e90886b021558e51e35870735e7a64500c59e/numexpr/necompiler.py#L819

I propose to leave the call towards `RefBase.evaluate` only in the base `Equation.evaluate` implementation, if possible - like in this PR - for simplicity in fixing future similar API changes.

### DIFF
--- a/tvb_library/tvb/datatypes/equations.py
+++ b/tvb_library/tvb/datatypes/equations.py
@@ -423,8 +423,7 @@ class Gamma(HRFKernelEquation):
             product *= i + 1
 
         self.parameters["factorial"] = product
-        _pattern = RefBase.evaluate(self.equation,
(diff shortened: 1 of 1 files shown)


maedoc (approved):
yes, it makes sense to me to let the base class keep the call DRY.
ty!

PR #713, opened 2024-04-09
EBR-30: Notify new TVB releases into EBRAINS KG

AdrianValentinCiu:
Considered:
- https://servicemeta.apps.tc.humanbrainproject.eu/
- https://gitlab.ebrains.eu/data-services/servicemeta
- https://gitlab.ebrains.eu/ri/tech-hub/issue-bonanza/-/issues/33


tvb_framework/tvb/interfaces/web/controllers/about_controller.py
```diff
@@ -0,0 +1,42 @@
...
+import cherrypy
+import json
+
+from tvb.interfaces.web.controllers.base_controller import BaseController
+
+
+class AboutController(BaseController):
+    @cherrypy.expose
+    def about(self):
+        with open('./codemeta.json', 'r') as f:
```
liadomide, line 40: This PATH will not work in case of `TVB_Distribution` or `Pypi` installations.
We need to ensure this file is found in all setups

liadomide (approved):
It is a bit odd, to have these JSON files in the code with the hard-coded version number and details, and the need to have 2 of them (one for the Software and one for the WebService)

apdavison:
The use of servicemeta is experimental for us (EBRAINS Curation team), so we might get back to you with change requests once this is deployed, but as far as I can see this looks good.

apdavison:
> It is a bit odd, to have these JSON files in the code with the hard-coded version number and details, and the need to have 2 of them (one for the Software and one for the WebService)

Depending on your build process, perhaps you could reduce duplication by generating the servicemeta from the codemeta with some kind of diff?

PR #714, opened 2024-04-10
EBR-45 Review settings page flow

AdrianValentinCiu:
When TVB web starts for the first time, we expect the user choose TVB data folder to be empty (or not exist).
If the user clicks the button "Validate DB", then an empty db file gets created into the freshly chooses folder.
This DB file can be ignored in the flow of setup

### DIFF
--- a/tvb_framework/tvb/core/services/settings_service.py
+++ b/tvb_framework/tvb/core/services/settings_service.py
@@ -184,7 +184,7 @@ class SettingsService(object):
 
         # Storage changed but DB didn't, just copy TVB storage to new one.
         if storage_changed and not db_changed:
-            shutil.copytree(previous_storage, new_storage)
+            shutil.copytree(previous_storage, new_storage, dirs_exist_ok=True)
 
         if not os.path.isdir(new_storage):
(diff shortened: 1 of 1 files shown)


tvb_framework/tvb/core/services/settings_service.py
```diff
@@ -184,7 +184,7 @@ def save_settings(self, **data):
 
         # Storage changed but DB didn't, just copy TVB storage to new one.
         if storage_changed and not db_changed:
-            shutil.copytree(previous_storage, new_storage)
+            shutil.copytree(previous_storage, new_storage, dirs_exist_ok=True)
```
liadomide, line 187: Is this line change needed also? or just the one bellow?

AdrianValentinCiu: Yes. shutil.copytree would throw an error because the folder with the db already existed since it was created when the validate db button was pressed on the first startup. dirs_exist_ok=True ensures that the missing filles and dirs are added alongside the db.

liadomide (approved):

liadomide (approved):

PR #717, opened 2024-05-17
EBR-53. Move RateML under TVB contrib

romina1601:


tvb_build/notebook_runner.py
```diff
@@ -65,7 +65,7 @@ def execute_notebook(notebook):
         'launching_bids_adapter.ipynb',
         'model_generation_using_dsl.ipynb',
         'RateML_CUDA_on_HPC.ipynb',
-        'RateML_Python_TVB.ipynb',
+        'exploring_K_ION_Exchange_model.ipynb',
```
liadomide, line 68: `exploring_K_ION_Exchange_model.ipynb` can be put back in the run. Then we can merge

PR #723, opened 2024-07-23
EBR-82 Compatibility with siibra 1.0a14

AdrianValentinCiu:


liadomide (changes requested):
The Siibra Notebooks seems to be broken also.
That one should also be fixed after this upgrade.

romina1601:
The latest version of the Julich parcellation (3.1) doesn't seem to have any information about structural or functional connectivity, so it doesn't make sense to show it as an option in the Siibra Connectivity page.
Maybe in the siibra_base.get_last_version() we could get the latest version that _has_ information about the connectivities

AdrianValentinCiu:
> The latest version of the Julich parcellation (3.1) doesn't seem to have any information about structural or functional connectivity, so it doesn't make sense to show it as an option in the Siibra Connectivity page. Maybe in the siibra_base.get_last_version() we could get the latest version that _has_ information about the connectivities

siibra_base.get_last_version() is needed to return the last version (in this case Julich 3.1). Maybe another methods which returns the latest version that has information about the connectivities should be implemented (in this case for Julich 3.0.3)

liadomide (approved):

PR #725, opened 2024-08-06
EBR-78 Windows build

AdrianValentinCiu:


tvb_build/docker/Dockerfile-build
```diff
@@ -4,8 +4,7 @@ ARG TVB_USER=tvb_user
 ARG USER_HOME=/home/$TVB_USER
 ARG TVB_STORAGE=$USER_HOME/TVB_STORAGE
 ARG KEYCLOAK_CONFIG=$USER_HOME/keycloak_config
-ARG JUPYTER_CONFIG=$USER_HOME/.jupyter
-
+ARG JUPYTER_CONFIG=$USER_HOME/.jupyter>
```
liadomide, line 7: This this change ?

tvb_build/docker/Dockerfile-win
```diff
@@ -27,6 +26,7 @@ RUN activate tvb-run && python setup.py develop
 
 COPY requirements_group requirements.txt
 RUN activate tvb-run && pip install -r requirements.txt
+RUN pip install -U scipy
```
liadomide, line 29: do we still need the pip install -U scipy if we removed the allensdk from req file ?

AdrianValentinCiu: I don't think so. I will remove it and build it again.

tvb_build/docker/Dockerfile-build
AdrianValentinCiu: I will revert this change to the original one.

liadomide:
#732 will redo this in another branch

PR #728, opened 2024-09-03
add a tvb_kernels package

maedoc:
This is a WIP to add a computation kernels package, focused on moving known hotspots from TVB's Python code into C. The set of kernels is for the moment purposefully small, covering just the delay-coupling calculation, in order to move through the whole integration, packaging & deploy process prior to achieving broader coverage.

While the kernels package will have wheels on every relevant platform, a basic NumPy fallback implementation will be maintained, with a primary goal of testing correctness.

The PR add a workflow to do standard pip-based build as well as a spack based build in anticipation of ebrains integration, but here's a few things required before a merge:

- [ ] add use of coupling kernels in TVB w/ benchmarks
- [ ] create a pypi project to submit wheels to
- [ ] add [ci build wheels](https://nanobind.readthedocs.io/en/latest/packaging.html#step-6-build-wheels-in-the-cloud) for all platforms
- [ ] add a Spack package definition
- [ ] add dep on this package to tvb library (both pip and spack) & test
- [ ] test within Ebrains & TVB Dockerfile builds
- [ ] clean up commit history


maedoc:
hm, actions run on my branch but don't show up here 🤔

PR #733, opened 2024-11-15
TVB-3109 created pyproject.toml for tvb packages

AdrianValentinCiu:


tvb_framework/tmp/pyproject.toml
```diff
@@ -0,0 +1,138 @@
+# -*- coding: utf-8 -*-
```
liadomide, line 1: do we need this `tmp` on git repo ?

AdrianValentinCiu: No. I will delete it.

Issue #544, opened 2022-04-01
Improve speed of reduced set models


maedoc:
#### Describe the new feature or enhancement
The reduced set models are quite slow and should be improved as with other models in the TVB library. 

#### Describe your proposed implementation
The [dfun](https://github.com/the-virtual-brain/tvb-root/blob/master/scientific_library/tvb/simulator/models/stefanescu_jirsa.py#L196) implementations can be replaced by Numba functions with explicit loops, which will provide an immediate speed up with the default backend.

#### Describe possible alternatives
The differential equations could annotated as with the [MPR model](https://github.com/the-virtual-brain/tvb-root/blob/master/scientific_library/tvb/simulator/models/infinite_theta.py#L148) to make them compatible with the Numba code gen backend.  The backend may need to be taught to generate code for the mode coupling.

#### Additional comments
Following a request from the TVB list this morning, cf https://groups.google.com/g/tvb-users/c/GovsAb-xc1k/m/vCTOQcd_FAAJ

peeplika:
Hey there , Can you eloborate on this a bit , I want to work on this ,but as a newbie I am a bit confused that how should I begin with that.

maedoc:
The `dfun` method linked to in the issue is a bottleneck for simulations using this model.  This issue would involve writing a new implementation with Numba.  This can be approached by writing a benchmark of the current code, creating a second implementation, checking the second implementation produces identical results as the first, and finally checking that the second implementation is faster.

peeplika:
Thanks for the detailed explanation. I’ll start working on this issue using the information you’ve provided. I’ve observed Numba implementations in other models and I think they can be used as references for this task.

peeplika:
Hello,
I've tried re-implementing the ```dfun``` functions, but unfortunately, the results are not identical to the previous implementation. I've attached a PR for your review. Could you please take a look and provide any feedback ?
Thank you

PR #738, opened 2025-01-23
Numba reduced set

samkitshah1262:
towards #544


samkitshah1262:
@liadomide @maedoc could you please review the following PR. Moreover, can you please guide me on how to test the reproducibility of similar results in old and new code. Thanks!

maedoc:
hi, thanks for your contribution. In fact, we cannot accept the current modifications because the existing NumPy code is the reference implementation. Instead, I would suggest approaching this by creating a separate Numba implementation, not modifying in-place the existing NumPy code.  This would then allow you to test your Numba against the existing NumPy code.

samkitshah1262:
@maedoc added back the numpy code. Can you please tell how can I carry out the testing ( I am quite new to TVB, apologies if it's something trivial). Thanks! :)

PR #740, opened 2025-03-12
EBR-75. Upgrade to numpy>2.0 and scipy>1.15

romina1601:


tvb_contrib/tvb/contrib/tests/cosimulation/parallel/double_proxy_precision_complex_delay_update_test.py
```diff
@@ -36,8 +36,8 @@ class TestDoubleProxyPrecisionComplexDelayUpdate(BaseTestCase):
     """
 
     def test_double_precision_complex_delay_update(self):
-        weight = np.array([[5, 2, 4, 0], [8, 5, 4, 1], [6, 1, 7, 9], [10, 0, 5, 6]], dtype=np.float_)
-        delay = np.array([[7, 8, 5, 1], [9, 3, 7, 9], [4, 3, 2, 8], [9, 10, 11, 5]], dtype=np.float_)
+        weight = np.array([[5, 2, 4, 0], [8, 5, 4, 1], [6, 1, 7, 9], [10, 0, 5, 6]], dtype=np.float64)
+        delay = np.array([[7, 8, 5, 1], [9, 3, 7, 9], [4, 3, 2, 8], [9, 10, 11, 5]], dtype=np.float64)
```
liadomide, line 40: why only here needs to be changed?

I see other places where `np.float_` is still used

romina1601: thanks for the observation, I will update them everywhere

tvb_library/tvb/basic/neotraits/_attr.py
```diff
@@ -298,8 +298,19 @@ def __validate(self, value):
...
+        if numpy.issubdtype(type(value), numpy.floating) and numpy.issubdtype(self.field_type, numpy.integer):
+            raise TypeError("can't be set to {!r}. Not safe to cast a float to an int.".format(value), attr=self)
+
+        # this replaces the old can_cast method, which changed drastically in v2.0
+        try:
+            value_as_array = numpy.array(value, dtype=self.field_type)
+        except Exception:
             raise TraitTypeError("can't be set to {!r}. No safe cast.".format(value), attr=self)
+        if value_as_array.item() != value:
+            raise TraitTypeError("can't be set to {!r}. No safe cast.".format(value), attr=self)
```
liadomide, line 312: what if here you write 
` value_as_array.item() != value and type(value) != type(value_as_array[0])`
Wound't that give you the previous behaviour ?

romina1601: I assume you mean to add  `and type(value) != type(value_as_array.item())` and remove the part with 
```
# check if trying to cast a float to an int
if numpy.issubdtype(type(value), numpy.floating) and numpy.issubdtype(self.field_type, numpy.integer):
```
?
This won't work, because in case of `value = 4.0` and `field_type=int`
`value_as_array.item() != value ` returns `False ` (they will be both 4)
so we still need to check the type first

liadomide: Sorry, I meant OR

liadomide: because `int -> float` vs  `float -> int` assignment were treated asymmetrically in the past, we propose now to make the check simpler, just `numpy.array(value, dtype=self.field_type)` and the outcome value check `value_as_array.item() != value` should be fine

liadomide (approved):

liadomide:
Are we compatible with these changes with `numpy< 2` ?
If not, we should put the constraint in the dependencies for `numpy>=2`

romina1601:
> Are we compatible with these changes with `numpy< 2` ? If not, we should put the constraint in the dependencies for `numpy>=2`

I tested the backward compatibility (ran the tests & started the application in an env with numpy 1.26.4) and everything seemed to work fine. So I would say we are backwards compatible with numpy <2.0 as well

liadomide:
@maedoc do you see a problem if we merge this ?

PR #743, opened 2025-03-21
another take on hybrid models

maedoc:
This is a another take on hybrid models, where multiple models are used at once, which is a more array oriented than #739 and closer to TVB style with traits etc.  It introduces the following elements to the TVB datatypes (naming suggestions welcome)

- *subnetwork*: compromises a set of nodes by combining a connectivity with a mask
- *projection*: connects a cvar from a source subnet to a cvar in a target subnet
- *networkset*: collects subnetworks and projections into a coherent whole where we can
  - compute the coupling across all projections
  - compute all dfuns
  - apply an integration scheme to all dfuns to take a step in time

From there, it turns out that flexible state shapes creates surprise in many places, where it's assumed that the state variables are homogenous across nodes.  For instance, the integrator noise is set per state variable, however in different models, these need to be set with different values.

Other major nuances are related to monitors, but I think some straightforward compromises are available:  for "debug" monitors like raw or temporal average, the monitor will be applied per model to produce a per-subnetwork trace.  Monitors more "neuroimaging" like will choose a particular state variable and apply the forward model (gain matrix, BOLD equations) to it.

Stimulation is another open question, but I wonder if, deparating from the monolithic style, a stimulus cannot just be a "subnetwork" with a projection to the nodes being stimulated? Ideas welcome.

Performance is terrible w/ ~5-10k iter/s on default connectome w/ two models, but this is expected w/ the naive implementation and will improve dramatically after massaging for jax or a c++ impl.

@i-Zaak @ReyJ94 feedback welcome!

### DIFF
--- a/tvb_library/tvb/simulator/hybrid.py
+++ b/tvb_library/tvb/simulator/hybrid.py
@@ -1,0 +1,126 @@
+# -*- coding: utf-8 -*-
+#
+#
+# TheVirtualBrain-Scientific Package. This package holds all simulators, and
+# analysers necessary to run brain-simulations. You can use it stand alone or
+# in conjunction with TheVirtualBrain-Framework Package. See content of the
+# documentation-folder for more details. See also http://www.thevirtualbrain.org
+#
+# (c) 2012-2024, Baycrest Centre for Geriatric Care ("Baycrest") and others
+#
+# This program is free software: you can redistribute it and/or modify it under the
+# terms of the GNU General Public License as published by the Free Software Foundation,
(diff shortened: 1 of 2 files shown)


maedoc:
(btw, this scratches the multiple-connectome itch, as well, since multiple projections can be configured with different connectomes, and will be summed to the same cvar)

maedoc:
at this point several issues to still address

- [ ] not all monitors work
- [ ] time delays not yet included
- [ ] source/target cvars should not be scalar
- [ ] it's slow

PR #746, opened 2025-03-30
[ISSUE-544] improve speed of reduced set models

omarshafik:
Reduced Set models don't use numba for enhanced performance. This uses numba njit wrapper to compile the deriviative functions during runtime, and provide considerable performance gain.

Since numba currently doesn't fully support jit-ed classes, derivative functions have to move outside the class scope. To improved readability, they were moved to a new file: _dfun_stefanescu_jirsa.py.

To test the changes a new unit test module was added: models_benchmark.py. Here, we compute the derivative multiple times within a time window with/without numba and compare the throughput. Around 5x performance gains were observed. We also verify that numba doesn't change derivative values.

### DIFF
--- a/tvb_library/tvb/tests/library/simulator/models_benchmark.py
+++ b/tvb_library/tvb/tests/library/simulator/models_benchmark.py
@@ -1,0 +1,78 @@
+import numpy as np
+import time
+from tvb.tests.library.base_testcase import BaseTestCase
+from tvb.simulator.models.stefanescu_jirsa import ReducedSetFitzHughNagumo
+from tvb.simulator.models.stefanescu_jirsa import ReducedSetHindmarshRose
+from tvb.simulator.models.base import Model
+
+
+class TestBenchmarkModels(BaseTestCase):
+    """Test cases that benchmark the performance of models' implementation
+    """
+
+    def randn_state_for_model(self, model: Model, n_node):
+        shape = (model.nvar, n_node, model.number_of_modes)
+        state = np.random.randn(*shape)
+        return state
+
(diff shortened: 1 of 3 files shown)


maedoc:
hi thx! can you just use non-zero coupling values, to ensure the coupling expressions inside the numba dfun are also tested?

omarshafik:
changed to random coupling with [d32512d](https://github.com/the-virtual-brain/tvb-root/pull/746/commits/d32512d0107cc8a464ec8cf3ad0e69b34aa4e2cc)

tvb_library/tvb/tests/library/simulator/models_benchmark.py
```diff
@@ -0,0 +1,78 @@
...
+        # start timing
+        tic = time.time()
+        n_eval = 0
+        while (time.time() - tic) < time_limit:
+            model.dfun(state, coupling)
+            n_eval += 1
+        toc = time.time()
+        return n_eval / (toc - tic)
+
+    def test_reduced_set_fitz_hugh_nagumo_when_using_numba_improve_performance(self):
```
maedoc, line 42: could we just shorten a bit these method names like `test_rsfhn_perf`?  when we run individual tests from the cli one has to type or copy/paste the method and it helps to be concise

maedoc:
looks good to me

tvb_library/tvb/tests/library/simulator/models_benchmark.py
omarshafik: done

PR #748, opened 2025-04-02
Add autotuning functionality to simulator

omarshafik:
The simulator should support monotonic autotuning parameters for quantities of interest. Here, I add a simple & generic autotuning functionality.

### DIFF
--- a/tvb_library/tvb/simulator/simulator.py
+++ b/tvb_library/tvb/simulator/simulator.py
@@ -37,6 +37,7 @@ simulation and the method for running the simulation.
 
 import math
 import time
+from typing import Callable
 import numpy
 
 from tvb.basic.neotraits.api import HasTraits, Attr, NArray, List, Float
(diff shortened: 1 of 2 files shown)


maedoc:
This is a pretty good first take on the idea.  I will think about some of the concrete use cases.  For instance, we might take a model like the reduced wong wang, and increase each region's excitability until the firing rates are all matching, then continue to increase them all until the functional connectivity match is maximized.

in any case, I think the implementation is nearly sufficient; in the case where we're trying to maximize some measurement, we'll eventually take a step and the value will go back down?  So it might make sense to trace the the tuning, and keep the best value.

What do you think?

omarshafik:
one thing to note here though is that delta is assumed to return a one dimensional value. This however, can be changed as simply as passing an n-dimensional threshold (and changing the type hints). I believe the rest of the logic should hold.. I could add that in a later commit if needed :)

PR #771, opened 2026-01-20
Hybrid model

maedoc:
Reopening because of accidental deletion of #743 , copying desc here:

This is a another take on hybrid models, where multiple models are used at once, which is a more array oriented than #739 and closer to TVB style with traits etc.  It introduces the following elements to the TVB datatypes (naming suggestions welcome)

- *subnetwork*: compromises a set of nodes + neural mass model + scheme + monitors
- *projection*: connects a cvar from a source subnet to a cvar in a target subnet
- *networkset*: collects subnetworks and projections into a coherent whole where we can
  - compute the coupling across all projections
  - compute all dfuns
  - apply an integration scheme to all dfuns to take a step in time

From there, it turns out that flexible state shapes creates surprise in many places, where it's assumed that the state variables are homogenous across nodes.  For instance, the integrator noise is set per state variable, however in different models, these need to be set with different values.

Other major nuances are related to monitors, but I think some straightforward compromises are available:  for "debug" monitors like raw or temporal average, the monitor will be applied per model to produce a per-subnetwork trace.  Monitors more "neuroimaging" like will choose a particular state variable and apply the forward model (gain matrix, BOLD equations) to it.

Stimulation is another open question, but I wonder if, deparating from the monolithic style, a stimulus cannot just be a "subnetwork" with a projection to the nodes being stimulated? Ideas welcome.

Performance is terrible w/ ~5-10k iter/s on default connectome w/ two models, but this is expected w/ the naive implementation and will improve dramatically after massaging for jax or a c++ impl.

## recent feedback

still some bugs and api is not intuitive from Frederico Tesler and Roberta Lorenzi.

## implementation status 

- [x] all monitors work
- [x] all integrators work
- [x] time delays implemented
- [x] source/target cvars are arrays not scales
- [x] make it faster #775 
- [x] subnetwork can have internal projection(s)
- [x] apis are documented
- [x] stimulation is supported
- [x] afferent coupling monitors (fix tests)
- [x] add tests against existing simulation scenarios
- [x] toy scenarios for tests integrated
- [x] better demos
- [x] clean up API for connectivity
- [x] consider re-ordering global monitor output to follow original connectome instead of concatenated subnetworks
- [x] improve coupling API
  - [x] named cvars instead of indices
  - [x] including support for more cvars than svars (`cvar=[0,0,0]` etc)
  - [x] cfuns
- [x] ensure integrators call `reconfigure_boundaries_and_clamping...`
- [x] check initial conditions are done correctly as in regular tvb
- [x] for given interprojection w/ multiple target cvars, we should have scaling value per target
- [x] for EEG/MEG/SEEG, want to have different subnets produce the EM field in their own way, but then sum it up globally on the electrodes
- [x] attach stim to subnets not netset


maedoc:
@popaula937 is merge ok if failures are "just" sonar as currently the case?

popaula937:
> @popaula937 is merge ok if failures are "just" sonar as currently the case?

Yes, seems like a configuration error, I will take a look next week.

brainy-joseph:
Hi @maedoc !

I am newer to computational neuroscience but here I have been reading through the hybrid model branch. Description here mentions JAX or a C++ implementation for speed and closed PR #775 explored Numba. Is then the plan to keep the NumPy path as a reference and add these as pluggable backends or like to replace it? I am asking this as a small doubt because per subnetwork state shapes vary and I am curious how that interacts with jit compilation.

Thank you! 😇

maedoc:
hi @brainy-joseph

> Is then the plan to keep the NumPy path as a reference and add these as pluggable backends or like to replace it?

yes, the numpy is for humans to read and the numba is the first fast-ish implementation.  

>  I am asking this as a small doubt because per subnetwork state shapes vary and I am curious how that interacts with jit compilation.

the numba templates generate code per subnetwork so the jit compiler sees lots of code as if it had all been written out by hand; jit compilation is a small weakness here. 

we can speculate about a jax and/or c++ impl being faster elsewhere, since this branch in particular is closed, feel free to open a new issue with a particular use case.

brainy-joseph:
Thank you @maedoc that makes sense. Template generated code per subnetwork explains how the varying shapes are handled. 

I am still learning about opensource computational neuroscience and would like to get hands on with the hybrid code. If there is a small task like a demo that needs testing or an open issue that would suit a newcomer, in the hybrid model or elsewhere in TVB, I would be glad to take a look :D

maedoc:
> would like to get hands on with the hybrid code. If there is a small task like a demo that needs testing or an open issue that would suit a newcomer, in the hybrid model or elsewhere in TVB, I would be glad to take a look

there are a small number of demos so far;  if you can go through them and raise any issues you find, it's more than welcome.  There certainly will need to be more documentation as well.

brainy-joseph:
Alrighty! I will set up a clean environment from the repo, work through the hybrid demo notebooks, and open issues for anything relevant I find

Thank you @maedoc :D

PR #774, opened 2026-04-07
Enable numba backend for k ion ex model

maedoc:
This PR 

- extends the Numba backend for more complicated models
- adds requisite model metadata to the KIonEx model to allow for use with Numba backend
- based on the other numba dfun kionex PR which is currently open but soon to merge


maedoc:
I also merged this despite failed actions related to Sonar

PR #775, opened 2026-04-09
Hybrid numba

maedoc:
In anticipation of merging #771 this adds a numba backend to accelerate hybrid simulations.


Ziaeemehr:
In stack result module ``_stack_numba_results`` we have the time axis always collapsed via .mean(axis=0)
so I think we need to modify that if we need time series for each sweep point.

shape (n_sweeps, n_time, n_voi, N_subnet, n_modes)

maedoc:
I'm running into Numba compile segfaults, filed an issue here https://github.com/numba/numba/issues/10593

popaula937 (commented):
@maedoc I see there are many PRs related to the Hybrid Simulator and would like to align with you on how we merge this functionality. The following points come to my mind:
- I see PR #771 is included here, could we close that one or would you prefer to merge 771 first and rebase this on top?
- there are multiple markdown files with review, fix, plan points. Are these still in progress, or are they intended as a record of future improvements? Should we consider a specific location for this type of files?
- would you like to keep the commit history as-is or should we squash/rebase into a smaller set of commits before merging? Here we should consider that PRs #778 and #784 use hybrid-numba as base branch.

maedoc:
hi @popaula937 I missed your comment during August.  I have since split this PR into several smaller components and merged the first three corresponding to the main Python implementation.  I have still 4 more PRs to do that add the numba implementation and demos.  I would like to keep this branch for reference but this PR will not be merged, so I will close it now.

PR #784, opened 2026-06-28
Chunksize fix

i-Zaak:
Relaxes the chunksize invariant. Might need a bit more testing and/or thought, but looks like it works :)


maedoc:
> I have only proved it correct, not tried it

PR #704, opened 2024-03-14
trying to solve #544

peeplika:
This is a draft re-implementation of the dfun functions in [stefanesu_jirsa.py](/tvb_library/tvb/simulator/models/stefanescu_jirsa.py) using numba, aimed at resolving issue #544

PR #796, opened 2026-09-20
Issue #544 fix : Optimize derivative evaluations and memory layouts for reduced set models

### LINKED ISSUE #544: Improve speed of reduced set models

AvipsaGhosh:
RESOLVES #544

Description
This pull request optimizes numerical integration throughput for three reduced neural models: ReducedSetHindmarshRose, ReducedSetFitzHughNagumo, and ReducedWongWangExcInh.

Profiling revealed execution overhead originating from universal function dispatching, repetitive transpositions, dynamic memory allocation, and arithmetic operations within inner loops. The following enhancements address these bottlenecks:

Stefanescu-Jirsa Models (ReducedSetHindmarshRose & ReducedSetFitzHughNagumo):

1. Kernel Migration: Replaced the previous @guvectorize dispatch with compiled @njit routines operating directly on contiguous arrays, eliminating dimension swapping, intermediate slice allocations, and Python dispatch overhead.

2. Direct Indexing: Read state variables directly across leading indices and coupling arrays without generating intermediate slice views.

3. Arithmetic Optimizations: Replaced power operations (cubing and squaring terms) with direct multiplications (xi * xi and xi_sq * xi), hoisted invariant parameter lookups outside the node loop, and converted divisions into reciprocal multiplications.

4. Buffer Management: Wrote derivative results directly into preallocated output arrays to prevent recurrent memory churn across integration steps.

5. Loop Parallelism: Distributed node iterations across available threads to scale efficiently with mode projections.

Wong-Wang Model (ReducedWongWangExcInh):

1. Parameter Extraction: Extracted wrapped NumPy arrays into scalar float primitives during configure() to eliminate array indirection inside the evaluation loop.

2, Buffer Preallocation: Preallocated a reusable derivative buffer in configure() so dfun avoids repeated heap allocations on every time-step.

3. Serial Execution: Retained single-threaded evaluation to eliminate thread coordination latency on small- to medium-sized networks, maximizing throughput across standard whole-brain parcellations.

Verification & Compatibility:

1. Preserved all original method signatures, input parameters, and output array shapes for full backward compatibility.

2. Numerical outputs remain identical to baseline implementations across integration steps.

3. Benchmarks confirm significant throughput gains (kHz) across standard parcellation sizes.


maedoc:
Thanks @AvipsaGhosh — this is squarely aimed at #544 and the right direction: compiling these reduced-set `dfun`s for the plain-library path and cutting the per-step allocations is exactly what that issue wants. Before we can merge, there are a few things I'd ask you to verify and fix:

**1. Correctness — two blockers I found reading the diff:**
- `ReducedSetHindmarshRose.dfun` passes `self.A_ik / self.B_ik / self.C_ik` into `_hr_dfun`, but `update_derived_parameters` only sets `self.Aik / self.Bik / self.Cik`. `A_ik` is never assigned anywhere, so every HR `dfun` call raises `AttributeError` — the existing `test_sj3d` (through `stationary_trajectory` → `dfun`) will fail. The FHN path correctly uses `self.Aik…`; looks like a copy/paste slip.
- `ReducedWongWangExcInh.dfun` returns the persistent `self._deriv_buffer` by reference\(^1\). The default `HeunDeterministic` (and `RK4`) scheme calls `dfun` twice per step and evaluates `m_dx_tn + dfun(inter, …)`; the second call overwrites the buffer the first result points to, so the first-stage derivative is lost (`dX = (dfun(inter)+dfun(inter))*dt/2`). The old gufunc path returned a fresh array, so this is a silent numerical regression on the default integrator. A single-call parity check won't catch it — a forward Heun simulation would.

**2. Parity test.** The description asserts outputs are identical to baseline, but the PR adds no test. Could you add a sweep (random states × params × a realistic coupling field, incl. array-valued / broadcast params and non-contiguous coupling) comparing the new `dfun` against the previous implementation, plus a short forward-time simulation through the default integrator? `test_nb_hybrid_regression_reducedset_modes.py` and `test_model_codegen_parity.py` are good scaffolds to reuse.

**3. Benchmark numbers.** The description mentions "significant throughput gains (kHz)" without figures or a script. Could you attach the before/after protocol (old vs new `dfun`, at 68/100/200/360 nodes, float64, warmup + ≥5 repeats, and report both isolated-`dfun` and full-simulation wall time under the default Heun integrator)? Without it the claim isn't reproducible.

**4. Shape semantics.** The new WW `njit` hardcodes `x[0,i,0]`/`c[0,i,0]` and `x.shape[1]`, and scalarises params via `float(self.X[0])`. The old gufunc path normalised layouts via `reshape(...).T` and broadcast `()` scalar params. Can you confirm (ideally with a small test) that the new path still accepts: extra leading dimensions, array-valued params broadcast against nodes (length > 1), non-contiguous coupling slices, and `number_of_modes > 1` (currently only mode 0 is computed)?

**5. Small things:** `fastmath=True` changes FP behavior — if bit-identity with `master` is a requirement, either drop it or state the tolerance, and please add the parity check that measures it. Also the description says FHN/HR previously used `@guvectorize` and mentions distributing node iterations across threads, but the old FHN/HR were pure NumPy and the new kernels use no `prange`/`parallel=True` — worth a quick edit so the changelog matches the code. Finally, note prior #544 attempts (#704/#738/#746) and the numba hybrid backend — this PR is complementary (plain-library path), but let's make sure it doesn't duplicate the reduced-set kernels the hybrid path already compiles.

Happy to look again once the HR attribute fix and the buffer fix are in, and the parity/benchmark artifacts are attached.

\(^1\) `_fast_ww_dfun` writes into and returns the same `deriv` array passed in; njit returns it by reference.

